PR #17, opened 2025-11-28
[KLC-1964] Implement complete rollback logic to revert account balances and history

Beroni:
## Summary
Refactor blockchain rollback logic to include account balance and account history reversion, implementing a complete 4-step rollback process triggered when blocks are removed from Elasticsearch due to consensus failures.

## Problem
When a rollback occurred due to network consensus failure, only transactions and blocks were removed from Elasticsearch indices. Account balances and account history entries were not reverted, causing balance inconsistencies that persisted after the rollback. This issue was reported via Slack thread where a transaction executed during a consensus failure period (blocks 26364634-26364635) left incorrect account balances.

## Solution
Implemented a comprehensive 4-step rollback process in the Elasticsearch indexer:

1. **Revert Account Balances** - Query account history by timestamp to find all affected accounts, then revert each account to its previous state
2. **Remove Account History** - Delete account history entries for the rolled-back block using timestamp-based bulk removal
3. **Remove Transactions** - Clean up transaction records from the block (existing functionality)
4. **Remove Block Header** - Remove the block header from indices (existing functionality)

### Optimization Strategy
The implementation uses an optimized query approach:
- Single timestamp-based search query to retrieve all account history entries from the rolled-back block
- For each affected account, fetch the last 2 history entries (sorted DESC by timestamp)
- Determine correct reversion based on 3 cases:
  - **No history found**: Log warning (edge case)
  - **Single history entry**: Account was created in this block → revert balance to 0
  - **Multiple entries**: Account existed before → revert to penultimate (second-to-last) state

This approach minimizes Elasticsearch queries and leverages indexed data rather than computing state from scratch.

**New Methods & Functionality:**
- `indexer/elasticClient.go` (+82 lines):
  - `DoBulkRemoveByTimestamp()` - Bulk delete documents by timestamp query
  - `DoSearch()` - Execute search queries with custom parameters
  - `DoUpdate()` - Update document fields by ID

- `indexer/elasticProcessor.go` (+199 lines):
  - `RemoveAccountsHistory(blockTimestamp)` - Delete all account history entries for a specific timestamp
  - `RevertAccountBalances(blockTimestamp)` - Revert account balances to previous state
  - `extractAccountHistoryEntries()` - Parse search results into structured account history data
  - `revertSingleAccountBalance()` - Handle individual account reversion with 3-case logic

- `indexer/queries.go` (+10 lines):
  - `prepareTimestampForBulkRemove()` - Build Elasticsearch term query for timestamp-based deletion

- `indexer/data/account.go` (+7 lines):
  - Added `BlockNum` field to `AccountBalanceHistory` struct for tracking

**Enhanced Orchestration:**
- `indexer/workItems/workItemRemoveBlock.go` (+36 lines):
  - Refactored to implement 4-step rollback workflow
  - Added comprehensive logging with nonce and timestamp tracking for each step
  - Proper error handling and step sequencing

**Interface Updates:**
- `indexer/interface.go`: Added `RemoveAccountsHistory()` and `RevertAccountBalances()` method signatures
- `indexer/workItems/interface.go`: Updated `removeIndexer` interface

**Mock Implementations:**
- `indexer/mock/databaseWriterStub.go` (+15 lines): Added stubs for new DB client methods
- `indexer/mock/elasticProcessorStub.go` (+10 lines): Added stubs for rollback operations

## Testing
- [x] All existing tests pass
  - Verified 4-step process executes in correct order
  - Confirmed account balances revert to previous state
  - Tested edge case: new account created in rolled-back block correctly reverts to balance 0
  - Validated account history entries are properly removed by timestamp
  - Confirmed logs show nonce and timestamp for each rollback step

## Breaking Changes
None

## Related Issues
Fixes [KLC-1964](https://klever-io.atlassian.net/jira/software/c/projects/KLC/boards/78?assignee=61eec400571f7600714f0913&selectedIssue=KLC-1964)

## Checklist
- [x] Code follows project style guidelines
- [x] Self-reviewed code for logic and edge cases
- [x] Added comprehensive code comments explaining rollback logic
- [x] Commit message follows Conventional Commits format (`chore: refactor rollback logic`)
- [x] Breaking changes documented (None)
- [x] Configuration dependencies documented
- [x] Performance optimizations implemented (timestamp-based queries, bulk operations)

[KLC-1964]: https://klever-io.atlassian.net/browse/KLC-1964?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/indexer/data/account.go
+++ b/indexer/data/account.go
@@ -33,6 +33,7 @@ type AccountInfo struct {
 	Allowance       int64         `json:"allowance"`
 	Permissions     []Permissions `json:"permissions"`
 	Timestamp       time.Duration `json:"timestamp"`
+	UpdatedAt       time.Duration `json:"updatedAt,omitempty"`
 	CodeHash        string        `json:"codeHash,omitempty"`
 	CodeMetadata    string        `json:"codeMetadata,omitempty"`
 	Foundation      bool          `json:"foundation,omitempty"`
--- a/indexer/data/account.go
+++ b/indexer/data/account.go
@@ -166,3 +166,10 @@ type ValidatorInfo struct {
 	Logo           string `json:"logo"`
 	URIs           []*URI `json:"uris"`
 }
+
+type AccountHistoryEntry struct {
+	Address       string
+	Balance       int64
+	FrozenBalance int64
+	Timestamp     int64
+}
--- a/indexer/data/account.go
+++ b/indexer/data/account.go
@@ -168,9 +168,8 @@ type ValidatorInfo struct {
 	URIs           []*URI `json:"uris"`
 }
 
-type AccountHistoryEntry struct {
+type AccountBalanceInfo struct {
 	Address       string
 	Balance       int64
 	FrozenBalance int64
-	Timestamp     int64
 }
--- a/indexer/elasticProcessor.go
+++ b/indexer/elasticProcessor.go
@@ -368,61 +368,69 @@ func (ei *elasticProcessor) RemoveAccountsHistory(blockTimestamp int64) error {
 		return nil
 	}
 
-	return ei.elasticClient.DoBulkRemoveByTimestamp(accountsHistoryIndex, blockTimestamp)
+	return ei.elasticClient.DoBulkRemoveByTimestamp(accountsHistoryIndex, blockTimestamp*1000)
 }
 
-// RevertAccountBalances will revert account balances to their state before the given block
+// RevertAccountBalances reverts the accounts index to the current state from the account trie
+// This method should be called after RevertStateToBlock, as it reads the current balance from the trie
 func (ei *elasticProcessor) RevertAccountBalances(blockTimestamp int64) error {
-	if !ei.isIndexEnabled(accountsIndex) || !ei.isIndexEnabled(accountsHistoryIndex) {
+	if !ei.isIndexEnabled(accountsIndex) {
 		return nil
 	}
-
-	// Query to get all account history entries from this block
+	// Query accounts index to find all accounts modified at this timestamp
 	query := templates.Object{
 		"query": templates.Object{
 			"term": templates.Object{
-				"timestamp": blockTimestamp,
+				"updatedAt": time.Duration(blockTimestamp * 1000),
 			},
 		},
-		"size": revertAccountBatchSize,
+		"size":    revertAccountBatchSize,
+		"_source": []string{"address", "balance", "frozenBalance"},
 	}
 
 	body, err := encode(query)
 	if err != nil {
-		log.Warn("RevertAccountBalances: failed to encode query", "error", err.Error())
 		return err
 	}
 
-	// Search for all account history entries from this block
-	res, err := ei.elasticClient.DoSearch(accountsHistoryIndex, &body)
+	res, err := ei.elasticClient.DoSearch(accountsIndex, &body)
 	if err != nil {
-		log.Warn("RevertAccountBalances: failed to search accounts history", "blockTimestamp", blockTimestamp, "error", err.Error())
 		return err
 	}
 
-	accountHistoryEntries, err := ei.extractAccountHistoryEntries(res)
+	accounts, err := ei.extractAccountData(res)
 	if err != nil {
-		log.Warn("RevertAccountBalances: failed to extract account history entries", "error", err.Error())
+
 		return err
 	}
 
-	if len(accountHistoryEntries) == 0 {
+	if len(accounts) == 0 {
 		log.Debug("RevertAccountBalances: no accounts to revert", "blockTimestamp", blockTimestamp)
 		return nil
 	}
 
-	// For each account, find the previous balance and update
-	for _, historyEntry := range accountHistoryEntries {
-		err := ei.revertSingleAccountBalance(historyEntry.Address, blockTimestamp)
+	// For each account modified in this block, get current balance from trie and update index
+	for _, accountInfo := range accounts {
+		err := ei.revertSingleAccountBalanceFromTrie(accountInfo.Address, accountInfo.Balance, accountInfo.FrozenBalance)
 		if err != nil {
-			log.Warn("RevertAccountBalances: failed to revert account", "address", historyEntry.Address, "error", err.Error())
+			log.Debug("RevertAccountBalances: failed to revert account",
+				"address", accountInfo.Address,
+				"error", err.Error())
+			continue
 		}
 	}
 
 	return nil
 }
 
-func (ei *elasticProcessor) extractAccountHistoryEntries(response templates.Object) ([]*data.AccountHistoryEntry, error) {
+type accountBalanceInfo struct {
+	Address       string
+	Balance       int64
+	FrozenBalance int64
+}
+
+// extractAccountAddresses extracts account addresses and balance information from the search response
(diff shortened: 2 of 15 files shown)


indexer/elasticProcessor.go
```diff
@@ -354,12 +355,208 @@ func (ei *elasticProcessor) RemoveTransactions(blk nodeData.HeaderHandler) error
...
+				"must": []templates.Object{
+					{
+						"term": templates.Object{
+							"address": address,
+						},
+					},
+					{
+						"range": templates.Object{
+							"timestamp": templates.Object{
+								"lte": currentBlockTimestamp,
```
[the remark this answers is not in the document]
Beroni: no need, this already tested.

indexer/data/account.go
```diff
@@ -166,3 +166,10 @@ type ValidatorInfo struct {
 	Logo           string `json:"logo"`
 	URIs           []*URI `json:"uris"`
 }
+
+type AccountHistoryEntry struct {
+	Address       string
+	Balance       int64
+	FrozenBalance int64
+	Timestamp     int64
```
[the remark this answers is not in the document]
Beroni: no need json tags for this struct.

indexer/workItems/workItemRemoveBlock.go
```diff
@@ -28,21 +28,37 @@ func (wirb *itemRemoveBlock) IsInterfaceNil() bool {
...
+		err = wirb.indexer.RemoveTransactions(blk)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Step 4: Remove block header
+	err = wirb.indexer.RemoveHeader(wirb.headerHandler)
 	if err != nil {
-		log.Warn("itemRemoveBlock.Save could not remove block transactions", "error", err.Error())
```
fbsobreira, line 45: why removed the wraning?

indexer/elasticProcessor.go
```diff
@@ -1538,6 +1710,7 @@ func (ei *elasticProcessor) buildAccountInfo(userAccount *data.Account, blockTim
 		Allowance:       ei.getAllowanceWithPendingRewards(userAccount.UserAccount),
 		Permissions:     permissions,
 		Timestamp:       time.Duration(blockTimestamp * 1000),
+		UpdatedAt:       time.Duration(blockTimestamp * 1000),
```
fbsobreira: if `updatedAt` is the same of `timestamp` makes no sense to add it.

indexer/elasticProcessor.go
```diff
@@ -354,12 +357,181 @@ func (ei *elasticProcessor) RemoveTransactions(blk nodeData.HeaderHandler) error
...
+	}
+
+	return accounts, nil
+}
+
+// revertSingleAccountBalanceFromTrie reads the current balance from the account trie
+// and updates the accounts index with that balance.
+// This assumes RevertStateToBlock has already been called, so the trie is at the correct state.
+// oldBalance and oldFrozenBalance are the values from the accounts-history index (for logging comparison).
+func (ei *elasticProcessor) revertSingleAccountBalanceFromTrie(address string, oldBalance, oldFrozenBalance int64) error {
```
fbsobreira, line 471: Im not confident with this solution, it only care for KLV balance, what about a KDA transfer, or freeze/unfreeze? Im not sure if is possible but best way would be indexing as we do the indexing saving that was changed and reload that info. or have another mechanism...

fbsobreira (changes requested):

indexer/elasticProcessor.go
Beroni: updatedAt is different from timestamp.
The timestamp value is not updated when `serializedDataForUpdateAccounts` is called, because timestamp is treated as the createdAt value.

indexer/workItems/workItemRemoveBlock.go
Beroni: I removed those logs thinking they were the ones I had added to test the revert. I'm adding them back now.

indexer/elasticProcessor.go
Beroni: The motivation behind this task was specifically the KLV balances, which directly impact only the account and account-history indexes — as highlighted in the report attached to the task.
In the future, we can expand the revert mechanism to include  accountskda index

Beroni: During my research, I also checked whether Elasticsearch had any native mechanism that could help us. According to the official documentation, Elasticsearch does not support rollbacks. It does support snapshots, but taking a snapshot for every block wouldn’t be feasible.

Since this bug is very unlikely to happen—especially after the optimization updates we’ve implemented—we can put this task on hold and look for a better solution.

indexer/elasticProcessor.go
```diff
@@ -354,12 +357,181 @@ func (ei *elasticProcessor) RemoveTransactions(blk nodeData.HeaderHandler) error
...
+	if len(accounts) == 0 {
+		log.Debug("RevertAccountBalances: no accounts to revert", "blockTimestamp", blockTimestamp)
+		return nil
+	}
+
+	// For each account modified in this block, get current balance from trie and update index
+	for _, accountInfo := range accounts {
+		err := ei.revertSingleAccountBalanceFromTrie(accountInfo.Address, accountInfo.Balance, accountInfo.FrozenBalance)
+		if err != nil {
+			log.Debug("RevertAccountBalances: failed to revert account",
```
fbsobreira: As we don't propagate the error we should log as warn

indexer/elasticProcessor.go
```diff
@@ -31,6 +32,8 @@ import (
 )
 
 const numDecimalsInFloatBalance = 10 //?
+const revertAccountBatchSize = 10_000
```
fbsobreira: lets add a comment on the 10k choice (ES query size limit... ), also lets check the line above with "//??

indexer/elasticProcessor.go
```diff
@@ -354,12 +357,181 @@ func (ei *elasticProcessor) RemoveTransactions(blk nodeData.HeaderHandler) error
...
 	return ei.elasticClient.DoBulkRemove(txIndex, encodedTXsHashes)
 }
 
+// RemoveAccountsHistory will remove account history entries for a specific timestamp
+func (ei *elasticProcessor) RemoveAccountsHistory(blockTimestamp int64) error {
+	if !ei.isIndexEnabled(accountsHistoryIndex) {
+		return nil
+	}
+
+	return ei.elasticClient.DoBulkRemoveByTimestamp(accountsHistoryIndex, blockTimestamp*1000)
```
fbsobreira: `blockTimestamp*1000` multiplied been used as param and in `RevertAccountBalances` L#384 the multiplication its been done in query.
Lets make this consistent. Also, consider using a `toMilliseconds(timestamp int64) time.Duration`

fbsobreira (changes requested):

indexer/elasticProcessor.go
Beroni: ok

indexer/elasticProcessor.go
Beroni: ok

indexer/elasticProcessor.go
Beroni: ok

fbsobreira (approved):

RomuloSiebra (approved):

PR #29, opened 2026-02-07
[KLC-962] Add missing index field in PeerAccountToValidatorInfo

phcarneirobc:
This pull request update functions to include the validator index when converting peer accounts to `ValidatorInfo`, ensuring that the index is consistently available across different parts of the codebase.

Enhancements to validator information mapping:

* Added the `Index` field to the `ValidatorInfo` struct in `PeerAccountToValidatorInfo` functions in `core/bootstrap/process.go`, `core/kapp/validators/validators.go`, and `core/process/peer/process.go`, improving access to validator indices throughout the system. [[1]](diffhunk://#diff-6b1d5fa06345e922a623398f82ba5c5e18c3c48ef284d44e02ef53d80ad366ceR917) [[2]](diffhunk://#diff-e27f2a44ee2f77bfa7977883f5c05a2893c579d37bd1e2f8a41d90b9fd5e1d20R1147) [[3]](diffhunk://#diff-84b8c7690aa2e7182abc05771043c4960f1965c452f4b8d92a8c1eaa335ae45bR288)

### DIFF
--- a/core/bootstrap/process.go
+++ b/core/bootstrap/process.go
@@ -914,6 +914,7 @@ func (e *epochStartBootstrap) PeerAccountToValidatorInfo(peer state.PeerAccountH
 	return &state.ValidatorInfo{
 		OwnerAddress:                    peer.GetOwnerAddress(),
 		PublicKey:                       peer.GetBLSPublicKey(),
+		Index:                           peer.GetIndex(),
 		List:                            peer.GetListString(),
 		TempRating:                      peer.GetTempRating(),
 		Rating:                          peer.GetRating(),
--- a/core/bootstrap/process_test.go
+++ b/core/bootstrap/process_test.go
@@ -11,6 +11,7 @@ import (
 	consensusMock "github.com/klever-io/klever-go/core/consensus/mock"
 	cryptoMock "github.com/klever-io/klever-go/crypto/mock"
 	"github.com/klever-io/klever-go/data/retriever"
+	"github.com/klever-io/klever-go/data/state"
 	"github.com/klever-io/klever-go/sharding"
 	"github.com/klever-io/klever-go/storage"
 	"github.com/klever-io/klever-go/tools/check"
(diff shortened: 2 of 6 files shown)


fbsobreira (changes requested):
- All 3 test files follow a nearly identical pattern — same structure, same assertions, same sub-test names. Consider extracting a shared test helper like:
```
  func assertValidatorInfoMapsIndex(t *testing.T, info *state.ValidatorInfo, expected uint32) {
      t.Helper()
      assert.Equal(t, expected, info.Index, "Index should match PeerAccount.GetIndex()")
  }
```

- Minor: Sub-tests use sentence-case names ("Correctly maps Index from PeerAccount"). Go convention typically uses lowercase ("correctly maps index from peer account" or snake-style descriptions).

fbsobreira (changes requested):
The `ValidatorInfo` flows through this consensus-critical chain:

```
PeerAccountToValidatorInfo()                        ← THE CHANGE
  → GetValidatorInfoForRootHash()                   ← core/process/peer/process.go:379
    → block.go:301-332                              ← epoch start processing
      → SetEpochValidatorsInfo()                    ← sharding/nodesCoordinator.go:699
        → computeNodesConfigFromList()              ← nodesCoordinator.go:884
          → NewValidator(..., validatorInfo.Index)   ← nodesCoordinator.go:893
            → sort.Sort(validatorList(...))          ← nodesCoordinator.go:921-924
```

The sort comparator at `sharding/nodesCoordinator.go:40-45`:

```go
func (v validatorList) Less(i, j int) bool {
    if v[i].Index() == v[j].Index() {
        return bytes.Compare(v[i].PubKey(), v[j].PubKey()) < 0
    }
    return v[i].Index() < v[j].Index()
}
```

**Before this PR**: All validators have `Index=0`, so sorting falls through to `PubKey` comparison only.

**After this PR**: Validators have their actual `Index` values, so sorting is primarily by `Index`, with `PubKey` as tiebreaker.

### Why this does NOT fork on its own

The sort order change **only takes effect during `EpochStartPrepare`** (`nodesCoordinator.go:736`), which runs when processing an epoch start meta block. The key insight is:

1. The sorted lists determine **consensus group selection** for the *next* epoch
2. This happens at `EpochStartPrepare` → `computeNodesConfigFromList` → `sort.Sort`
3. **All nodes in a network must run the same code version** at epoch boundaries — this is already the deployment model

However, there is a **transitional risk** if nodes upgrade at different times within the same epoch:

| Scenario                                     | Result                                                |
| -------------------------------------------- | ----------------------------------------------------- |
| All nodes upgrade **before** epoch boundary  | No fork — all compute same sort order                 |
| All nodes upgrade **after** epoch boundary   | No fork — all still use Index=0                       |
| **Mixed versions** at epoch boundary         | **FORK** — nodes disagree on validator ordering       |

### Concrete example of a mixed-version fork

Assume 4 validators with these peer account states:

- Validator A: PubKey=`0xAA`, Index=3
- Validator B: PubKey=`0xBB`, Index=1
- Validator C: PubKey=`0xCC`, Index=2
- Validator D: PubKey=`0xDD`, Index=0

**Old node sort (Index=0 for all, fallback to PubKey):**

`[A(0xAA), B(0xBB), C(0xCC), D(0xDD)]`

**New node sort (by actual Index):**

`[D(idx=0), B(idx=1), C(idx=2), A(idx=3)]`

These different orderings feed into `UpdateNodeLists` → `SetNodes`, producing **different consensus group selections**. Nodes would reject each other's blocks.

the `Index` field should be populated. But it requires **coordinated deployment**:

1. **Must be deployed as part of a hard fork / coordinated upgrade** where all nodes activate the change at the same epoch boundary simultaneously
2. Alternatively, guard behind a **fork controller flag** (similar to the existing `vs.forkController.FixStakingBuckets()` at `core/process/peer/process.go:319`) so the change activates at a predetermined epoch

@nickgs1337 The cost of not adding it is low. It's a data fidelity issue, not a correctness issue. The network runs fine with Index=0. The only real concern is whether the validator ordering by position was intentionally designed and you're losing that behavior — but the network has been running without it, so it's clearly not critical today. We should take a decision here add in this coming fork or maybe not add at all...

nickgs1337:
As requested by @fbsobreira, we are moving this task back to backlog to further investigation.

PR #46, opened 2026-05-12
[KLC-2382] deflake TestConsensus_RevertBlockAndTransactions

fbsobreira:
## Summary

Three independent root causes were behind the intermittent failure of
`TestConsensus_RevertBlockAndTransactions` in CI. Fixing all three brings the
test to **44 consecutive passes** in a 50-iteration fresh-process loop at the
**mainnet 4 s slot budget**, up from a ~45% failure rate at the same budget
beforehand.

- **Production race in `delayedBlockBroadcaster.SetHeaderForValidator`**
  (`core/consensus/broadcast/delayedBroadcast.go`). It was the only function
  in the file that appended to `valHeaderBroadcastData` without acquiring
  `mutDataForBroadcast`, while `interceptedHeader` iterated the same slice
  under the lock. The unsynchronized append could let `interceptedHeader`
  cancel the wrong header alarm — validators then missed the leader header,
  failed to sign within the slot, and the cluster stalled below quorum
  (`worker statistics: small consensus quorum sigsNum=3`). Fixed by wrapping
  the slice mutation in `Lock`/`Unlock`, matching the pattern of
  `SetValidatorData`. This is the proximate cause of the CI flake.
- **Test-side data race on `SlotManagerMock.SlotIndex`**. Plain `int64` read
  by `baseBootstrap.GetNodeState` sync goroutines via `Index()`, written by
  the test goroutine via the `UpdateSlotCalled` closure. Converted to
  `atomic.Int64`; updated the five external call sites and promoted the
  test-local `currentSlot` to `atomic.Int64` for the same reason.
- **Silent timeouts hiding real failures**. `waitForBlockConditionOrTimeOut`
  used to `log.Error` and `return`, so the next assertion (`header.GetSlot()`)
  fired against an out-of-sync cluster — producing the original
  `expected 0x10 / actual 0xf` CI symptom. Now threads `*testing.T` and calls
  `t.Fatalf` with the per-node `nodesComplete` map at the point of
  divergence. The per-step budget is intentionally pinned to the mainnet
  slot duration (4 s) via `simulatedSlotDuration` / `blockWaitTimeoutSeconds`
  — this test is also a regression guard for production cadence and must not
  be relaxed.

## Verification

Local stress on macOS / Go 1.25.7 / M-series CPU, 50 fresh `go test` processes
at the original 4 s budget (CI-equivalent):

| | Pre-fix | Post-fix |
|---|---|---|
| Pass / fail | 5 / 4 (~45% fail) | **48 / 2 (96% pass)** |
| Max consecutive PASS | n/a | **44** (iters 7–50) |

The two failures in the post-fix loop both occurred within the first ~60 s after
killing a previous stress run; they show the same single-node-lag pattern and
are consistent with libp2p port `TIME_WAIT` residue rather than intrinsic flake.
AC of \"20+ consecutive runs\" is exceeded by 2×.

## Out of scope (follow-up needed)

Running with \`-race\` after the \`SetHeaderForValidator\` fix still surfaces a
**separate production race** on \`ConsensusState\` shared fields:

- Write at \`core/consensus/slot/consensusState.go:61\` (\`ResetConsensusState\`)
- Read at \`core/consensus/slot/consensusState.go:156\` (\`IsConsensusDataSet\`)

\`Data\`, \`Header\`, \`SlotCanceled\`, \`ExtendedCalled\`,
\`WaitingAllSignaturesTimeOut\` on \`ConsensusState\` lack a mutex and are
touched from many call sites across \`core/consensus/slot/...\`. CI does not
enable \`-race\`, so this does not affect KLC-2382 directly — but it is a real
production race and a follow-up Jira should be filed.

## Test plan

- [x] Build clean (\`go build ./...\`)
- [x] \`golangci-lint run\` on all changed packages — 0 issues attributable to
      these changes (2 pre-existing \`gosec\` findings in \`utils.go\` are on
      unrelated lines)
- [x] Single race-detector sanity (\`go test -race -count=1\`): test now fails
      cleanly via \`t.Fatalf\` on the remaining \`ConsensusState\` race, with
      the new diagnostic message instead of misleading downstream assertions
- [x] 50-iteration fresh-process stress at the original 4 s budget: 48/50 PASS,
      44 consecutive passes (iters 7–50)
- [ ] CI pipeline green on this PR

[Review Change Stack](https://app.coderabbit.ai/change-stack/klever-io/klever-go/pull/46)

### DIFF
--- a/core/consensus/broadcast/delayedBroadcast.go
+++ b/core/consensus/broadcast/delayedBroadcast.go
@@ -201,12 +201,6 @@ func (d *delayedBlockBroadcaster) SetHeaderForValidator(vData *validatorHeaderBr
 		return common.ErrNilHeaderHash
 	}
 
-	log.Trace("delayedBlockBroadcaster.SetHeaderForValidator",
-		"nbDelayedBroadcastData", len(d.delayedBroadcastData),
-		"nbValBroadcastData", len(d.valBroadcastData),
-		"nbValHeaderBroadcastData", len(d.valHeaderBroadcastData),
-	)
-
 	// skip alarm if the block was not finalized yet
 	if len(vData.header.GetSignature()) == 0 {
 		log.Trace("delayedBlockBroadcaster.SetHeaderForValidator: header alarm has not been set",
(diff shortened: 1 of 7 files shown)


fbsobreira:
Closing in favour of widened-scope PR — branch renamed to bugfix/KLC-2382-consensus-data-races to reflect the broader race-fix work uncovered during this investigation. New PR forthcoming on the renamed branch.

PR #66, opened 2026-05-25
[KLC-2394] feat: operator add ms sign subcommand to sign multisi

n4t4l:
## Summary

Add operator ms sign subcommand to sign a multisign transaction and post the signature to the multisign API, also supporting an interactive mode to choose between pending multisignature transactions.

## Problem

Previously the user would have to fetch, sign and post the signature manually with multiple commands, which was fragile.

## Solution

- Add operator ms sign to fetch a pending transaction, sign it, and post the signature to the multisign API.

- Support both a specific TX hash and interactive pending transaction selection.

- Consolidate helper logic for fetching, signing, encoding, and posting multisign data.

- Keep --multisign-api persistent on ms.

- Having no pending transactions returns a friendly message

- Trying to sign an already signed transaction returns a message indicating the no-op.

## Key Changes

-[cmd/operator/multisign.go](): add sign subcommand, interactive selection UI, and reusable helper flow.

-[cmd/operator/multsignHelper.go](): add getMSApiTransaction, doSignAndPost, and doPostMSTransaction; clarify API signature posting.

-[cmd/operator/main.go](): tighten command detection for operator sign and operator ms encode.

-Flags: moved --multisign-api to ms persistent flags;.

## Testing

- [x] cmd/operator existing tests pass

- [x] Added new tests

- [x] Manually tested reading updates to transaction's signature states both between addresses and from the same address. Also tested impeding the signing of a transaction multiple times from the same user.

## Configuration Changes

None

## Breaking Changes

None

## Related Issues

--

## Checklist

- [x] Documentation updated (added ms sign markdown file to cmd/operator/docs)

## Blockchain Impact Assessment

This PR is a CLI operator tooling enhancement with limited direct impact on core blockchain-critical components. It focuses on multisignature transaction workflow automation at the operator level.

### Transaction Signing Workflow Enhancement

Adds a new `operator ms sign` subcommand that integrates with the multisignature transaction workflow:
- Fetches pending multisignature transactions from a multisign API endpoint (by hash or via interactive selection from signer's pending list)
- Signs transactions using the operator's private key and multisignature protocol
- Posts signatures back to the multisign API with proper encoding

### Error Handling and Data Integrity

**Improved validation and duplicate-signing prevention:**
- Transaction hash validation (length and hex format checks) prevents invalid API requests
- Authorization checking ensures only required signers can sign a transaction
- **Duplicate signing prevention**: The code explicitly checks if `signer.Signed` is already true and returns a no-op success (line 173-175 in multsignHelper.go), preventing duplicate signatures from being submitted
- API response validation: signature posting only succeeds when the API returns `status: ok`; API-reported errors are propagated to the operator
- Graceful handling of edge cases (no pending transactions, already-signed transactions)

### API and Networking

Consolidates multisign API interactions into reusable helpers:
- `getMSApiTransaction()`: fetches and validates transaction data from the multisign API
- `doPostMSTransactionSignature()`: marshals and POSTs signatures with proper error handling
- `doSignAndPost()`: orchestrates the full signing workflow

### Scope Limitations

- **No consensus changes**: Multisignature coordination happens via external API, not node consensus
- **No state management impact**: Transaction state is managed by the multisign API
- **No core transaction processing changes**: Signing occurs in operator tool, not in node transaction processing
- **No concurrency concerns**: Single-threaded CLI execution model

[Review Change Stack](https://app.coderabbit.ai/change-stack/klever-io/klever-go/pull/66?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


cmd/operator/docs/operator_ms.md
```diff
@@ -0,0 +1,53 @@
+# operator ms
```
fbsobreira: This doc supposed to be auto generated, but I wee is kinda outdated... lets do this, do not change in this PR, lets defere into a new ticket, just run the ` GEN_DOCS=1 go run ./cmd/operator version`
@Duukaa

cmd/operator/multsignHelper.go
```diff
@@ -41,6 +41,86 @@ type MSApiTransaction struct {
 	Raw *transaction.Transaction `json:"raw"`
 }
 
+func getMSApiTransaction(hash string) (*MSApiTransaction, error) {
```
fbsobreira, line 243: since you are refactoring here would be good to have the unit test as this now have very specific test flow

cmd/operator/multsignHelper.go
```diff
@@ -41,6 +41,86 @@ type MSApiTransaction struct {
...
+		}
+		input = strings.ToLower(strings.TrimSpace(input))
+		if input != "y" {
+			fmt.Println("aborting")
+			return nil
+		}
+	}
+	return doPostMSTransaction(encoded)
+}
+func doPostMSTransaction(encoded *MSApiEncoded) error {
```
fbsobreira: same here, lets add the unit test

cmd/operator/multsignHelper.go
```diff
@@ -41,6 +41,86 @@ type MSApiTransaction struct {
...
+	err = utils.GetURL(fmt.Sprintf("%s/transaction/%s", multisignAPI, hash), &result)
+	if err != nil {
+		if err.Error() == "EOF" {
+			return nil, fmt.Errorf("transaction with hash %s not found in multisign API", hash)
+		}
+		return nil, err
+	}
+	return &result, nil
+}
+func doSignAndPost(tx *MSApiTransaction) error {
```
fbsobreira, line 263: same here, add unit test

cmd/operator/multisign.go
```diff
@@ -184,15 +151,127 @@ func subMS() []*cobra.Command {
 			return DumpAsJson(result)
 		},
 	}
-	cmdMultisignByAddress.Flags().StringVar(&multisignAPI, "multisign-api", "https://multisign.mainnet.klever.org", "multisign API URL")
+	cmdMultiSignAndPost := &cobra.Command{
+		Use:   "sign [txHash]",
+		Short: "sign and broadcast a transaction from multisign API",
```
fbsobreira: `sign and broadcast` this should only sign and post to multisign API, not broadcast to network

cmd/operator/multsignHelper.go
```diff
@@ -41,6 +41,86 @@ type MSApiTransaction struct {
...
+				fmt.Printf("Transaction %s has already been signed by %s \n", tx.Hash, signerAddress)
+				return nil
+			}
+			authorized = true
+		}
+	}
+	if !authorized {
+		return fmt.Errorf("signer %s is not required to sign transaction %s", signerAddress, tx.Hash)
+	}
+	_, err := SignTX(tx.Raw)
```
fbsobreira: should only sign after approval

cmd/operator/multsignHelper.go
```diff
@@ -41,6 +41,86 @@ type MSApiTransaction struct {
...
+	if err != nil {
+		return err
+	}
+
+	encoded, err := encodeMSApiData(tx.Raw)
+	if err != nil {
+		return err
+	}
+	if !multisignYes {
```
fbsobreira: we already have a auto sign flag check for `shouldSign(tx)`

cmd/operator/multsignHelper.go
```diff
@@ -41,6 +41,86 @@ type MSApiTransaction struct {
...
+	data, err := json.Marshal(encoded)
+	if err != nil {
+		return err
+	}
+
+	broadcastResult := struct {
+		Status string `json:"status"`
+		Error  string `json:"error"`
+	}{}
+	log.Info("broadcasting...")
```
fbsobreira: its not broadcast, we are sending to multisign API

cmd/operator/multisign.go
```diff
@@ -205,6 +284,7 @@ func init() {
 		},
 	}
 	cmdMS.AddCommand(subMS()...)
-
+	cmdMS.PersistentFlags().StringVar(&multisignAPI, "multisign-api", "https://multisign.mainnet.klever.org", "multisign API URL")
+	cmdMS.PersistentFlags().BoolVarP(&multisignYes, "yes", "y", false, "skip the confirmation prompt; useful for scripts and non-interactive use cases")
```
fbsobreira: we already have a auto sign flag

cmd/operator/multisign.go
```diff
@@ -184,15 +151,127 @@ func subMS() []*cobra.Command {
...
+				if err != nil {
+					return err
+				}
+				err = doSignAndPost(tx)
+				if err != nil {
+					return err
+				}
+				return nil
+			}
+			log.Info("fetching pending transactions for signing")
```
fbsobreira: should add this flow to a sub-function reduce cognitive complexity

fbsobreira (changes requested):

cmd/operator/multisign.go
n4t4l: The use of "broadcast" here and later in doPostMsTransaction is leftover from the previous implementation; While "Sign and post transaction" is objectively better than the current verbiage, could i suggest the use of "append signature" when referencing this operation? IMHO "sign and post" could be confused as "sign and execute" or "sign and create", and "post" alone also has the same drawbacks.

cmd/operator/multsignHelper.go
n4t4l: shouldSign returns an error, should we bubble it up here or eat it since it's actually the correct behavior to not sign on user cancel? maybe also do the friendly message?

fbsobreira (changes requested):
you should not update docs now, way too many changes, create a new ticket to be address later...

cmd/operator/multisign.go
```diff
@@ -184,15 +149,51 @@ func subMS() []*cobra.Command {
...
+			log.Info("signing transaction from multisign API", "address", signerAddress)
+			var tx *MSApiTransaction
+			var err error
+			if len(args) == 1 {
+				tx, err = getMSApiTransaction(args[0])
+			} else {
+				log.Info("fetching pending transactions for signing")
+				result := make([]MSApiTransaction, 0)
+				err = utils.GetURL(fmt.Sprintf("%s/transaction/by-address/%s", multisignAPI, signerAddress), &result)
+				if err != nil && err.Error() != "EOF" {
```
fbsobreira: 🟡 **Warning (R2): Fragile EOF handling + questionable API contract**

`if err != nil && err.Error() != "EOF"` matches the error *text* instead of the sentinel. It works only because `utils.GetURL` returns the raw `json.Decoder.Decode` error unwrapped and `io.EOF.Error()` happens to be `"EOF"` — any future wrapping silently breaks it.

More importantly, **relying on an empty body (→ `EOF`) to mean "no pending transactions" is a smell on the API side**: a no-results response should be an empty JSON array `[]` (or a proper status code), not an empty body. `GetURL` also ignores the HTTP status, so an empty-body `5xx` is indistinguishable from "no transactions" and is silently treated as success.

**Recommendation**
- Client: use `errors.Is(err, io.EOF)` (consistent with the recent KLC-2355 sentinel work) and surface non-2xx status in `GetURL`.
- Server/API: please confirm the `by-address` endpoint returns `[]` for no results rather than an empty body.

*Ref: CWE-396 — fragile error classification*

cmd/operator/multsignHelper.go
```diff
@@ -27,20 +29,193 @@ type MSApiEncoded struct {
...
+	}
+	_, err := hex.DecodeString(hash)
+	if len(hash) != 64 || err != nil {
+		return nil, fmt.Errorf("invalid TX hash %s", hash)
+	}
+
+	result := MSApiTransaction{}
+	err = utils.GetURL(fmt.Sprintf("%s/transaction/%s", multisignAPI, hash), &result)
+	if err != nil {
+		if err.Error() == "EOF" {
```
fbsobreira: 🟡 **Warning (S3): Same EOF-as-control-flow in `getMSApiTransaction`**

`if err.Error() == "EOF"` translates an empty body into "transaction not found". Same fragility as R2 (string compare vs `errors.Is(err, io.EOF)`), and the same API-contract concern: **a not-found transaction should return HTTP 404 / a structured error, not an empty body** that the JSON decoder reports as `EOF`. Today a genuine server error with an empty body would be misreported as "not found".

**Recommendation** — Use `errors.Is(err, io.EOF)`, and verify the multisign API signals not-found via status code. If the API is fixed to return proper statuses, this `EOF` special-casing can be removed entirely.

cmd/operator/multisign_test.go
```diff
@@ -0,0 +1,181 @@
...
+	tx := &MSApiTransaction{
+		Hash:    validHash,
+		Signers: []MSApiSigner{{Address: "addr_A", Signed: false}},
+		Raw: &transaction.Transaction{
+			Signature: make([][]byte, 0),
+			RawData:   &transaction.Transaction_Raw{},
+		},
+	}
+	err := doSignAndPost(tx)
+	assert.NoError(t, err)
```
fbsobreira, line 156: 🟡 **Warning (R1): Test passes on a malformed payload**

`TestDoSignAndPost_AutoSign_Success` builds a transaction with an empty `Sender` (`RawData: &transaction.Transaction_Raw{}`). Inside `encodeMSApiData`, `walletPubKeyConverter.Encode` returns `""` (it only `log.Warn`s on wrong-length input — see `bech32PubkeyConverter.Encode`) and **propagates no error**, so the flow posts an `MSApiEncoded` with an empty `Address` — yet the test still passes because it only asserts `assert.NoError(t, err)`. It green-lights a broken result instead of proving a correct sign-and-post.

**Recommendation** — Capture the request body on the test HTTP server and assert (a) the posted `Address` is the expected non-empty bech32 address and (b) a signature was appended to `tx.Raw.Signature`.

cmd/operator/multisign.go
```diff
@@ -184,15 +149,51 @@ func subMS() []*cobra.Command {
 			return DumpAsJson(result)
 		},
 	}
-	cmdMultisignByAddress.Flags().StringVar(&multisignAPI, "multisign-api", "https://multisign.mainnet.klever.org", "multisign API URL")
-
+	cmdMultiSignAndPost := &cobra.Command{
+		Use:   "sign [txHash]",
+		Short: "sign a transaction from multisign API and post the signature",
+		Example: `operator ms sign <txHash>   — sign specific TX by hash
```
fbsobreira, line 157: 💡 **Suggestion (R3): `--yes/-y` flag in the PR description does not exist**

The PR description mentions "add `--yes/-y` to skip confirmation", but no such flag is registered anywhere. Confirmation-skipping for `ms sign` works only via the pre-existing global `--sign/-s` (`autoSign`), which `shouldSign` honors.

**Recommendation** — Correct the PR description to reference `--sign/-s`, and consider extending this command's `Example` to show the non-interactive form, e.g. `operator ms sign <txHash> -s`, so the auto-confirm path is discoverable.

fbsobreira (commented):
## Automated Code Review

Reviewed by Claude Code — 4 findings (3 warnings, 1 suggestion).

Legend: 🔴 Critical | 🟡 Warning | 💡 Suggestion | 🟢 Praise

cmd/operator/multsignHelper.go
```diff
@@ -27,20 +30,193 @@ type MSApiEncoded struct {
...
+		}
+	}
+	if !authorized {
+		return fmt.Errorf("signer %s is not required to sign transaction %s", signerAddress, tx.Hash)
+	}
+	err := shouldSign(tx.Raw)
+	if err != nil {
+		return err
+	}
+	_, err = SignTX(tx.Raw)
```
fbsobreira, line 281: 🔴 **Critical** (S1): Signing API-returned data without binding it to an independent reference (blind signing)

`doSignAndPost` signs `tx.Raw` (`SignTX(tx.Raw)`), where the full `Raw *transaction.Transaction` and the `Signers` list both come from the multisign API.

Note that the hash *is* computed during the flow — three times, in fact: `SignTX` → `computeTxHash`, `formatAndDumpRawTX` (the confirmation view), and `encodeMSApiData` (the posted payload). But all three re-derive the hash from the **same API-supplied `tx.Raw.RawData`** and it is never compared against an independent reference (the hash the operator typed, or any out-of-band value). Re-hashing attacker-supplied bytes and signing the result provides no integrity guarantee — a compromised/malicious API can return a `Raw` whose contents differ from what the operator intended, and the operator's ed25519 key signs it.

**Severity depends on the path:**

- **By-hash** (`operator ms sign <txHash>`): the operator supplies a hash they trust from an out-of-band source, so there *is* a reference to check against. This path should verify it.
- **Interactive** (`operator ms sign`): the operator picks from the API's own list, so there is no operator-supplied hash to compare — the only safeguard is the human reviewing the decoded transaction in `shouldSign`/`formatAndDumpRawTX`.
- **`-s` / `--sign`**: `shouldSign` returns immediately with no display, removing the only safeguard on *both* paths → genuinely blind signing.

**Recommendation:**

1. **By-hash path** — compare `computeTxHash(tx.Raw)` against the operator-supplied hash (`args[0]`), **not** against `tx.Hash` (which the API also controls, so a fully malicious API sets both to match). Run this unconditionally, before signing — i.e. not inside `shouldSign`, which `-s` short-circuits:
```go
want := strings.TrimPrefix(args[0], "0x")
got, err := computeTxHash(tx.Raw)
if err != nil { return err }
if hex.EncodeToString(got) != want {
    return fmt.Errorf("multisign API returned tx whose raw hash %x != requested %s", got, want)
}
```
2. **Interactive path** — there is no out-of-band reference, so don't let `-s` skip the decoded-content review here; require the human confirmation (or an explicit extra acknowledgement) before signing a tx chosen from the API list.

Net: treat the API as untrusted transport — the correct posture for a signing tool.

*Ref: CWE-345 (Insufficient Verification of Data Authenticity)*

cmd/operator/main.go
```diff
@@ -99,8 +99,8 @@ var rootCmd = &cobra.Command{
 		}
 
 		if createOnly ||
-			strings.HasPrefix(cmd.Use, "sign") ||
-			strings.HasPrefix(cmd.Use, "encode") {
+			cmd.CommandPath() == "operator sign" ||
```
fbsobreira, line 102: 🟡 **Warning** (R1): `operator ms sign` silently dropped from the log-silencing gate

The gate changed from `strings.HasPrefix(cmd.Use, "sign")` to exact `cmd.CommandPath()` matching. The old prefix form also matched the new subcommand (its `Use` is `"sign [txHash]"`), so it would have silenced logs; the new exact list (`"operator sign"`, `"operator ms encode"`) deliberately does **not** include `operator ms sign`. As a result `log.Info(...)` from the `sign` `RunE`, `doSignAndPost`, and `doPostMSTransactionSignature` now write interleaved with the interactive picker's `fmt.Print` table and the `shouldSign` prompt, and `ms sign <hash>` output is no longer clean/pipeable the way `ms encode` is.

**Recommendation**: Make this a deliberate decision and document it. If `ms sign` output should be machine-parseable, add `"operator ms sign"` to the silenced list (at least when a hash arg is supplied); otherwise confirm the logger writes to stderr so stdout stays clean, and add a comment noting the asymmetry with `ms encode` is intentional.

cmd/operator/multsignHelper.go
```diff
@@ -27,20 +30,193 @@ type MSApiEncoded struct {
...
+	if currPage > 0 {
+		ln += ", [p]revious page"
+	}
+	if currPage < pagesCount-1 {
+		ln += ", [n]ext page"
+	}
+	ln += ", [q]uit \n Input: "
+	return ln, nil
+}
+func pendingMsTransactionPicker(txs *[]MSApiTransaction) (*MSApiTransaction, error) {
```
fbsobreira: 🟡 **Warning** (R2): Picker silently ignores invalid input, and the index math is untested

In the `for {}` loop, the `switch` handles `0..n-1`, `n`, `p`, `q`. Any other input — a non-numeric key, an out-of-range number, `n` on the last page, `p` on the first page — matches no case and falls through to the ANSI redraw + `continue` with **no feedback**, which reads as the tool hanging. Separately, this is the most complex function in the PR (cognitive 16 / cyclomatic 13) and the index arithmetic `idx+(currPage*perPage)` is at **0% coverage**.

**Recommendation**:
1. Add a `default:` arm to the switch that sets an "invalid selection" message rendered on the next `buildPageString` redraw.
2. Extract the pure decision `(input, currPage, pageCount) → {selectIndex | next | prev | quit | invalid}` into its own function so the index math is unit-testable without stdin — this also drops the I/O loop's complexity below threshold.

fbsobreira (changes requested):
## Automated Code Review

Reviewed by Claude Code (code-review + security-auditor + code-metrics) — 3 inline findings (1 critical, 2 warnings).

Nice work on the consolidation and on prioritizing tests for the security-relevant signing branches (`not-authorized`, `already-signed`, auto-sign success). The idempotent re-sign guard and the consistent key/identity derivation (`signerAddress` and `privateKey` both from `LoadKey`) are solid.

One blocking item (**S1** — blind signing). A fourth, pre-existing finding (`utils.GetURL` ignores HTTP status codes) is out of this PR's scope and is tracked separately in **[KLC-2448](https://klever-io.atlassian.net/browse/KLC-2448)**.

Legend: 🔴 Critical | 🟡 Warning | 💡 Suggestion

cmd/operator/main.go
n4t4l: all of the cited methods run after the picker has finished, this doesnt really apply.

fbsobreira (approved):

PR #67, opened 2026-05-29
[KLC-2388] Gate klv_node_type / klv_peer_type writes on cache miss

nickgs1337:
On validator startup, the heartbeat sender's `updateMetrics` was overwriting the correct startup-time `klv_node_type=validator` with `observer`, because `peerTypeProvider.ComputeForPubKey` returns `ObserverList` silently on cache miss and the cache isn't populated until the first epoch-start event fires. Same root cause hit `klv_peer_type`. Operators and monitoring tooling saw the wrong node type for the entire bootstrap window — minutes to hours depending on epoch length.

The fix adds an `IsCachePopulated()` readiness gate on `PeerTypeProvider` that flips true only when the epoch-start handler fires. `Sender.updateMetrics` early-returns when the gate is closed, preserving the startup-init values. A fourth commit also seeds `klv_peer_type` at startup alongside `klv_node_type`, so consumers never observe an empty/absent field during the bootstrap window.

## What changed

4 atomic commits:
- `Add IsCachePopulated readiness gate to PeerTypeProvider` — new `isReady` field flipped inside the existing epoch-start handler.
- `Expose IsCachePopulated on PeerTypeProviderHandler interface` — wired through the heartbeat-side interface + mock stub.
- `Skip klv_node_type and klv_peer_type writes on cache miss` — early-return in `Sender.updateMetrics` + sender tests via an `UpdateMetrics` exporter.
- `Initialize klv_peer_type at startup alongside klv_node_type` — one-line seed of `ObserverList` so the field is never empty pre-gate.

## Validation

1. **Unit tests** (4 new): `IsCachePopulated` false at construction, flips true after epoch-start event; `updateMetrics` skips writes when not ready, writes correct values when ready. All green via `go test ./node/heartbeat/process/... ./core/process/peer/...`.

2. **Binary verification**: extracted the validator binary from the docker image and confirmed via `go tool nm` that `IsCachePopulated` + `markReady` symbols compile in, and via `go tool objdump -s 'Sender.*updateMetrics'` that the gate is actually called at `sender.go:177` *before* `computePeerList` at `sender.go:181`.

3. **In-the-wild observation**: during this sprint's KLC-1920 multi-validator harness work, captured the live bug on the fallback container reporting `klv_node_type=observer` despite `klv_redundancy_level=2` — independent pre-fix proof.

4. **End-to-end Docker A/B/C/D runs** on a 3-validator + 1-redundancy-2-fallback localnet (kleverchain-localnet harness, both images built from the same Klever HEAD):
   - **Run A** — pre-fix baseline, \`slotsPerEpoch=300\`: fallback flips to \`observer\` at T+25s — KLC-2388 bug reproduces cleanly.
   - **Run B** — patched gate-only, same config: fallback stays at \`validator\` for 4+ minutes through 50+ post-genesis slots; \`klv_peer_type\` empty (gate holds).
   - **Run C** — patched gate-only, \`slotsPerEpoch=20\`: captures the gate *opening* at slot 26 when the first epoch-start event fires; transitions to legitimate values (\`validator/elected\` for primaries, \`observer/observer\` for the fallback's observer-key heartbeat).
   - **Run D** — patched + \`klv_peer_type\` startup-init, \`slotsPerEpoch=20\`: confirms \`peer_type=observer\` from t=0 (vs empty in B/C), held through gate-closed window, transitions cleanly at gate open. No dangling-dash TUI artifact, no empty Prometheus label.

5. **Side-effect audit**: grep'd every reader of `MetricNodeType` / `MetricPeerType` in the codebase. Only consumers are the statusHandler presenter (TUI dashboard) and the `/node/status` + Prometheus exposers. No consensus, sync, fork-detection, heartbeat-message, or block-processing code branches on these values. Fix has no internal impact beyond observability.

## Notes

The KLC-2388 bug requires the `peerTypeProvider` cache to be unpopulated when the heartbeat first ticks (a production peer-churn condition). On localnet the cache populates correctly at construction, so the A/B runs deliberately exercise the cache-not-yet-populated codepath. The gate is correct regardless of whether the underlying cache-miss is reproducible in any specific environment.

One build-system gotcha worth flagging for reviewers: the first two patched docker builds produced a binary where the \`IsCachePopulated\` symbol existed but \`updateMetrics\` wasn't calling it — BuildKit cache mounts (\`--mount=type=cache,target=/root/.cache/go-build\`) persist across \`docker build --no-cache\` and reused a stale \`.a\` for \`node/heartbeat/process\`. Dropping the cache mounts temporarily for the rebuild defeated it; the upstream Dockerfile is unchanged in this PR.

This PR fixes a startup observability bug by preventing the heartbeat sender from emitting incorrect peer-type metrics during node bootstrap. It gates metric writes on a peer-type cache readiness flag and seeds a bootstrap peer-type value to avoid empty metrics before the cache is populated.

Impact assessment
- Affected components:
  - Heartbeat sender metrics (node/heartbeat/process/sender.go / Sender.updateMetrics)
  - Peer-type provider cache/state and readiness API (core/process/peer/peerTypeProvider.go)
  - Observability surface: TUI, /node/status, Prometheus exporter (metrics and status codepaths)
- Blockchain-critical components: none. The change is confined to observability and the peer-type cache readiness API. It does not change consensus, transaction processing, state management, KVM, or networking decision logic.
- Node stability and data integrity: unchanged. No on-chain state, consensus flows, or persistent node state are modified. The gating only prevents emission of misleading metric labels during bootstrap.

Key changes
- Readiness gate
  - PeerTypeProvider gains an isReady field and an exported IsCachePopulated() bool (protected by RW locks). isReady is set inside existing cache-refresh/updateCache logic (including epoch-start refreshes) when a non-empty cache is produced.
- Interface and test stubs
  - PeerTypeProviderHandler interface now includes IsCachePopulated(); mocks/stubs updated to support test control.
- Metric gating and bootstrap seeding
  - Sender.updateMetrics early-returns when IsCachePopulated() is false, skipping ComputeForPubKey and writes of klv_node_type / klv_peer_type to avoid falsely writing "observer" on cache miss.
  - InitMetrics seeds core.MetricPeerType at startup (ObserverList placeholder) alongside core.MetricNodeType so klv_peer_type is not empty before the gate opens.
- Concurrency and error handling
  - isReady is read/written under locks to ensure safe concurrent access by the heartbeat sender.
  - The gate prevents fallback-to-observer behavior that previously occurred when ComputeForPubKey ran on cache miss or error during bootstrap.

Validation
- Unit tests: added coverage for IsCachePopulated transitions (construction and epoch-start) and for Sender.UpdateMetrics behavior (skips writes when not ready; writes correct values when ready). Tests passed for relevant packages.
- Binary inspection: go tool nm/objdump verified IsCachePopulated symbol and that updateMetrics checks the gate before computing peer lists.
- Local harness: Docker runs reproduced the pre-fix misreporting; the gate + startup seed preserved correct metrics through bootstrap and transitioned correctly when the cache populated.

Notes
- A stale Docker build-cache produced a binary where updateMetrics didn't call the new gate; rebuilding without cache resolved it.

[Review Change Stack](https://app.coderabbit.ai/change-stack/klever-io/klever-go/pull/67?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cmd/node/metrics/metrics.go
+++ b/cmd/node/metrics/metrics.go
@@ -45,6 +45,12 @@ func InitMetrics(
 
 	appStatusHandler.SetStringValue(core.MetricPublicKeyBlockSign, pubkeyStr)
 	appStatusHandler.SetStringValue(core.MetricNodeType, string(nodeType))
+	// KLC-2388: seed klv_peer_type alongside klv_node_type so consumers see a
+	// real value during the bootstrap window before the peerTypeProvider cache
+	// is populated. The heartbeat sender's IsCachePopulated gate keeps this
+	// startup value in place until the first epoch-start event refreshes the
+	// cache, at which point the real peer-list classification takes over.
+	appStatusHandler.SetStringValue(core.MetricPeerType, string(core.ObserverList))
 	appStatusHandler.SetUInt64Value(core.MetricSlotTime, slotInterval/millisecondsInSecond)
 	appStatusHandler.SetStringValue(core.MetricAppVersion, version)
 	appStatusHandler.SetUInt64Value(core.MetricSlotsPerEpoch, uint64(slotsPerEpoch))
--- a/cmd/node/metrics/metrics.go
+++ b/cmd/node/metrics/metrics.go
@@ -44,12 +44,16 @@ func InitMetrics(
 	initZeroString := "0"
 
 	appStatusHandler.SetStringValue(core.MetricPublicKeyBlockSign, pubkeyStr)
+	// KLC-2388: node_type stays "validator" during the bootstrap window even
+	// though peer_type below is seeded to "observer". The two intentionally
+	// diverge until the peerTypeProvider cache populates: node_type is the
+	// PR's protected value (the original bug was validators flipping to
+	// observer on cache miss), so we cannot seed it observer without
+	// re-introducing exactly that symptom; peer_type is left pessimistic
+	// because it's a peer-list classification that we genuinely don't know
+	// yet. Once the gate opens the sender writes the real values from the
+	// fork detector / nodes coordinator.
 	appStatusHandler.SetStringValue(core.MetricNodeType, string(nodeType))
-	// KLC-2388: seed klv_peer_type alongside klv_node_type so consumers see a
-	// real value during the bootstrap window before the peerTypeProvider cache
-	// is populated. The heartbeat sender's IsCachePopulated gate keeps this
-	// startup value in place until the first epoch-start event refreshes the
-	// cache, at which point the real peer-list classification takes over.
 	appStatusHandler.SetStringValue(core.MetricPeerType, string(core.ObserverList))
 	appStatusHandler.SetUInt64Value(core.MetricSlotTime, slotInterval/millisecondsInSecond)
 	appStatusHandler.SetStringValue(core.MetricAppVersion, version)
--- a/core/process/peer/peerTypeProvider.go
+++ b/core/process/peer/peerTypeProvider.go
@@ -55,10 +55,10 @@ func NewPeerTypeProvider(arg ArgPeerTypeProvider) (*PeerTypeProvider, error) {
 	return ptp, nil
 }
 
-// IsCachePopulated returns true once the cache has been refreshed by an epoch-start event.
-// During the bootstrap window before the first epoch-start notification fires, the cache
-// is either empty or seeded from a NodesCoordinator that is not yet fully populated, so
-// callers should treat its peer-type result as unreliable.
+// IsCachePopulated returns true once at least one updateCache call has produced a
+// non-empty cache. Until then, the NodesCoordinator has not yet exposed any
+// validator keys for the active epoch, so callers should treat the peer-type result
+// as unreliable and avoid overwriting startup-seeded values.
 func (ptp *PeerTypeProvider) IsCachePopulated() bool {
 	ptp.mutCache.RLock()
 	defer ptp.mutCache.RUnlock()
@@ -103,7 +103,6 @@ func (ptp *PeerTypeProvider) epochStartEventHandler() sharding.EpochStartActionH
 				"slot", hdr.GetSlot(),
 				"epoch", hdr.GetEpoch())
 			ptp.updateCache(hdr.GetEpoch())
-			ptp.markReady()
 		},
 		func(_ data.HeaderHandler) {},
 		core.IndexerOrder,
@@ -117,12 +116,9 @@ func (ptp *PeerTypeProvider) updateCache(epoch uint32) {
 
 	ptp.mutCache.Lock()
 	ptp.cache = newCache
-	ptp.mutCache.Unlock()
-}
-
-func (ptp *PeerTypeProvider) markReady() {
-	ptp.mutCache.Lock()
-	ptp.isReady = true
+	if len(newCache) > 0 {
+		ptp.isReady = true
+	}
 	ptp.mutCache.Unlock()
 }
 
--- a/core/process/peer/peerTypeProvider.go
+++ b/core/process/peer/peerTypeProvider.go
@@ -23,6 +23,7 @@ type PeerTypeProvider struct {
 	nodesCoordinator sharding.NodesCoordinator
 	cache            map[string]*peerListAndShard
 	mutCache         sync.RWMutex
+	isReady          bool
 }
 
 // ArgPeerTypeProvider contains all parameters needed for creating a PeerTypeProvider
@@ -52,6 +53,16 @@ func NewPeerTypeProvider(arg ArgPeerTypeProvider) (*PeerTypeProvider, error) {
 	arg.EpochStartEventNotifier.RegisterHandler(ptp.epochStartEventHandler())
 
 	return ptp, nil
+}
+
+// IsCachePopulated returns true once the cache has been refreshed by an epoch-start event.
+// During the bootstrap window before the first epoch-start notification fires, the cache
+// is either empty or seeded from a NodesCoordinator that is not yet fully populated, so
+// callers should treat its peer-type result as unreliable.
+func (ptp *PeerTypeProvider) IsCachePopulated() bool {
+	ptp.mutCache.RLock()
+	defer ptp.mutCache.RUnlock()
+	return ptp.isReady
 }
 
 // ComputeForPubKey returns the peer type for a given public key and shard id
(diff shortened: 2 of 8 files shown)


core/process/peer/peerTypeProvider.go
```diff
@@ -92,6 +103,7 @@ func (ptp *PeerTypeProvider) epochStartEventHandler() sharding.EpochStartActionH
 				"slot", hdr.GetSlot(),
 				"epoch", hdr.GetEpoch())
 			ptp.updateCache(hdr.GetEpoch())
+			ptp.markReady()
```
fbsobreira: `markReady()` only runs from the epoch-start handler, so readiness is gated on *"an epoch-start event has been observed"* rather than *"the cache is populated"*. The constructor already populates the cache via `updateCache(arg.StartEpoch)` (line 50) without flipping the flag.

On a mid-epoch reboot the `NodesCoordinator` is already restored, so that constructor build should be correct and `ComputeForPubKey` returns the real type immediately — yet `IsCachePopulated()` stays `false` until the next epoch-start fires, and `Sender.updateMetrics` early-returns the whole time. The node would keep advertising the seeded `klv_node_type=observer` even though it's an eligible/elected validator, for up to ~a full epoch.

Quick repro against this branch — correct data in hand, gate still closed:

```
FACT (a): ComputeForPubKey(validatorKey) at construction = "eligible"
FACT (b): IsCachePopulated() at construction = false
after epoch-start event: IsCachePopulated() = true   // classification unchanged throughout
```

Can you verify/validate this mid-epoch-reboot path? Specifically: is the `NodesCoordinator` actually populated at the point `NewPeerTypeProvider` runs `updateCache(arg.StartEpoch)` during a real restart? If it is, the gate is suppressing *correct* data (the inverse of the bug we're fixing) and we'd want to mark ready once the cache is genuinely populated — e.g. `markReady()` at the end of the constructor's initial `updateCache`, or gate on a successful coordinator fetch / `len(cache) > 0` instead of the epoch event. If the coordinator is *not* populated at that point, then the current gate is correct and it'd be worth a comment saying so.

Observability-only either way (nothing in consensus/sync/fork reads these metrics), so non-blocking — but worth confirming before merge.

fbsobreira (commented):
One comment on the readiness gate — needs verification: the metric suppression is correct on cold start, but on a mid-epoch reboot where the cache is already valid it may suppress correct data. Please validate whether the coordinator is populated at construction time. Observability-only, non-blocking. Concurrency on `isReady` and the test coverage look good.

core/process/peer/peerTypeProvider.go
nickgs1337: Good catch, you're right. Pushed a fix — flips ready inline when the cache lands non-empty (from the constructor's refresh or any later one). All three primaries now show `elected` within epoch 0 instead of waiting for the first epoch event.

cmd/node/metrics/metrics.go
```diff
@@ -45,6 +45,12 @@ func InitMetrics(
 
 	appStatusHandler.SetStringValue(core.MetricPublicKeyBlockSign, pubkeyStr)
 	appStatusHandler.SetStringValue(core.MetricNodeType, string(nodeType))
+	// KLC-2388: seed klv_peer_type alongside klv_node_type so consumers see a
+	// real value during the bootstrap window before the peerTypeProvider cache
+	// is populated. The heartbeat sender's IsCachePopulated gate keeps this
+	// startup value in place until the first epoch-start event refreshes the
+	// cache, at which point the real peer-list classification takes over.
+	appStatusHandler.SetStringValue(core.MetricPeerType, string(core.ObserverList))
```
fbsobreira, line 57: One thing I want to check before this goes in: seeding `peer_type=observer` here is fine, but `node_type` is still seeded to `validator` up at line 47 (it's hardcoded to `NodeTypeValidator` in `startup.go`). Once we gate the sender, both seeds get frozen until the cache populates — so during the whole bootstrap window the node publishes `node_type=validator` + `peer_type=observer`, which contradict each other.

The part that worries me more than the cosmetics: on `develop` the sender corrected a bad `node_type` on the first heartbeat, so an **observer** node only showed `validator` for a second or two. With the gate, that wrong `validator` now sticks for the entire bootstrap window until the first cache refresh. So we fix "validator briefly shown as observer" but introduce "observer shown as validator for longer."

Is keeping `node_type=validator` stable the intent here, or should bootstrap be conservative? Two ways to square it:
- seed both to `observer` (matches what the sender derives from `ObserverList` anyway — `ObserverList → NodeTypeObserver`), so observers are correct from t=0 and validators flip up once the cache loads; or
- keep `node_type` as-is but drop a comment that `peer_type` is intentionally pessimistic and the two are expected to diverge until the cache is ready.

Either way it's a one-liner — just want to make sure we pick deliberately since this is the exact metric the PR is meant to make trustworthy.

core/process/peer/peerTypeProvider.go
```diff
@@ -105,6 +116,9 @@ func (ptp *PeerTypeProvider) updateCache(epoch uint32) {
 
 	ptp.mutCache.Lock()
 	ptp.cache = newCache
+	if len(newCache) > 0 {
```
fbsobreira, line 119: Not a blocker for this PR — more a note for whoever touches this next. The `isReady` latch only ever flips `false→true`, but `updateCache` still swaps in the new map unconditionally, and `createNewCache` swallows the coordinator errors (logs at Debug and carries on with a nil slice). So the cache can go populated→empty on a later epoch refresh while `IsCachePopulated()` keeps returning true, and in that window the sender reclassifies a real validator as observer again.

To be clear this isn't something you introduced — on `develop` the sender already ran unconditionally, so that empty-cache mislabel was always possible at epoch boundaries; the new gate just doesn't extend its protection there. So nothing to fix here for the PR's scope.

If we ever want to close it though, the cheap option is to not clobber a good cache with an empty one:
```go
ptp.mutCache.Lock()
if len(newCache) > 0 {
    ptp.cache = newCache
    ptp.isReady = true
}
ptp.mutCache.Unlock()
```
which keeps the last-known-good classification through a transient empty refresh. Fine to leave as-is for now — just flagging so the latch's sticky semantics are on record.

fbsobreira (changes requested):

cmd/node/metrics/metrics.go
nickgs1337: Good catch on picking deliberately. Seeding both to `observer` re-introduces the exact symptom this PR is for (validators briefly showing observer on cache miss), so going the other way: keep `node_type=validator`, accept that observers briefly show validator in the bootstrap window. Pushed a comment in metrics.go making the choice explicit.

core/process/peer/peerTypeProvider.go
nickgs1337: Yeah, sticky-true on a transient empty refresh is the gap. Your `len(newCache) > 0` guard is the clean way to close it. Leaving for a follow-up per your call.

PR #76, opened 2026-06-15
[KLC-2440] Bound /subscribe WebSocket reads, connections, and subscriptions (GHSA-4fwh-wrm6-97xm)

fbsobreira:
## Summary

Hardens the unauthenticated `/subscribe` WebSocket endpoint, which had no resource bounds. Because the REST/WS API runs in-process with the node, unbounded reads/connections/subscriptions are an unauthenticated remote resource-exhaustion vector (CWE-770).

## Changes

- **Inbound read limit** on both read paths (handshake `processSubscription` and `client.loopIn`), plus a **lifetime read deadline** refreshed by a server ping ticker — passive-but-live subscribers stay connected while dead/idle ones are reclaimed at `pongWait`.
- **Address caps** per subscribe call and per connection, rejected before mutating hub state and counting distinct new addresses once. The read limit is **derived** from the per-subscribe cap, so a maximal subscribe always fits while staying bounded (it is never a separate, weaker-able knob).
- **Map-leak fix**: `handleClientDelete`/`deleteAll` now reclaim empty `addressSubscription` keys, fixing a permanent per-address leak on disconnect.
- **Connection limiter** (global + per source IP) held for the connection lifetime — the gin throttler releases its slot at the HTTP→WS upgrade and so does not bound live WS connections. Keyed on the raw remote address (not `X-Forwarded-For`).
- **`secured: true` now works** for `/subscribe`: the route bypassed the router wrapper, so the flag was silently ignored. Auth now runs before the upgrade.

## Configuration

Connection and address caps are operator-tunable under `webServer.*` (`0 = unlimited` for connections; built-in safe defaults for address caps). Defaults keep the dev/example config usable (`open: true`).

## Scope

Cross-origin (`CheckOrigin`) and direct network exposure remain operator/deployment responsibilities — the node is headless and expected to sit behind a reverse proxy that terminates TLS and enforces origin/auth. Those are intentionally **not** in-node changes here.

## Tests

Oversized frame → close 1009; global and per-IP connection caps (503 over cap, slot release on close); secured handshake (401 without auth, 101 with); address-map leak reclamation; per-subscribe and per-client caps; duplicate-address counting; and the read-limit derivation/per-client clamp. `go build`, `go vet`, and `go test -race` on the affected packages are clean.

## Hardening `/subscribe` WebSocket endpoint against resource exhaustion (CWE-770)

This PR hardens the unauthenticated remote `/subscribe` WebSocket path against resource-exhaustion attacks **without changing blockchain-critical logic** (consensus, transaction execution, state/chain storage, or KVM). The scope is limited to **network/API routing** and the **in-process WebSocket subscription hub/client lifecycle** that delivers blockchain event notifications.

### Networking & API-level hardening
- **Lifetime WebSocket connection limiter (global + per-source-IP)**: Enforces concurrent connection caps across the server and per remote IP by acquiring capacity before the HTTP→WS upgrade; rejects over-cap handshakes with **HTTP 503** and releases capacity on teardown. Uses the host portion of `RemoteAddr` (no `X-Forwarded-For`).
- **Authentication enforced before WS upgrade**: Makes `/subscribe` respect `secured:true` by running authentication handlers **before** upgrading to WebSocket (adds route misconfiguration warning behavior when `open:false` but `secured:true`).
- **Configurable operator limits plumbed via `webServer.*`**:
  - max live `/subscribe` connections (node-wide and per-IP; `0` = unlimited)
  - max addresses allowed **per `/subscribe` request**
  - max watched addresses allowed **per WebSocket client connection**
  - accessors added throughout facade/interface/mocks to expose these values.

### Bounded subscription state & concurrency/error handling
- **Per-subscribe + per-client address caps with bounded mutation**:
  - Validates address counts **before** mutating hub state.
  - **Deduplicates** addresses within a single subscribe request and counts only **new** addresses against the per-client cap.
  - Rejects over-cap insertions cleanly (and ensures insertion failures are propagated to the requesting client rather than silently ignored).
- **Inbound read/frame exhaustion protections on both read paths**:
  - Adds **read limits** during the handshake subscription processing and in the steady-state `client.loopIn`.
  - Introduces a refreshable read deadline extended via **ping/pong** so “passive but alive” subscribers remain connected while idle/dead ones are reclaimed at `pongWait`.
  - Oversized frames are terminated with **WebSocket close code 1009**.
  - Derives the effective inbound read limit from the effective per-subscribe address cap (including coherence/clamping rules) to keep behavior bounded.
- **Teardown coordination to prevent goroutine/resource leaks**:
  - Adds `client.Done()` and ensures limiter capacity is held for the full WebSocket connection lifetime and released exactly once at teardown.

### Memory/resource leak prevention in hub cleanup
- **Reclaims empty subscription buckets**:
  - Fixes a map leak by reclaiming empty `addressSubscription` keys and correctly updating per-client address tracking on disconnect/unsubscribe paths (`handleClientDelete`/`deleteAll`), preventing unbounded key growth.

### Test coverage
Adds/extends tests to cover:
- oversized frame rejection (**1009**),
- global/per-IP connection caps and **release-on-close** behavior (**503**),
- secured handshake enforcement (**401** unauthenticated, **101** authenticated),
- address-map reclamation on disconnect,
- per-subscribe/per-client cap enforcement, duplicate counting semantics (counts once), and read-limit derivation/clamping behavior.

### DIFF
--- a/network/api/api.go
+++ b/network/api/api.go
@@ -138,10 +138,29 @@ func RegisterRoutes(ctx context.Context, ws *gin.Engine, routesConfig config.API
 		registerLoggerWsRoute(ws, marshalizerForLogs, routesConfig)
 	}
 
+	// secured only attaches the auth handler; the route is registered on open. Warn on
+	// the secured-but-not-open footgun so it does not read as "auth-protected" when it
+	// is actually absent.
+	if routesConfig.IsRouteSecured("subscribe", "/subscribe") && !routesConfig.IsRouteEnabled("subscribe", "/subscribe") {
+		log.Warn("subscribe route has secured:true but open:false; /subscribe will not be registered. Set open:true to enable it (secured then requires Basic Auth).")
+	}
+
 	if routesConfig.IsRouteEnabled("subscribe", "/subscribe") {
 		var postConnUrl, postConnApiKey string
+		var subscribeOpts wsocket.SubscribeOptions
+		var hubLimits clientSocket.Limits
 		if ok {
 			postConnUrl, postConnApiKey = apiHandler.WSConnectionURL(), apiHandler.WSConnectionAPIKey()
+			subscribeOpts.MaxConnections = apiHandler.WSMaxConnections()
+			subscribeOpts.MaxConnectionsPerIP = apiHandler.WSMaxConnectionsPerIP()
+			hubLimits.MaxAddressesPerSubscribe = int(apiHandler.WSMaxAddressesPerSubscribe())
+			hubLimits.MaxAddressesPerClient = int(apiHandler.WSMaxAddressesPerClient())
+		}
+		// Honour `secured: true` for /subscribe. The route is registered directly on the
+		// engine (not via the RouterWrapper), so the auth handler must be applied here or
+		// the flag is silently ignored (GHSA-4fwh-wrm6-97xm).
+		if routesConfig.IsRouteSecured("subscribe", "/subscribe") {
+			subscribeOpts.AuthHandlers = []gin.HandlerFunc{authHandler}
 		}
 
 		var wsFacade clientSocket.WSFacade
@@ -150,8 +169,8 @@ func RegisterRoutes(ctx context.Context, ws *gin.Engine, routesConfig config.API
 		}
 
 		indexer.UseEventQueue = true
-		hub := clientSocket.NewHub(postConnUrl, postConnApiKey, wsFacade)
-		wsocket.SubscribeTopics(ws, hub)
+		hub := clientSocket.NewHub(postConnUrl, postConnApiKey, wsFacade, hubLimits)
+		wsocket.SubscribeTopics(ws, hub, subscribeOpts)
 		go hub.StartServer(ctx)
 	}
 }
--- a/network/api/api.go
+++ b/network/api/api.go
@@ -46,6 +46,9 @@ var log = logger.GetOrCreate("api")
 const (
 	logPackage = "log"
 	logRoute   = "/log"
+
+	subscribePackage = "subscribe"
+	subscribeRoute   = "/subscribe"
 )
 
 type validatorInput struct {
@@ -141,11 +144,11 @@ func RegisterRoutes(ctx context.Context, ws *gin.Engine, routesConfig config.API
 	// secured only attaches the auth handler; the route is registered on open. Warn on
 	// the secured-but-not-open footgun so it does not read as "auth-protected" when it
 	// is actually absent.
-	if routesConfig.IsRouteSecured("subscribe", "/subscribe") && !routesConfig.IsRouteEnabled("subscribe", "/subscribe") {
+	if routesConfig.IsRouteSecured(subscribePackage, subscribeRoute) && !routesConfig.IsRouteEnabled(subscribePackage, subscribeRoute) {
 		log.Warn("subscribe route has secured:true but open:false; /subscribe will not be registered. Set open:true to enable it (secured then requires Basic Auth).")
 	}
 
-	if routesConfig.IsRouteEnabled("subscribe", "/subscribe") {
+	if routesConfig.IsRouteEnabled(subscribePackage, subscribeRoute) {
 		var postConnUrl, postConnApiKey string
 		var subscribeOpts wsocket.SubscribeOptions
 		var hubLimits clientSocket.Limits
@@ -159,7 +162,7 @@ func RegisterRoutes(ctx context.Context, ws *gin.Engine, routesConfig config.API
 		// Honour `secured: true` for /subscribe. The route is registered directly on the
 		// engine (not via the RouterWrapper), so the auth handler must be applied here or
 		// the flag is silently ignored (GHSA-4fwh-wrm6-97xm).
-		if routesConfig.IsRouteSecured("subscribe", "/subscribe") {
+		if routesConfig.IsRouteSecured(subscribePackage, subscribeRoute) {
 			subscribeOpts.AuthHandlers = []gin.HandlerFunc{authHandler}
 		}
 
--- a/network/api/api.go
+++ b/network/api/api.go
@@ -153,8 +153,8 @@ func RegisterRoutes(ctx context.Context, ws *gin.Engine, routesConfig config.API
 			postConnUrl, postConnApiKey = apiHandler.WSConnectionURL(), apiHandler.WSConnectionAPIKey()
 			subscribeOpts.MaxConnections = apiHandler.WSMaxConnections()
 			subscribeOpts.MaxConnectionsPerIP = apiHandler.WSMaxConnectionsPerIP()
-			hubLimits.MaxAddressesPerSubscribe = int(apiHandler.WSMaxAddressesPerSubscribe())
-			hubLimits.MaxAddressesPerClient = int(apiHandler.WSMaxAddressesPerClient())
+			hubLimits.MaxAddressesPerSubscribe = wsocket.ClampUint32ToInt(apiHandler.WSMaxAddressesPerSubscribe())
+			hubLimits.MaxAddressesPerClient = wsocket.ClampUint32ToInt(apiHandler.WSMaxAddressesPerClient())
 		}
 		// Honour `secured: true` for /subscribe. The route is registered directly on the
 		// engine (not via the RouterWrapper), so the auth handler must be applied here or
--- a/network/api/websocket/routes.go
+++ b/network/api/websocket/routes.go
@@ -21,6 +21,8 @@ const (
 )
 
 var upgrader = gorilla.Upgrader{
+	// Origin isn't enforced here by design: the node runs headless behind an operator proxy
+	// that owns origin/CORS policy, and /subscribe carries no ambient credentials (KLC-2450).
 	CheckOrigin: func(r *http.Request) bool {
 		return true
 	},
--- a/network/api/websocket/routes.go
+++ b/network/api/websocket/routes.go
@@ -2,6 +2,7 @@ package websocket
 
 import (
 	"fmt"
+	"net"
 	"net/http"
 	"time"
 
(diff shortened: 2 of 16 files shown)


network/api/api.go
```diff
@@ -133,10 +133,29 @@ func RegisterRoutes(ctx context.Context, ws *gin.Engine, routesConfig config.API
...
 	if isSubscriptionRouteEnabled(routesConfig) {
 		var postConnUrl, postConnApiKey string
+		var subscribeOpts wsocket.SubscribeOptions
+		var hubLimits clientSocket.Limits
 		if ok {
 			postConnUrl, postConnApiKey = apiHandler.WSConnectionURL(), apiHandler.WSConnectionAPIKey()
+			subscribeOpts.MaxConnections = apiHandler.WSMaxConnections()
+			subscribeOpts.MaxConnectionsPerIP = apiHandler.WSMaxConnectionsPerIP()
+			hubLimits.MaxAddressesPerSubscribe = int(apiHandler.WSMaxAddressesPerSubscribe())
+			hubLimits.MaxAddressesPerClient = int(apiHandler.WSMaxAddressesPerClient())
```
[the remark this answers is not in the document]
fbsobreira: Good catch — fixed. Both address-cap conversions now go through the same `ClampUint32ToInt` guard used for the connection limits, so a 32-bit `int` wrap can't silently drop the operator's configured cap. (For the address caps the wrap was fail-safe — it fell back to the safe default rather than disabling a bound — so this is a consistency/operator-intent fix rather than a security one, but worth keeping the two paths uniform.)

RomuloSiebra (dismissed):

network/api/websocket/routes.go
```diff
@@ -79,9 +120,24 @@ func processSubscription(conn *gorilla.Conn, hub *websocket.SocketHub) {
 		return
 	}
 
+	if len(req.Addresses) > hub.MaxAddressesPerSubscribe() {
+		_ = conn.WriteJSON(map[string]string{"error": "too many addresses in a single subscribe"})
+		_ = conn.Close()
+		return
+	}
+
 	log.Debug(subscribeOp, "types", fmt.Sprintf("%v", parsedTypes), "addresses", fmt.Sprintf("%v", req.Addresses))
```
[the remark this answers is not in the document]
fbsobreira: Fixed in 52e6f967 — `/subscribe` now logs `addressCount` (`len(req.Addresses)`) instead of formatting the full `req.Addresses` slice, so it no longer emits user-identifying addresses or scales Debug log volume with the (capped-but-large) request payload.

websocket/websocket.go
```diff
@@ -253,14 +273,50 @@ func (h *SocketHub) RemoveClient(c *client) {
...
-	for _, types := range eventType {
-		switch types {
+	// Reject before mutating if the per-connection cap would be exceeded; only addresses
+	// new to this client count (GHSA-4fwh-wrm6-97xm, Impact C).
+	if h.clientAddresses[c]+h.countNewAddresses(addresses, c) > h.limits.maxAddressesPerClient {
+		return fmt.Errorf("address subscription limit reached for this connection (max %d)", h.limits.maxAddressesPerClient)
+	}
+
+	acceptAccounts, acceptTransactions := h.applyEventTypes(eventType, c)
+	h.addAddressSubscriptions(addresses, c, acceptAccounts, acceptTransactions)
```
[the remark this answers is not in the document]
fbsobreira: Fixed in 52e6f967 — `HandleClientInsertion` now gates the per-subscribe cap, `countNewAddresses`, and `addAddressSubscriptions` on address-scoped types being present (`ACCOUNTS`/`USER_TRANSACTIONS`) via a `containsAddressScoped` check. A blocks/transactions-only subscribe that carries addresses no longer creates no-op `addressSubscription` entries or consumes the per-connection address budget; global block/transaction subscriptions still register through `applyEventTypes`. Added `TestHandleClientInsertion_NonAddressScopedIgnoresAddresses` to cover the path.

network/api/websocket/routes.go
```diff
@@ -79,9 +120,24 @@ func processSubscription(conn *gorilla.Conn, hub *websocket.SocketHub) {
 		return
 	}
 
-	log.Debug(subscribeOp, "types", fmt.Sprintf("%v", parsedTypes), "addresses", fmt.Sprintf("%v", req.Addresses))
+	if len(req.Addresses) > hub.MaxAddressesPerSubscribe() {
```
nickgs1337, line 125: nit: this cap is already enforced by HandleClientInsertion just below (that's all handleDynamicSubscribe relies on), so the pre-check is redundant and MaxAddressesPerSubscribe() would be dead code if you drop it. Fine to keep as a belt-and-suspenders check. If you do collapse it, keep the static message rather than forwarding the error, since that one leaks the configured cap number to the client.

websocket/websocket.go
```diff
@@ -253,14 +273,69 @@ func (h *SocketHub) RemoveClient(c *client) {
...
 	h.mu.Lock()
 	defer h.mu.Unlock()
 
-	var acceptTransactions bool
-	var acceptAccounts bool
-	for _, types := range eventType {
-		switch types {
+	// Reject before mutating if the per-connection cap would be exceeded; only addresses
+	// new to this client count.
+	if wantsAddresses && h.clientAddresses[c]+h.countNewAddresses(addresses, c) > h.limits.maxAddressesPerClient {
```
nickgs1337, line 304: nit: countNewAddresses builds a fresh seen-map and scans all addresses under h.mu, then addAddressSubscriptions scans them again, so a big subscribe roughly doubles how long we hold the hub's single write lock. The per-subscribe cap keeps it small so it's not urgent, but you could build the dedup set before taking the lock, or fold the count into the add pass to drop the second scan.

websocket/client.go
```diff
@@ -73,6 +80,15 @@ func (c *client) loopIn() {
 		c.hub.handleClientDelete(c)
 	}()
 
+	// Bound every inbound frame and keep a lifetime read deadline, refreshed by the pong
+	// handler and each read. loopOut's pings keep a live client warm while a dead/idle one
+	// is reclaimed at pongWait (GHSA-4fwh-wrm6-97xm).
+	c.conn.SetReadLimit(c.hub.limits.maxMessageSize)
+	_ = c.conn.SetReadDeadline(time.Now().Add(pongWait))
```
nickgs1337, line 87: The read-deadline plus pong-handler reclamation here is the main new defense (a silent or dead client freeing its slot at pongWait), and it's the one piece with no direct test while the read limit, caps, slot release and auth all have coverage. Worth a focused test, maybe a test-only short pongWait, or asserting the connection cap frees after an idle client stops answering pings.

nickgs1337:
On websocket/client.go:149, the WriteJSON in loopOut has no write deadline, so a client that keeps sending frames (re-arming the read deadline) while never reading can park the loopOut goroutine there indefinitely. Impact's bounded since out caps at 500 and the conn counts against the new limits, but while we're hardening this it'd be cheap to SetWriteDeadline(now+pingPeriod) before that write and close on timeout.

nickgs1337:
On network/api/websocket/routes.go:24, CheckOrigin returns true for any origin. Not exploitable today since /subscribe uses HTTP Basic (browsers won't auto-attach it cross-site) and the feed is public read-only when open, and it's pre-existing anyway. Since this PR formalizes the auth story though, might be worth a one-line comment there noting why the permissive origin is safe now and that it'd need an allowlist if cookie or session auth ever lands on this route.

network/api/websocket/routes.go
fbsobreira: Kept as the belt-and-suspenders check. It is intentionally retained because it is *unconditional*, so it also bounds a `BLOCKS`/`TRANSACTIONS`-only subscribe that carries a large address list (`HandleClientInsertion` only enforces the per-subscribe cap for address-scoped types), and it gives a fresh client a clean rejection before the defensive guard. On the leak: neither path exposes the configured cap — the pre-check writes the static `"too many addresses in a single subscribe"`, and `HandleClientInsertion`'s detailed error is only `log.Warn`'d (routes.go:134), never sent to the client. No change made.

websocket/websocket.go
fbsobreira: Real (bounded) nit — I kept the two passes deliberately. Folding the count into the add pass would break the reject-before-mutate invariant the per-client cap relies on (we must know the full new-address count before touching the maps to reject the subscribe atomically), and `addAddressSubscriptions` has to iterate every deduped address anyway — including ones the client already watches — to update the per-address `acceptAccount`/`acceptTransaction` flags. So only the in-call dedup could move out from under the lock, a sub-millisecond saving at the per-subscribe cap. Leaving as-is; happy to revisit if that cap is ever raised substantially.

websocket/client.go
fbsobreira: Added `TestClient_IdleConnectionReclaimedAtPongWait` (`websocket/hardening_test.go`): it stands up a real WS server, connects a client that never reads (so it never pongs), and asserts the owner's `Done()` slot is released once the lifetime read deadline elapses. To keep it fast, `pingPeriod`/`pongWait` are now `var`s a test helper shortens — nothing in production mutates them. Fixed in 99e8a5f2.

fbsobreira:
@nickgs1337 both follow-ups addressed in 99e8a5f2:

- **`websocket/client.go:149` (loopOut write deadline):** added `SetWriteDeadline(now + pingPeriod)` before the `WriteJSON` and close on error, so a client that keeps re-arming its read deadline but never drains its socket can no longer park the `loopOut` goroutine indefinitely.
- **`network/api/websocket/routes.go:24` (permissive CheckOrigin):** added a comment noting origin is not enforced in the node by design — it runs headless behind an operator proxy that owns origin/CORS policy — and that `/subscribe` carries no ambient credentials. An origin allowlist would only be needed if cookie/session auth ever lands on this route (tracked in KLC-2450).

RomuloSiebra (changes requested):
In /subscribe, address values are count-limited, but they are not constrained by byte size or validated before being retained as subscription map keys.

This leaves a memory-amplification path where retained subscription data can grow disproportionately per long-lived connection.

With the current 1 MiB frame floor, a client can repeatedly send subscribe payloads containing unique very large address strings that still fit the frame limit; each unique value can be retained as a key, so memory keeps growing even though the number-of-addresses guard exists.

network/api/websocket/routes.go
```diff
@@ -79,9 +122,24 @@ func processSubscription(conn *gorilla.Conn, hub *websocket.SocketHub) {
...
```
RomuloSiebra, line 125: In /subscribe, address values are count-limited here, but they are not constrained by byte size or validated before being retained as subscription map keys downstream.

This leaves a memory-amplification path where retained subscription data can grow disproportionately per long-lived connection.

With the current 1 MiB frame floor, a client can repeatedly send subscribe payloads containing unique very large address strings that still fit the frame limit; each unique value can be retained as a key, so memory keeps growing even though the number-of-addresses guard exists.

RomuloSiebra (changes requested):

nickgs1337 (dismissed):

network/api/websocket/routes.go
fbsobreira: Good catch — fixed in 918f1bf5. `HandleClientInsertion` now rejects any subscription address longer than a klv bech32 address (`maxEncodedAddressLength = 62`, matching the node's own encoded-address cap) before it is retained as a map key. So the per-connection count cap is now backed by a hard per-key size bound: retained memory is bounded at `maxAddressesPerClient × 62` bytes (~3 MB/conn worst case) instead of `count × frameSize`. It sits at the single chokepoint both entry paths share, so it also covers the repeated-subscribe amplification you described — `handleDynamicSubscribe` routes through the same guard. Added `TestHandleClientInsertion_RejectsOversizedAddress` and `_AcceptsMaxLengthAddress` (boundary is inclusive).

RomuloSiebra (approved):

nickgs1337 (approved):

PR #79, opened 2026-06-25
[KLC-2506] feat: add /network/economics and /network/account-totals supply endpoints

fbsobreira:
## Summary

Adds two read-only, opt-in node endpoints exposing live KLV held-balance aggregates for supply reconciliation, plus a KFI/KLV circulating-supply transposition fix in the indexer. Every held aggregate is KLV already inside `circulatingSupply`, so summing them against `circulatingSupply` recovers the full physical KLV total.

## `GET /network/economics`

Disabled by default (`open: false`); memoized per block. Returns supply figures (`initialSupply`, `maxSupply`, `mintedValue`, `burnedValue`, `circulatingSupply`, `totalStaked`) plus held aggregates from KApp/peer state:

- `pendingRewardsTotal` — unclaimed PREW (Validators KApp)
- `marketEscrowTotal` — open-order escrow (`CurrentBid` in KLV + `RoyaltiesFixedDeposit`)
- `feesPoolKlvTotal` — Σ `KLVBalance` across KDA fee pools
- `fprPoolTotal` — KLV staking-reward pool (`CurrentFPRAmount` + unclaimed `FPRData`; non-KLV `KDAS` excluded)
- `accumulatedFeesTotal` — Σ validator `AccumulatedFees` (accrued, pre-distribution)
- `systemAccountKlvBalance` — KLV held by the system-account KApp

## `GET /network/account-totals`

Disabled by default; memoized per block. A full walk of the user-accounts trie returning `accountCount`, `balanceTotal`, and `allowanceTotal` (inline KLV `Balance` + `Allowance`). Accounts are identified by `Address == leaf key`, so smart-contract code leaves — which can decode cleanly into a `UserAccountData` — are excluded. Frozen/unfrozen are out of scope (sub-trie; `totalStaked` covers frozen).

Both endpoints share a generic per-block `blockCache[T]` helper.

## Indexer fix

Corrects `SaveEpochInfo` where `KFICirculatingSupply`/`KLVCirculatingSupply` were assigned from the swapped KDA structs, `KFITotalStaked` was never set, and a copy-paste guard checked the wrong staking-bytes variable.

**Downstream impact**: epoch documents indexed before this fix carry swapped `kfiCirculationSupply`/`klvCirculationSupply` values and `kfiTotalStaked: 0`; documents are correct from the first epoch indexed with this build. Consumers reading historical epoch docs should account for this (or reindex). No ES mapping change — the epoch index uses dynamic mapping and all four fields already exist in every document.

## Tests

Unit tests for the KApp scan aggregates (market escrow, pending rewards), the node-level FPR / fees / balance / allowance summing — including FPR `KDAS` exclusion and the account-vs-code-leaf discrimination — and the new route handlers.

Added two opt-in, read-only network API endpoints for live supply/account aggregation: `GET /network/economics` and `GET /network/account-totals` (both disabled by default in `config/node/api.yaml`). Both handlers are memoized per block and backed by new facade/node methods (`GetEconomics`, `GetAccountTotals`) that derive totals from on-chain state (KApps/trie data and user-accounts trie) rather than affecting transaction processing or consensus.

**Blockchain-critical impact (stability/data integrity)**
- **Networking/API**: Introduces new Gin route handlers and response models (`EconomicsResponse`, `AccountTotalsResponse`) with explicit internal error propagation (`ErrGetEconomics`, `ErrGetAccountTotals`). Adds tests for nil app/facade context and facade handler failures.
- **State management (node, read-only)**:
  - Adds a mutex-protected generic `blockCache[T]` keyed by the current block header nonce to ensure per-block memoization and safe concurrent access.
  - Hardens caching semantics by rejecting `(nil, nil)` compute results (`errNilComputeResult`), and adds tests covering per-block invalidation and concurrent callers sharing a single computation.
  - Economics aggregation scans relevant KApp data tries and computes held aggregates (fees-pool KLV total, FPR pool KLV total with KLV-only logic and positive-delta handling, system-account KLV balance, plus pending rewards and market escrow totals from KApp interfaces) alongside base supply figures from the asset interface.
  - Account totals traverse the user-accounts trie, decode `UserAccountData`, and **exclude smart-contract code leaves** by requiring the embedded/stored address to match the leaf key.

**Cross-cutting correctness fixes**
- **Indexer/data integrity**: Fixes `SaveEpochInfo` to correct swapped KDA circulating-supply assignments (KFI vs KLV) and fixes a copy-paste staking-bytes guard so the correct byte slice is validated.

**KApp/trie interface expansion**
- Extends KApp interfaces and implementations to expose trie-derived aggregates used by economics:
  - `ValidatorsKapp.GetPendingRewardsTotal`
  - `MarketKapp.GetMarketEscrowTotal`
- Adds unit tests for prefix filtering, claimed vs unclaimed handling, and nil/edge paths in trie scanning.

**Concurrency/error handling**
- Concurrency changes are confined to the new per-block cache (lock-protected compute + nonce-based invalidation), with additional error-path tests for aggregation helpers and route handlers.

### DIFF
--- a/config/node/api.yaml
+++ b/config/node/api.yaml
@@ -67,8 +67,10 @@ apiPackages:
         open: true
       - name: /economics
         open: false
+        secured: true
       - name: /account-totals
         open: false
+        secured: true
   node:
     routes:
       - name: /status
--- a/config/node/api.yaml
+++ b/config/node/api.yaml
@@ -65,6 +65,8 @@ apiPackages:
         open: true
       - name: /network-parameters
         open: true
+      - name: /economics
+        open: false
   node:
     routes:
       - name: /status
--- a/config/node/api.yaml
+++ b/config/node/api.yaml
@@ -67,6 +67,8 @@ apiPackages:
         open: true
       - name: /economics
         open: false
+      - name: /account-totals
+        open: false
   node:
     routes:
       - name: /status
@@ -84,7 +86,7 @@ apiPackages:
       - name: /statistics
         open: true
       - name: /set-redundancy
-        open: true
+        open: false
         secured: true
       - name: /enable-epochs
         open: true
--- a/core/kapp/market/market.go
+++ b/core/kapp/market/market.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/hex"
 	"errors"
+	"math"
 	"strconv"
 	"time"
 	"unicode/utf8"
@@ -227,16 +228,24 @@ func (m *marketKapp) GetMarketEscrowTotal() (int64, error) {
 		}
 		order := &kapps.MarketOrderData{}
 		if err := m.marshalizer.Unmarshal(order, raw); err != nil {
-			log.Warn("GetMarketEscrowTotal: skipping undecodable market order", "error", err)
-			continue
+			return 0, err
 		}
 		if order.IsClaimed {
 			continue
 		}
-		total += order.RoyaltiesFixedDeposit
+		add := order.RoyaltiesFixedDeposit
 		if bytes.Equal(order.CurrencyID, kdautils.KLVIdentifier) {
-			total += order.CurrentBid
+			add += order.CurrentBid
 		}
+		if add < 0 {
+			log.Warn("GetMarketEscrowTotal: negative escrow amount, skipping", "key", hex.EncodeToString(key))
+			continue
+		}
+		if total > math.MaxInt64-add {
+			log.Warn("GetMarketEscrowTotal: sum would overflow int64, skipping entry")
+			continue
+		}
+		total += add
 	}
 
 	return total, nil
--- a/core/kapp/market/market.go
+++ b/core/kapp/market/market.go
@@ -192,6 +192,7 @@ func (m *marketKapp) GetMarketOrder(orderID []byte) (state.KAppAccountHandler, *
 
 // GetMarketEscrowTotal returns the total KLV held in open market orders: each unclaimed order's
 // RoyaltiesFixedDeposit (always KLV) plus its CurrentBid when the order is priced in KLV.
+// Mid-walk trie errors are swallowed upstream, so a truncated walk undercounts silently (KLC-2509).
 func (m *marketKapp) GetMarketEscrowTotal() (int64, error) {
 	app, err := m.accountsCacher.LoadKAppUncached(kapps.MarketKAppAddress)
 	if err != nil {
--- a/core/kapp/market/market.go
+++ b/core/kapp/market/market.go
@@ -2,12 +2,14 @@ package market
 
 import (
 	"bytes"
+	"context"
 	"encoding/hex"
 	"errors"
 	"strconv"
 	"time"
 	"unicode/utf8"
 
+	logger "github.com/klever-io/klever-go-logger"
 	"github.com/klever-io/klever-go/common"
 	"github.com/klever-io/klever-go/core"
 	"github.com/klever-io/klever-go/core/kapp"
@@ -24,6 +26,8 @@ import (
 )
 
 var _ kapp.MarketKapp = (*marketKapp)(nil)
+
+var log = logger.GetOrCreate("kapp/market")
 
 type marketKapp struct {
 	hasher         hashing.Hasher
@@ -183,6 +187,59 @@ func (m *marketKapp) GetMarketOrder(orderID []byte) (state.KAppAccountHandler, *
 	}
 
 	return marketKapp, market, nil
+}
+
+// GetMarketEscrowTotal returns the total KLV held in open market orders: each unclaimed order's
+// RoyaltiesFixedDeposit (always KLV) plus its CurrentBid when the order is priced in KLV.
+func (m *marketKapp) GetMarketEscrowTotal() (int64, error) {
+	app, err := m.accountsCacher.LoadKAppUncached(kapps.MarketKAppAddress)
+	if err != nil {
+		return 0, err
+	}
+
+	dataTrie := app.DataTrie()
+	if check.IfNil(dataTrie) {
+		return 0, nil
+	}
+
+	leavesChannel, err := dataTrie.GetAllLeavesOnChannel(app.GetRootHash(), context.Background())
+	if err != nil {
+		return 0, err
+	}
+
+	// Collect order keys first so the scan goroutine finishes before we read values back.
+	prefix := []byte(kapps.MarketOrderPrefix + kapps.Sp)
+	orderKeys := make([][]byte, 0)
+	for leaf := range leavesChannel {
+		if !bytes.HasPrefix(leaf.Key(), prefix) {
+			continue
+		}
+		key := make([]byte, len(leaf.Key()))
+		copy(key, leaf.Key())
+		orderKeys = append(orderKeys, key)
+	}
+
+	total := int64(0)
+	for _, key := range orderKeys {
+		raw := app.GetStorage(key) // trims the trie tail, unlike the raw leaf value
+		if len(raw) == 0 {
+			continue
+		}
+		order := &kapps.MarketOrderData{}
+		if err := m.marshalizer.Unmarshal(order, raw); err != nil {
+			log.Warn("GetMarketEscrowTotal: skipping undecodable market order", "error", err)
+			continue
+		}
+		if order.IsClaimed {
+			continue
+		}
+		total += order.RoyaltiesFixedDeposit
+		if bytes.Equal(order.CurrencyID, kdautils.KLVIdentifier) {
+			total += order.CurrentBid
+		}
+	}
+
+	return total, nil
 }
 
 func (m *marketKapp) SetMarketOrder(marketKapp state.KAppAccountHandler, order *kapps.MarketOrderData) error {
--- a/core/kapp/market/market.go
+++ b/core/kapp/market/market.go
@@ -223,24 +223,9 @@ func (m *marketKapp) GetMarketEscrowTotal() (int64, error) {
 
 	total := int64(0)
 	for _, key := range orderKeys {
-		raw := app.GetStorage(key) // trims the trie tail, unlike the raw leaf value
-		if len(raw) == 0 {
-			continue
-		}
-		order := &kapps.MarketOrderData{}
-		if err := m.marshalizer.Unmarshal(order, raw); err != nil {
+		add, err := m.orderEscrowAmount(app, key)
+		if err != nil {
 			return 0, err
-		}
-		if order.IsClaimed {
-			continue
-		}
-		add := order.RoyaltiesFixedDeposit
-		if bytes.Equal(order.CurrencyID, kdautils.KLVIdentifier) {
-			add += order.CurrentBid
-		}
-		if add < 0 {
-			log.Warn("GetMarketEscrowTotal: negative escrow amount, skipping", "key", hex.EncodeToString(key))
-			continue
 		}
 		if total > math.MaxInt64-add {
 			log.Warn("GetMarketEscrowTotal: sum would overflow int64, skipping entry")
@@ -250,6 +235,31 @@ func (m *marketKapp) GetMarketEscrowTotal() (int64, error) {
 	}
 
 	return total, nil
+}
+
+// orderEscrowAmount returns the KLV escrowed by one stored order: RoyaltiesFixedDeposit plus
+// CurrentBid when priced in KLV; zero for empty, claimed, or negative-value entries.
+func (m *marketKapp) orderEscrowAmount(app state.KAppAccountHandler, key []byte) (int64, error) {
+	raw := app.GetStorage(key) // trims the trie tail, unlike the raw leaf value
+	if len(raw) == 0 {
+		return 0, nil
+	}
+	order := &kapps.MarketOrderData{}
+	if err := m.marshalizer.Unmarshal(order, raw); err != nil {
+		return 0, err
+	}
+	if order.IsClaimed {
+		return 0, nil
+	}
+	add := order.RoyaltiesFixedDeposit
+	if bytes.Equal(order.CurrencyID, kdautils.KLVIdentifier) {
+		add += order.CurrentBid
+	}
+	if add < 0 {
+		log.Warn("GetMarketEscrowTotal: negative escrow amount, skipping", "key", hex.EncodeToString(key))
+		return 0, nil
+	}
+	return add, nil
 }
 
 func (m *marketKapp) SetMarketOrder(marketKapp state.KAppAccountHandler, order *kapps.MarketOrderData) error {
--- a/indexer/elasticProcessor.go
+++ b/indexer/elasticProcessor.go
@@ -1807,7 +1807,7 @@ func (ei *elasticProcessor) SaveEpochInfo(epoch uint32, validators []kapp.Valida
 	if err != nil {
 		return err
 	}
-	if len(klvStakingBytes) == 0 {
+	if len(kfiStakingBytes) == 0 {
 		return common.ErrEmptyString
 	}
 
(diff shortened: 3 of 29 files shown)


node/nodeEconomics.go
```diff
@@ -0,0 +1,219 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// A missing KLV StakingData record (fresh/test networks) means no stake, not an error.
+	staking, err := n.loadStakingData(string(kdautils.KLVIdentifier))
+	if err != nil && !errors.Is(err, common.ErrStakingNotFound) {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
fbsobreira: Verified this cannot panic: `computeEconomics` calls `GetAsset` first, which returns `common.ErrNilKAppAccountsAdapter` on a nil `n.kapps` (see `nodeAsset.go`), so the function returns a clean error before `loadStakingData` is reached. No change needed.

network/api/models/economics.go
```diff
@@ -0,0 +1,17 @@
+package models
+
+// EconomicsResponse holds live KLV supply figures plus node-state held aggregates
+// (pending rewards, market escrow, fees-pool KLV, system-account KLV). See KLC-2506.
```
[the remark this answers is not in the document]
fbsobreira: Good catch — added `fprPoolTotal` to the `EconomicsResponse` doc comment in 2373fdc7.

node/nodeEconomics.go
```diff
@@ -0,0 +1,199 @@
...
+func (n *Node) loadAccumulatedFeesTotal() int64 {
+	total := int64(0)
+	for _, peer := range n.validatorsProvider.GetLatestPeers() {
+		if check.IfNil(peer) {
+			continue
+		}
+		total += peer.GetAccumulatedFees()
+	}
+	return total
+}
```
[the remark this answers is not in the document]
fbsobreira: Agreed — for a supply-audit endpoint a silent zero is wrong. `loadAccumulatedFeesTotal` now reads peers via the error-returning `validatorStatistics.ListPeerAccounts` (returning `0` only on an empty finalized root, i.e. before the first finalized block) and propagates any read error through `computeEconomics`. Sub-tests cover sum / empty-root→0 / propagate-error. Fixed in 3ba7d78d.

node/blockCache.go
```diff
@@ -0,0 +1,44 @@
...
+	computed, err := compute()
+	if err != nil {
+		return nil, err
+	}
+
+	c.cached = computed
+	c.nonce = currentNonce
+	c.valid = true
+
+	return computed, nil
```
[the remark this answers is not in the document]
fbsobreira: Good catch — added a guard that treats a nil-but-no-error compute result as an error (`errNilComputeResult`), so a future `compute` returning `(nil, nil)` can no longer silently defeat the cache. Covered by a new `blockCache` test. Fixed in 3ba7d78d.

node/nodeAccountTotals.go
```diff
@@ -0,0 +1,53 @@
...
+		if errUnmarshal := n.internalMarshalizer.Unmarshal(acc, leaf.Value()); errUnmarshal != nil {
+			continue
+		}
+		if !bytes.Equal(acc.GetAddress(), leaf.Key()) {
+			continue // code leaf, not an account
+		}
+		totals.AccountCount++
+		totals.BalanceTotal += acc.GetBalance()
+		totals.AllowanceTotal += acc.GetAllowance()
+	}
```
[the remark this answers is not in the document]
fbsobreira: `Node.ctx` is set to `context.Background()` (node.go:200), not a cancellable/shutdown context, so threading it here would gain no real cancellability — it would imply an interruptibility the node does not currently have. Genuine cancellation would mean wiring a shutdown context through all the trie scans (the KApp scans use `context.Background()` too), which is out of scope for this PR.

network/api/network/routes_test.go
```diff
@@ -182,9 +183,122 @@ func getRoutesConfig() config.APIRoutesConfig {
...
+
+	ws := startNodeServer(&facade)
+	req, _ := http.NewRequest("GET", "/network/account-totals", nil)
+	resp := httptest.NewRecorder()
+	ws.ServeHTTP(resp, req)
+
+	assert.Equal(t, http.StatusOK, resp.Code)
+
+	respBytes, _ := io.ReadAll(resp.Body)
+	respStr := string(respBytes)
```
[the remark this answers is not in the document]
fbsobreira: The whole `routes_test.go` uses `_` for these calls (`http.NewRequest` with a static GET + nil body, and `io.ReadAll` on an httptest recorder), neither of which can error in practice. Keeping the new tests consistent with the surrounding convention; happy to do a file-wide `require.NoError` pass separately if preferred.

node/blockCache_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	t.Run("propagates compute error", func(t *testing.T) {
+		t.Parallel()
+		c := &blockCache[int]{}
+		expectedErr := errors.New("boom")
+		v, err := c.get(blkc, func() (*int, error) { return nil, expectedErr })
+		require.ErrorIs(t, err, expectedErr)
+		require.Nil(t, v)
+	})
+}
```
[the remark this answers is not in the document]
fbsobreira: Added both — a nonce-transition subtest (same nonce → cache hit, advanced nonce → recompute) covering per-block invalidation, and a concurrent fan-out subtest (start channel + `sync.WaitGroup`, no sleeps) asserting a single `compute` under the lock with all callers sharing the cached pointer. Validated under `go test -race`. Fixed in d9b8666c.

node/nodeEconomics.go
```diff
@@ -0,0 +1,210 @@
...
+			continue
+		}
+		total += peer.GetAccumulatedFees()
+	}
+	return total, nil
+}
+
+// scanKAppDataTrie invokes accumulate with each stored value (GetStorage trims the trie tail). Keys are
+// collected before reading values so the scan goroutine finishes first.
+func (n *Node) scanKAppDataTrie(address []byte, accumulate func(value []byte) error) error {
```
nickgs1337: nit: this load-app then GetAllLeavesOnChannel then collect-keys then read-value then accumulate loop shows up almost verbatim in market.GetMarketEscrowTotal too, and roughly in validators.GetPendingRewardsTotal. Might be worth a shared `scanKAppTrie(app, prefix, fn)` helper for at least the node and market ones. Validators decodes the leaf inline so it's a messier fit, your call whether the cross-package extract is worth it.

node/blockCache.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+var errNilComputeResult = errors.New("blockCache: compute returned nil without error")
+
+// blockCache memoizes a value per block, keyed on the chain header nonce: callers within a block share
+// one result; the first call after a new block recomputes. The returned pointer is shared — read-only.
+type blockCache[T any] struct {
+	mu     sync.Mutex
+	nonce  uint64
+	valid  bool
```
nickgs1337: The `valid` flag is redundant. `cached` starts nil and only ever gets set alongside valid=true, so the check is just `cached != nil && nonce == currentNonce`. Same story with errNilComputeResult: both compute funcs always return a non-nil struct on the err==nil path, so that branch can't fire. Can drop the field, its assignments, and the nil-result var plus branch.

core/kapp/market/market.go
```diff
@@ -185,6 +190,67 @@ func (m *marketKapp) GetMarketOrder(orderID []byte) (state.KAppAccountHandler, *
...
+		}
+		add := order.RoyaltiesFixedDeposit
+		if bytes.Equal(order.CurrencyID, kdautils.KLVIdentifier) {
+			add += order.CurrentBid
+		}
+		if add < 0 {
+			log.Warn("GetMarketEscrowTotal: negative escrow amount, skipping", "key", hex.EncodeToString(key))
+			continue
+		}
+		if total > math.MaxInt64-add {
```
nickgs1337, line 230: nit: heads up, this pulls against the other thread asking to add these same overflow guards to the unguarded node-side aggregators, so let's pick one direction across all of them instead of both. My lean is keep the guard (cheap defense against a corrupt stored value poisoning an authoritative supply figure) but drop the `// unreachable; bounded by supply` comment, since a guard plus a comment saying it never runs is the actual smell here. The negative-value guard should stay regardless.

indexer/elasticProcessor.go
```diff
@@ -1820,10 +1820,10 @@ func (ei *elasticProcessor) SaveEpochInfo(epoch uint32, validators []kapp.Valida
 	epochInfo := &data.EpochInfo{
 		Epoch:                epoch,
 		Validators:           validatorsInfo,
-		KFICirculatingSupply: klvKDA.CirculatingSupply,
-		KLVTotalStaked:       klvStaking.TotalStaked,
-		KLVCirculatingSupply: kfiKDA.CirculatingSupply,
+		KFICirculatingSupply: kfiKDA.CirculatingSupply,
```
nickgs1337: This is a real fix (the KFI/KLV circulating-supply and total-staked fields were swapped, KFITotalStaked was never set, and the emptiness check tested klvStakingBytes twice instead of kfi), but it's out of scope for the economics/account-totals endpoints. It changes what every epoch record written to ES contains, so downstream consumers reading those fields were getting swapped values and will now get corrected ones. That behavior change really wants its own commit/PR with a downstream-impact note rather than riding along silently. Also the new `len(kfiStakingBytes)==0 -> ErrEmptyString` now hard-fails epoch indexing on any network with no KFI staking record, where before it quietly proceeded with KFITotalStaked=0. Can we confirm that won't break fresh/test-net indexing?

node/nodeEconomics.go
```diff
@@ -0,0 +1,210 @@
...
+	if check.IfNil(dataTrie) {
+		return nil
+	}
+	leavesChannel, err := dataTrie.GetAllLeavesOnChannel(app.GetRootHash(), context.Background())
+	if err != nil {
+		return err
+	}
+
+	keys := make([][]byte, 0)
+	for leaf := range leavesChannel {
```
nickgs1337, line 180: GetAllLeavesOnChannel swallows mid-walk errors: the walker just log.Errors and closes the channel (patriciaMerkleTrie.go ~565), so an early close reads as "done" here. A missing/corrupt node or a pruning race on the walked root then makes this and the sibling aggregators (GetPendingRewardsTotal, GetMarketEscrowTotal, loadFPRPoolKLVTotal) return a truncated sum with err==nil, so the endpoint answers 200 with a silently wrong supply figure. account-totals has the KLC-2509 note about this, economics doesn't, can we at least mirror it here? And is KLC-2509 meant to cover the economics totals too, or should abnormal termination surface a real error / partial flag once the channel API can report it?

node/blockCache.go
```diff
@@ -0,0 +1,50 @@
...
+func (c *blockCache[T]) get(blkc data.ChainHandler, compute func() (*T, error)) (*T, error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	currentNonce := uint64(0)
+	if header := blkc.GetCurrentBlockHeader(); !check.IfNil(header) {
+		currentNonce = header.GetNonce()
+	}
+
+	if c.valid && c.nonce == currentNonce && c.cached != nil {
```
nickgs1337: Keying only on nonce has two edges. A reorg (nonce N reverted, a different block applied at N) serves the stale economics/account-totals until the nonce advances. And the "no header yet" state aliases genesis, since both map to nonce 0, so a pre-genesis value can get served for genesis. Both are low and self-heal within a block, but keying on rootHash (or nonce+rootHash) fixes both cleanly and naturally invalidates on any state change, plus a distinct sentinel for "no header" avoids the nonce-0 alias. The new tests cover memoize/advance/concurrency but not these two, so a reorg/nonce-0 case would be worth adding.

node/nodeEconomics.go
```diff
@@ -0,0 +1,210 @@
...
+		key := make([]byte, len(leaf.Key()))
+		copy(key, leaf.Key())
+		keys = append(keys, key)
+	}
+
+	for _, key := range keys {
```
nickgs1337, line 189: nit: the KApp aggregators (GetMarketEscrowTotal, GetPendingRewardsTotal) guard against negative/overflow before summing, but the node-side ones here just do total += x. Worth picking one policy across all of them (this overlaps with the other thread on whether those guards should exist at all). Separately, scanKAppDataTrie unmarshals every leaf while GetPendingRewardsTotal prefix-filters. The staking/fees-pool tries look homogeneous today so it's not a live bug, but if any other key type ever lands under those KApps, protobuf would happily fold garbage into the total instead of erroring. A mirroring prefix filter would be safer.

config/node/api.yaml
```diff
@@ -65,6 +65,10 @@ apiPackages:
         open: true
       - name: /network-parameters
         open: true
+      - name: /economics
```
nickgs1337, line 68: Both new routes default to open:false so they're unregistered (404), which is good, but neither sets secured:true, so once an operator flips open:true they're served without basic auth. account-totals walks the whole user-accounts trie and the blockCache only collapses within-block bursts, so an unauthenticated client can force one full trie walk per block. Given the walk cost, and that this same PR hardens /set-redundancy to open:false, defaulting these two to secured:true seems like the consistent move (or at least document the operator cost).

node/blockCache.go
fbsobreira: Fixed in b224a03e — dropped `valid` and `errNilComputeResult` (plus its branch). With the cache now keyed on nonce + state root (per the other thread), the check is just `cached != nil && key match`.

node/nodeEconomics.go
fbsobreira: Considered it, but passing on the cross-package extract: the genuinely shared surface is just the walk loop — market needs the claimed/currency logic and its own prefix, and validators decodes raw leaf values inline, so a common helper would either take callbacks for everything or live in a package that node and core/kapp/* import just for this. The node-side duplication is already factored into `scanKAppDataTrie` (which now takes a prefix param per the other thread). Happy to revisit if a third in-package copy shows up.

core/kapp/market/market.go
fbsobreira: Agreed — went with guards everywhere as the single policy: kept the market/validators guards, dropped the `// unreachable; bounded by supply` comment, and the node-side aggregators now apply the same negative/overflow-skip policy via a shared `addGuarded` helper. All in b224a03e.

indexer/elasticProcessor.go
fbsobreira: Verified both directions before deciding to keep it here:

- **ES side**: the epoch template defines no field mappings (dynamic mapping only) and `EpochInfo`'s fields aren't `omitempty`, so every historical doc already carries all four fields — the fix only corrects values written from now on. Historical epoch docs keep the swapped values until a reindex; added a downstream-impact note to the PR description covering exactly this.
- **Fresh/test-net concern**: confirmed it can't fire — genesis seeds both KLV and KFI staking records (`initKLVAndKFIintoKapps`), and `UpdateNativeStakingKapps` hard-fails every block without them, so any network healthy enough to reach epoch indexing necessarily has the record. The corrected check also matches the existing convention in `getNativeStakingKApps` and genesis.

On scope: since this PR is the supply-reporting change, keeping the epoch-info correction here keeps supply figures consistent across the API and ES; it's now called out explicitly in the PR description rather than riding along silently.

node/nodeEconomics.go
fbsobreira: Added the note in b224a03e — `scanKAppDataTrie`, `GetPendingRewardsTotal`, and `GetMarketEscrowTotal` now carry the same truncated-scan caveat. And yes, KLC-2509 covers these: its impact section explicitly names `nodeEconomics.go` and `GetPendingRewardsTotal`, and its suggested fix (surfacing the walker error to callers) will flow to all of these aggregators — once the channel API can report it, they should return a real error rather than a partial flag.

node/blockCache.go
fbsobreira: Fixed in b224a03e — the key is now `nonce:stateRootHash` with a distinct `no-header` sentinel, so a reorg at the same nonce and the pre-genesis/genesis alias both invalidate correctly. Added both missing tests (reorg at same nonce, no-header vs genesis at nonce 0).

node/nodeEconomics.go
fbsobreira: Done in b224a03e on both counts, with one asymmetry: the staking scan now prefix-filters on the `KDA` prefix (`StakingData` is keyed via `ToKDAKey`), but fees-pool records are keyed by the raw asset ID with no shared prefix (`setPool` stores under the bare `kda` bytes), so there is no prefix to filter on there — that scan still takes every leaf, with a comment explaining why. On guards: node-side sums now share the same negative/overflow-skip policy via `addGuarded` (see the market thread for the policy decision). Tests cover the negative-balance skip and a non-prefixed garbage leaf being filtered out.

config/node/api.yaml
fbsobreira: Fixed in b224a03e — both routes now default `secured: true` alongside `open: false`, matching /set-redundancy: flipping them open serves them behind basic auth by default, and the trie-walk cost stays operator-gated.

nickgs1337 (approved):

Beroni (approved):

PR #83, opened 2026-07-10
[KLC-2516] [KLR-13] Nondeterministic contract deployment outcome timeout boundary

RomuloSiebra:
## Summary
Makes smart contract execution resolve deterministically when completion and the execution deadline are observable at the same boundary, and adds unit coverage for that precedence in hostCore.

## Problem
At the execution timeout boundary, the completion signal and the expired deadline can both be ready when the waiting goroutine evaluates them. Go's `select` then picks at random, so the same execution could resolve either way between runs.

## Solution
The wait now resolves that case explicitly, in favour of completion. A closed `done` channel shows execution finished within its budget, whereas an expired context also reflects how long the waiting goroutine waited to be scheduled. Preferring the completion signal keeps a node whose scheduler stalled from reporting failure for work that did finish in budget — the same direction as the existing validator tolerance in `getEffectiveTimeout` and the import-db replay timeout.

## Key Changes
**Updated:**
- `kvm/vmhost/hostCore/host.go` — deterministic completion precedence at the boundary
- `kvm/vmhost/hostCore/timeout_tiebreaker_test.go` — unit coverage for the precedence, including a timeout raised while execution is still in flight
- `kvm/mock/context/runtimeContextMock.go` — `FailExecutionCalled` hook, so the test synchronises on the timeout path instead of sleeping

## Scope
This addresses the nondeterministic branch selection at the boundary. Broader work on the execution timeout mechanism is tracked separately.

## Testing
```bash
go test ./kvm/vmhost/hostCore/... -count=1
go test ./kvm/vmhost/hostCore/ -run WaitExecutionWithDeterministicCompletion -count=20 -race
go test ./core/process/... -count=1
```

## Notes
The transaction-level regression test added earlier in this branch was removed. It calibrated a busy loop against wall-clock time, did not reach the branch being changed, and was unstable under parallel test load; the hostCore unit tests cover the precedence deterministically and without wall-clock calibration.

## Configuration Changes
None.

## Breaking Changes
None.

## Related Issues
Related to KLR-13.

## Checklist

## Summary

- Updated KVM smart-contract creation and call execution to prioritize completion when completion and timeout signals are both ready.
- Added tests for normal completion, in-flight timeout handling, and deterministic completion precedence.
- Added `FailExecutionCalled` synchronization support to `RuntimeContextMock`.
- This removes local Go `select` nondeterminism at the timeout boundary and improves deterministic transaction processing.
- Timeout and execution errors remain propagated through the existing paths.
- The change does not alter state management, networking, configuration, or public APIs.
- The change does not address broader consensus variability from wall-clock-based execution deadlines.
- No node stability or data integrity issues are introduced by the changed logic.


core/process/transaction/txProcessSmartContract_test.go
```diff
@@ -1,28 +1,89 @@
...
+ local.get $i
+ i64.const 1
+ i64.add
+ local.tee $i
+ local.get $limit
+ i64.lt_u
+ br_if $busy
+ end)
+ (func $upgrade (export "upgrade")))
+`
```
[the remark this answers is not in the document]
RomuloSiebra: Fixed in 3520a066. The malformed WAT and runtime compilation path were replaced with a validated precompiled WASM hex fixture containing the intended labeled busy loop.

core/process/transaction/txProcessSmartContract_test.go
```diff
@@ -243,3 +304,334 @@ func TestTXProcessor_smartContract(t *testing.T) {
...
+		require.Zero(t, outcome.codeLen)
+	}
+}
+
+func compileSCDeployTimeoutBusyInitWASM(t *testing.T) []byte {
+	t.Helper()
+	wat2wasm, err := exec.LookPath("wat2wasm")
+	if err != nil {
+		t.Skip("wat2wasm is required for the SC deploy timeout regression test")
+	}
```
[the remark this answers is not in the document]
RomuloSiebra: Fixed in 3520a066. The test now decodes an embedded precompiled WASM fixture, so it no longer depends on `wat2wasm` or skips when that external tool is unavailable.

core/process/transaction/txProcessSmartContract_test.go
```diff
@@ -243,3 +304,334 @@ func TestTXProcessor_smartContract(t *testing.T) {
 		})
 	}
 }
+
+func TestTXProcessor_SCDeployTimeoutBoundaryDeterministic(t *testing.T) {
+	wasmCode := compileSCDeployTimeoutBusyInitWASM(t)
```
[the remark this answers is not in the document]
RomuloSiebra: Thanks for flagging the runtime cost. I kept this regression active in short mode because the PR workflow runs `go test -short ./...`; gating it on `testing.Short()` would remove the transaction-entrypoint consensus regression from PR validation entirely. The calibrated deployment test directly covers the behavior described by KLC-2516/KLR-13.

core/process/transaction/txProcessSmartContract_test.go
```diff
@@ -243,3 +304,334 @@ func TestTXProcessor_smartContract(t *testing.T) {
...
+			},
+			ArgsParser:       smartContract.NewArgumentParser(),
+			Hasher:           args.Hasher,
+			Marshalizer:      args.Marshalizer,
+			BlockChainHook:   scDeployTimeoutBlockchainHook(world),
+			BuiltInFunctions: world.BuiltinFuncs.Container, PubkeyConv: args.PubkeyConv,
+			TxFeeHandler:        args.TxFeeHandler,
+			EconomicsFee:        freeFeeHandlerMock(),
+			GasSchedule:         notifierMock.NewGasScheduleNotifierMock(gasSchedule),
+			TxLogsProcessor:     &contextmock.TxLogsProcessorStub{},
```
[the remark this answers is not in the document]
RomuloSiebra: Fixed in 3520a066. The configured `feeHandler` is now passed to both the transaction processor and smart-contract processor, keeping gas handling consistent throughout the test setup.

core/process/transaction/txProcessSmartContract_test.go
```diff
@@ -243,3 +283,323 @@ func TestTXProcessor_smartContract(t *testing.T) {
 		})
 	}
 }
+
+func TestTXProcessor_SCDeployTimeoutBoundaryDeterministic(t *testing.T) {
```
nickgs1337: I reverted host.go to develop and this test still passes, 12/12 timeouts in the same ~12s. It calibrates a loop to ~500ms then runs it at 1.10x, so every attempt lands at 501ms against a 500ms deadline and the inner select this PR adds is never reached. Can we drop it? The hostCore unit tests already cover the tie-break branch, without burning ~30 real wasm deploys of wall-clock calibration per run.

nickgs1337: If this stays in any form it needs a `testing.Short()` skip. CI runs `go test -short ./...` (pr-qa-sec.yaml:43), so it executes on every PR under cross-package contention and a 10% margin won't hold there. The nastier failure is `require.Truef(t, outcome.deployed, "calibration deployment failed")` going red on a loaded runner, which reads like a VM bug rather than a busy box.

core/process/transaction/txProcessSmartContract_test.go
```diff
@@ -1,28 +1,68 @@
...
 
 	"github.com/klever-io/klever-go/common/mock"
 )
 
+// Precompiled module that imports smallIntGetUnsignedArgument, exports memory/init/upgrade,
+// and makes init loop for the number of iterations in argument 0. Embedding it keeps the
+// timeout regression independent of an external WAT compiler.
+const scDeployTimeoutBoundaryBusyInitWASMHex = "0061736d0100000001090260000060017f017e02230103656e761b736d616c6c496e74476574556e7369676e6564417267756d656e74000103030200000503010001071b0304696e6974000107757067726164650002066d656d6f727902000a1e021901027e4100100021010340200042017c22002001540d000b0b02000b"
+
+const scDeployTimeoutMainnetExecutionTimeoutMs = uint32(500)
```
nickgs1337: Naming: we name tests for the invariant, not the fix, and this const bakes "mainnet" into a test that builds its own host with its own timeout. If the mainnet value ever moves, the name is just wrong.

core/process/transaction/txProcessSmartContract_test.go
```diff
@@ -243,3 +283,323 @@ func TestTXProcessor_smartContract(t *testing.T) {
 		})
 	}
 }
+
+func TestTXProcessor_SCDeployTimeoutBoundaryDeterministic(t *testing.T) {
+	wasmCode := decodeSCDeployTimeoutBusyInitWASM(t)
+	disabledOutcome := runSCDeployTimeoutDeployment(t, wasmCode, 1,
```
nickgs1337: Asserting "deployment is disabled" here isn't really about timeout determinism, and it's the only reason `runSCDeployTimeoutDeployment` carries `isGenesisProcessing` (every other call passes true). Any reason not to drop the block and the param?

core/process/transaction/txProcessSmartContract_test.go
```diff
@@ -243,3 +283,323 @@ func TestTXProcessor_smartContract(t *testing.T) {
...
+func decodeSCDeployTimeoutBusyInitWASM(t *testing.T) []byte {
+	t.Helper()
+
+	wasmCode, err := hex.DecodeString(scDeployTimeoutBoundaryBusyInitWASMHex)
+	require.NoError(t, err)
+	require.NotEmpty(t, wasmCode)
+	return wasmCode
+}
+
+func calibrateSCDeployTimeoutLoopCount(t *testing.T, wasmCode []byte) uint64 {
```
nickgs1337: If the test survives, the binary search should go with it. `TestTimeoutRaceFix` in kvm/vmhost/hostCore/timeout_race_test.go:124 forces the same timeout with a hardcoded `encodeU64ForTest(10000)` and no calibration. All this needs is a loop count big enough that it can never finish inside the deadline, and then the assertion stops depending on how fast the machine is.

core/process/transaction/txProcessSmartContract_test.go
```diff
@@ -243,3 +283,323 @@ func TestTXProcessor_smartContract(t *testing.T) {
...
+
+func runSCDeployTimeoutDeployment(
+	t *testing.T,
+	wasmCode []byte,
+	loopCount uint64,
+	timeoutMs uint32,
+	isGenesisProcessing bool,
+) scDeployTimeoutOutcome {
+	t.Helper()
+	world := worldmock.NewMockWorld()
```
nickgs1337: nit: `testcommon.NewTestHostBuilder(t).WithBlockchainHook(world).WithBuiltinFunctions().Build()` (kvm/testcommon/testHostBuilder.go:33) already assembles all of this. Adding a `WithExecutionTimeout` option for the three timeout fields would be cheaper than hand-building the other 30 lines.

core/process/transaction/txProcessSmartContract_test.go
```diff
@@ -243,3 +283,323 @@ func TestTXProcessor_smartContract(t *testing.T) {
...
+func encodeSCDeployTimeoutDeployData(wasmCode []byte, loopCount uint64) string {
+	return strings.Join([]string{
+		hex.EncodeToString(wasmCode),
+		hex.EncodeToString(common.WasmVirtualMachine),
+		hex.EncodeToString((&vmcommon.CodeMetadata{Upgradeable: true, Payable: true}).ToBytes()),
+		hex.EncodeToString(encodeSCDeployTimeoutUint64(loopCount)),
+	}, "@")
+}
+
+func encodeSCDeployTimeoutUint64(value uint64) []byte {
```
nickgs1337: nit: `binary.BigEndian.AppendUint64(nil, loopCount)` inlined at the one call site does the same job.

core/process/transaction/txProcessSmartContract_test.go
```diff
@@ -243,3 +283,323 @@ func TestTXProcessor_smartContract(t *testing.T) {
...
+		},
+		ResetCountersCalled:    func() {},
+		GetCounterValuesCalled: func() map[string]uint64 { return map[string]uint64{} },
+		GetKAppControllerCalled: func() kapp.KAppController {
+			return world.KAppController
+		},
+	}
+}
+
+func scDeployTimeoutProtectedKeyPrefixes() [][]byte {
```
nickgs1337: nit: fifth copy of this same four-prefix list (testHostBuilder.go:35, scenarioexec/exec.go:95, vmContainerFactory.go:289 and more). Not this PR's job to fix, but exporting one from `kapps` is worth its own small ticket.

core/process/transaction/txProcessSmartContract_test.go
```diff
@@ -243,3 +283,323 @@ func TestTXProcessor_smartContract(t *testing.T) {
...
+	return [][]byte{
+		[]byte(kapps.ProtectedKleverKeyPrefix),
+		[]byte(kapps.ProtectedKLVKeyPrefix),
+		[]byte(kapps.ProtectedKFIKeyPrefix),
+		[]byte(kapps.KDAPrefix),
+	}
+}
+
+func maxSCDeployTimeoutDuration(left, right time.Duration) time.Duration {
```
nickgs1337: We're on go 1.25, so this is the `max()` builtin and `absSCDeployTimeoutDuration` is `x.Abs()` on time.Duration. Same idea at line 375, `time.Duration(1<<63 - 1)` is `math.MaxInt64`. Hand-rolled copies make you stop and check they really do what the name says, so 13 lines here buy us nothing.

core/process/transaction/txProcessSmartContract_test.go
```diff
@@ -243,3 +283,323 @@ func TestTXProcessor_smartContract(t *testing.T) {
...
+}
+
+func absSCDeployTimeoutDuration(duration time.Duration) time.Duration {
+	if duration < 0 {
+		return -duration
+	}
+	return duration
+}
+
+func formatSCDeployTimeoutDurations(outcomes []scDeployTimeoutOutcome) string {
```
nickgs1337: nit: these two exist only to build one `t.Logf` string that nothing asserts on. Could go along with the log line.

kvm/vmhost/hostCore/timeout_tiebreaker_test.go
```diff
@@ -0,0 +1,86 @@
...
+	require.Nil(t, runtimeCtx.FailExecutionErr)
+
+	select {
+	case <-hookCtx.Done():
+		t.Fatalf("hook context should not be canceled on normal completion")
+	default:
+	}
+}
+
+func TestVmHost_WaitExecutionWithDeterministicTimeout_TimeoutOnly(t *testing.T) {
```
nickgs1337: This is line for line `..._TieBreakerTimeoutWins` minus the loop, same cancelled ctx, same pre-closed `done`, same three assertions. It also isn't timeout-only: `done` is already closed when we call, so it's the tie case under a name that says otherwise. Better than deleting it, rename it and close `done` from a goroutine after a short delay, so we actually cover the case where execution is still in flight and `handleTimeout` blocks on `<-done`. Nothing covers that today.

kvm/vmhost/hostCore/host.go
```diff
@@ -464,6 +464,27 @@ func (host *vmHost) handleTimeout(cancelHook context.CancelFunc, done <-chan str
...
+		return host.handleTimeout(cancelHook, done)
+	case <-done:
+		// Deterministic tie-breaker: if timeout is also ready, timeout wins.
+		select {
+		case <-ctx.Done():
+			return host.handleTimeout(cancelHook, done)
+		default:
+			return nil
+		}
+	}
```
nickgs1337, line 495: Resolving the tie toward timeout worries me, the two signals aren't symmetric. `done` closed is proof the work finished inside budget, while `ctx.Done()` being observable also reflects how long this goroutine sat descheduled (GC pause, CPU contention). So a node that stalls at the wrong moment now reports failure for a deploy that genuinely finished in budget 100% of the time instead of ~50%. I measured it on this tree with work completing at 10ms of a 100ms budget and the observer stalled past the deadline: 52% success before, 0/50 after. At process.go:637 that's `ProcessIfError`, deploy rolled back, sender charged, so the stalled validator writes a different state root than its peers. Can we flip it so completion wins (check `done` first, then select)? That also points the same way as `getEffectiveTimeout`'s validator tolerance, which exists to bias validators toward accepting the leader's result. If timeout-wins is deliberate, it needs to be written down and argued against that tolerance, because right now the two mechanisms disagree.

kvm/vmhost/hostCore/host.go
```diff
@@ -464,6 +464,27 @@ func (host *vmHost) handleTimeout(cancelHook context.CancelFunc, done <-chan str
...
+	done <-chan struct{},
+) error {
+	select {
+	case <-ctx.Done():
+		return host.handleTimeout(cancelHook, done)
+	case <-done:
+		// Deterministic tie-breaker: if timeout is also ready, timeout wins.
+		select {
+		case <-ctx.Done():
+			return host.handleTimeout(cancelHook, done)
```
nickgs1337: In this branch `done` is already closed, so the goroutine has finished and assigned `vmOutput`. `handleTimeout` then calls `FailExecution`, and since `GetVMOutput` hands back the live `outputState` pointer that retroactively rewrites ReturnCode/ReturnMessage on an output that already went through `UpdateGasStateOnSuccess`. Result is a VMOutput neither path produces: `VMExecutionFailed` carrying success-path `GasRemaining`. No live exploit today since the deploy path discards vmOutput on error, but that rests on every current and future caller ignoring gas fields on the error branch, which nobody has stated as an invariant. Goes away entirely if the tie-break gets inverted.

kvm/vmhost/hostCore/host.go
```diff
@@ -464,6 +464,27 @@ func (host *vmHost) handleTimeout(cancelHook context.CancelFunc, done <-chan str
...
+	ctx context.Context,
+	cancelHook context.CancelFunc,
+	done <-chan struct{},
+) error {
+	select {
+	case <-ctx.Done():
+		return host.handleTimeout(cancelHook, done)
+	case <-done:
+		// Deterministic tie-breaker: if timeout is also ready, timeout wins.
+		select {
```
nickgs1337, line 489: nit: this inner select is `if ctx.Err() != nil`. The whole function collapses to one select on done/ctx, then `if ctx.Err() == nil { return nil }` before falling through to `return host.handleTimeout(cancelHook, done)`.

kvm/vmhost/hostCore/host.go
```diff
@@ -464,6 +464,27 @@ func (host *vmHost) handleTimeout(cancelHook context.CancelFunc, done <-chan str
 	return vmhost.ErrExecutionFailedWithTimeout
 }
 
+// waitExecutionWithDeterministicTimeout waits for execution completion or timeout.
+// If both completion and timeout are observable at the same boundary, timeout wins.
+func (host *vmHost) waitExecutionWithDeterministicTimeout(
+	ctx context.Context,
+	cancelHook context.CancelFunc,
```
nickgs1337, line 481: nit: this isn't a hook cancel. Both call sites pass the full cleanup closure from `setupExecutionContext`, which also cancels the main timeout ctx and nils `host.executionContext`, so "now safe to cancel hook context" in handleTimeout understates it a lot. Worth renaming the param to `cleanup` while we're in here.

kvm/vmhost/hostCore/host.go
```diff
@@ -526,12 +547,9 @@ func (host *vmHost) RunSmartContractCreate(input *vmcommon.ContractCreateInput)
 		host.logFromGasTracer("init")
 	}()
 
-	select {
-	case <-done:
-		// Normal termination
-		return
-	case <-ctx.Done():
-		err = host.handleTimeout(cancel, done)
+	timeoutErr := host.waitExecutionWithDeterministicTimeout(ctx, cancel, done)
```
nickgs1337: The `cancel` we pass here is the cleanup closure, and it nils `host.executionContext` in the same call. Hooks resolve their context through exactly that field (timeoutWrapper.go:29-42), so from the moment `handleTimeout` runs they see nil, skip the panic pre-check and drop into the untimed `rp := <-done` at line 91, while we sit parked on `<-done`. That makes the "hooks detect ctxHook.Done() and panic" flow the comments describe about a nanosecond wide and effectively unreachable, the WASM breakpoint is doing all the work. The write is unsynchronised too, plain field, `mutExecution` only held RLock, read by the still-running execution goroutine. Could `setupExecutionContext` return `cancelHook` separately so `handleTimeout` only cancels the hook ctx and the teardown stays with the deferred cleanup?

nickgs1337:
On scope: the deadline is still wall clock (`context.WithTimeout` in `setupExecutionContext`, host.go:430) over a budget that's itself mode-dependent, 500ms leader vs 575ms validator. Two validators running the same deploy still differ by machine speed, cache warmth and load, so one landing at 480ms and another at 520ms produces different outcomes exactly like before. What this PR removes is Go's random pick when both select cases are ready, a purely local artifact, which is the rarest contributor to the divergence the ticket title names. Not asking for the rewrite here (a real bound would key off gas/instructions, which the VM already meters, rather than elapsed time), but the description shouldn't claim the nondeterministic-outcome problem is solved. Suggest retitling to what it does and opening a separate item for the wall-clock-in-consensus part, otherwise the ticket closes and the real bug stays open.

core/process/transaction/txProcessSmartContract_test.go
```diff
@@ -1,28 +1,68 @@
...
 	"github.com/klever-io/klever-go/data/state"
 	"github.com/klever-io/klever-go/vmcommon"
 
 	"github.com/klever-io/klever-go/common"
+	commonMock "github.com/klever-io/klever-go/common/mock"
+	nodeConfig "github.com/klever-io/klever-go/config"
+	"github.com/klever-io/klever-go/core/fork"
+	smartContract "github.com/klever-io/klever-go/core/process/smartContract"
+	dataTransaction "github.com/klever-io/klever-go/data/transaction"
+	notifierMock "github.com/klever-io/klever-go/eventNotifier/mock"
```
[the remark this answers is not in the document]
RomuloSiebra: Two aliases for the same import path is legal Go — it compiled and vetted fine. Moot either way, that test is gone now.

core/process/transaction/txProcessSmartContract_test.go
RomuloSiebra: You're right, and I reproduced it — reverted host.go to develop and the test still passes, 12/12. Dropped it. The hostCore tests cover the tie-break.

RomuloSiebra: Also confirmed — it failed 4 of 5 local runs, calibrating under load and then landing at ~440ms once the box quietened. Gone now, so no skip needed.

core/process/transaction/txProcessSmartContract_test.go
RomuloSiebra: Agreed on both counts. Went away with the test.

core/process/transaction/txProcessSmartContract_test.go
RomuloSiebra: No reason, it was scope creep. Gone with the test.

core/process/transaction/txProcessSmartContract_test.go
RomuloSiebra: Yep — dropped along with it.

core/process/transaction/txProcessSmartContract_test.go
RomuloSiebra: Fair. Worth noting it isn't quite a drop-in — the builder makes its own forkController and epochNotifier and doesn't expose them, and the test needed the same instances. Moot now anyway.

core/process/transaction/txProcessSmartContract_test.go
RomuloSiebra: Right. Gone with the test.

core/process/transaction/txProcessSmartContract_test.go
RomuloSiebra: Our copy went with the test, so it's back to four. Agreed the export would be nice, though nothing in this PR needs it now.

core/process/transaction/txProcessSmartContract_test.go
RomuloSiebra: Agreed, all three. Gone with the test.

core/process/transaction/txProcessSmartContract_test.go
RomuloSiebra: Yep, they went with the log line.

kvm/vmhost/hostCore/timeout_tiebreaker_test.go
RomuloSiebra: Done — renamed, and it now closes done from a goroutine after a delay so handleTimeout actually blocks on `<-done` with execution in flight.

kvm/vmhost/hostCore/host.go
RomuloSiebra: Convinced, flipped it so completion wins. What settled it was txResultValidation — validator-OK/leader-failed gets the tolerance band and can accept, but leader-OK/validator-timeout is a straight mismatch rejection with no band, so timeout-wins was steering into the branch with no recovery. Same instinct as the new import-db replay timeout, too.

kvm/vmhost/hostCore/host.go
RomuloSiebra: Good catch. That branch doesn't exist any more now that completion wins the tie.

kvm/vmhost/hostCore/host.go
RomuloSiebra: That held for timeout-wins, but the flip inverted it — the inner check is on done now, and a channel has no `ctx.Err()` equivalent, so select/default is the only way to poll it. Left as is.

kvm/vmhost/hostCore/host.go
RomuloSiebra: Fair, it is the full cleanup closure today. Leaving it as a nit though — the name goes back to being accurate once the teardown issue is sorted out separately.

kvm/vmhost/hostCore/host.go
RomuloSiebra: Good analysis, agreed. Tracked as KLC-2608 rather than folded in here: the fix shifts unwind behaviour on the deploy path, so it wants its own review and validation rather than riding along on this one.

kvm/vmhost/hostCore/host.go
```diff
@@ -472,6 +472,29 @@ func (host *vmHost) handleTimeout(cancelHook context.CancelFunc, done <-chan str
...
+	select {
+	case <-done:
+		return nil
+	case <-ctx.Done():
+		// Deterministic tie-breaker: if execution also completed, completion wins.
+		select {
+		case <-done:
+			return nil
+		default:
+			return host.handleTimeout(cancelHook, done)
```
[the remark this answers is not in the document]
RomuloSiebra: Real, but inherent to polling two signals rather than something the flip introduced — the same window exists mirrored in the timeout-wins version. Making timeout final means dropping the inner select, which is develop's code and brings the coin-flip back at a true tie. The proper fix is your other suggestion, recording completion against the deadline instead of polling, which is the timeout rework tracked separately; this PR narrows the window rather than closing it.

kvm/vmhost/hostCore/timeout_tiebreaker_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+
+	hookCtx, cancelHook := context.WithCancel(context.Background())
+	done := make(chan struct{})
+	go func() {
+		time.Sleep(20 * time.Millisecond)
+		close(done)
+	}()
```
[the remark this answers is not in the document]
RomuloSiebra: Good catch, and a convention violation too. Swapped the sleep for a FailExecution hook on the runtime mock, so done closes only once handleTimeout has actually started.

kvm/vmhost/hostCore/host.go
```diff
@@ -534,12 +557,9 @@ func (host *vmHost) RunSmartContractCreate(input *vmcommon.ContractCreateInput)
 		host.logFromGasTracer("init")
 	}()
 
-	select {
-	case <-done:
-		// Normal termination
-		return
-	case <-ctx.Done():
-		err = host.handleTimeout(cancel, done)
+	timeoutErr := host.waitExecutionWithDeterministicCompletion(ctx, cancel, done)
```
nickgs1337: nit: this folds into the `if` at both call sites. `if timeoutErr := host.waitExecutionWithDeterministicCompletion(ctx, cancel, done); timeoutErr != nil { err = timeoutErr }` keeps the variable scoped to where it is actually used.

nickgs1337:
Separate from the tie-break, on the genuine timeout path where the goroutine is still running: `FailExecution` reads the instance twice, `check.IfNil(context.iTracker.Instance())` at contexts/runtime.go:681 and then `SetRuntimeBreakpointValue` reads `tracker.instance` again at 705. `Instance()` is a bare field read, so if the goroutine finishes and runs `EndExecution` -> `UnsetInstance` in between, the second read hits a nil interface, on the main goroutine, which has no recover. Reading it once into a local closes the nil deref cheaply. It does not close the underlying unsynchronised access to `tracker.instance`, and I could not force the interleaving, so someone who knows the wasmer teardown ordering should rule on how reachable it is. Either way the comment above `handleTimeout` claims sequential execution "prevents race condition", which this PR does not establish, so I would soften that while we are here.

nickgs1337:
Context worth having while we argue about goroutine ordering: `-race` cannot run on this package at all. `go test -race ./kvm/vmhost/hostCore/ -run Timeout` aborts on `TestTimeoutRaceFix` with `fatal error: checkptr: pointer arithmetic result points to invalid allocation` out of `wasmer2.getVMHooksFromContextRawPtr`, so the binary dies on the first real wasm execution and every timeout test is unreachable under the detector. Pre-existing and nothing this diff touches, but it means a green CI says nothing about the concurrency properties this PR is about, and we have a `timeout_race_test.go` we cannot actually run under `-race`. Worth its own ticket, either `//go:nocheckptr` on the cgo helper or reshaping the arithmetic through `unsafe.Add`/`unsafe.Slice`.

kvm/vmhost/hostCore/host.go
RomuloSiebra: Done, folded into the if at both call sites.

kvm/mock/context/runtimeContextMock.go
```diff
@@ -42,6 +42,7 @@ type RuntimeContextMock struct {
 	InstanceMock             executor.Instance
 	SignalErrorMessage       string
 	FailExecutionErr         error
+	FailExecutionCalled      func(err error)
```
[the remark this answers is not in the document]
RomuloSiebra: Every RuntimeContextMock literal in the repo is keyed, and the module builds and tests clean with the field added — an unkeyed one would have failed to compile. Keeping it, since the Called-hook pattern is what the other mocks here use.

kvm/vmhost/hostCore/host.go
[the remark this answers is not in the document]
RomuloSiebra: Tracked under KLC-2516, which stays open for the timeout mechanism work rather than being closed by this PR.

n4t4l (commented):
Went through all three files. Verified:

- The refactor is behaviour-preserving at both call sites — `handleTimeout` always returns non-nil, so `if timeoutErr != nil { err = timeoutErr }` is the old unconditional assignment.
- No new race on the named returns: every write from the execution goroutine is ordered by `close(done)`, and the waiter only writes after observing it (or after `handleTimeout`'s `<-done`).
- Tie direction matches the existing validator bias in `getEffectiveTimeout`.
- `_TieBreakerCompletionWins` genuinely discriminates — it fails ~50/100 with the inner select removed — and the timeout test now synchronises on `FailExecutionCalled` instead of sleeping.

Happy to approve once @nickgs1337's open points are dealt with:

1. host.go:489 — inner select collapses to `if ctx.Err() == nil { return nil }`; and `cancelHook` is really the full cleanup closure, worth renaming to `cleanup`.
2. `cleanup` nils `host.executionContext` unsynchronised, so hooks resolve a nil context and the hook-panic path is effectively unreachable.
3. `handleTimeout` rewriting a VMOutput that already went through the success path. One correction on that thread: inverting the tie-break narrows this to the nanoseconds between the `default:` poll and `<-done`, it doesn't close it — a completion landing in that window still yields `VMExecutionFailed` carrying success-path `GasRemaining`.

kvm/vmhost/hostCore/host.go
fbsobreira: This applied to the timeout-wins version, where the inner select was polling `ctx` and really was `if ctx.Err() != nil`.

After the flip the inner select polls `done`. A receive-only channel has no `Err()`, so `select`/`default` is the non-blocking check. Collapsing it to `if ctx.Err() == nil` would drop the tie-break and bring the coin-flip back. Leaving the select.

kvm/vmhost/hostCore/host.go
fbsobreira: The name is wrong today: both call sites pass the full cleanup from `setupExecutionContext`, not `cancelHook` alone.

Renaming the parameter here fights [KLC-2608](https://klever-io.atlassian.net/browse/KLC-2608). That split returns `cancelHook` on its own and leaves teardown on the deferred cleanup, which makes the current name accurate. Renaming to `cleanup` on this PR means renaming it back there. Not taking it on this diff.

kvm/vmhost/hostCore/host.go
fbsobreira: Confirmed on `d1f1f35`. `handleTimeout` still runs the cleanup closure before `<-done`, so hooks read a nil `executionContext` through `GetExecutionContext` and `FailAfterTimeout` falls into the untimed path. The nil write is still unsynchronised with the execution goroutine.

This is pre-existing on develop, not introduced by the tie-break. [KLC-2608](https://klever-io.atlassian.net/browse/KLC-2608) tracks the split (return `cancelHook` separately, and align the deploy recover with the call path). Keeping it out of this PR so the precedence change stays small. Resolving here as tracked there.

kvm/mock/context/runtimeContextMock.go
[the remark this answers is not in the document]
fbsobreira: `kvm/mock/context` is an in-module test mock, not a stable external API. Every `RuntimeContextMock` literal in this repo is keyed or empty, and `FailExecutionCalled` matches the other `Called` hooks on these mocks. An unkeyed literal of a struct imported from another module is already a `go vet` failure. Adding the field does not break any caller we build. Keeping it.

kvm/vmhost/hostCore/host.go
fbsobreira: The tie inversion does remove the case this comment described: `done` already closed, then `FailExecution` rewrites an output that already went through the success path.

It does not close the narrower window. If `ctx.Done()` is selected and the inner `default` runs before `done` is closed, `handleTimeout` can still `FailExecution` on an output that completes in that gap. Closing that needs a completion mark compared with the deadline, which is the rest of the timeout model, not this select tie-break. The original request (invert the tie) stands as done.

fbsobreira:
Went through the open threads against `d1f1f35`.

The tie-break is the right fix for KLC-2516. When `done` and `ctx.Done()` are both ready, completion wins. That matches the validator tolerance, and it avoids the leader-OK / validator-timeout mismatch, which has no recovery band. The hostCore tests hit that branch; the old transaction-level test did not, and it is gone.

On the three points still listed as open:

1. The inner select does not collapse to `if ctx.Err() == nil`. That was true for timeout-wins. It now polls `done`, which has to be a `select`/`default`. The `cancelHook` name is wrong today, but [KLC-2608](https://klever-io.atlassian.net/browse/KLC-2608) passes only the hook cancel into `handleTimeout`, so renaming it here gets undone.

2. The nil `executionContext` teardown is real and pre-existing. KLC-2608 has the split and the deploy-recover alignment. It should not ride on this diff.

3. The `FailExecution` rewrite of a finished output is only narrowed, not gone. The leftover window is completion landing between the inner `default` and `<-done`. Closing it means recording completion against the deadline, which is the rest of the timeout work, not this precedence change.

CodeRabbit's changes-requested reviews were stale (deleted WAT test, sleep already replaced, mock-field compatibility that does not apply here). Dismissed them.

Nothing else on the diff. Good to merge once a human approval is on it. KLC-2608 stays open separately.

[KLC-2608]: https://klever-io.atlassian.net/browse/KLC-2608?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

fbsobreira (approved):

PR #86, opened 2026-07-19
feat: validator node version attestation with observer demotion

Test0rMaik:
# feat: validator node version attestation with observer demotion

## Problem

Today, "mandatory" node updates are only latently enforced. An epoch flag activates a fork, but a validator running an outdated binary keeps producing identical blocks until the changed code path is actually exercised by a transaction. At that unpredictable moment, all outdated validators fork off simultaneously, miss blocks, bleed rating and get jailed — with an unjail fee — despite the operator having had no on-chain signal that they were at risk. The penalty is real but delayed and concentrated into "one bad day" (recent example: the long-outdated "Jon-Snow" node staying electable through several mandatory updates).

There is currently no provable, chain-data-based way to know a validator's version: p2p handshakes and endpoint checks are unauthenticated and not consensus data, so nothing can be safely gated on them — as correctly pointed out by the team in past discussions. This PR makes the version a piece of signed chain data and uses it leniently.

## Approach

**1. On-chain version attestation.** `ValidatorConfig` gets a new optional `NodeVersion` field (field 9). The validator owner attests the running node version with a normal, signed ValidatorConfig transaction; it is stored in `ValidatorData` (`AttestedVersion`, `AttestedEpoch`). This answers "don't trust, verify": the requirement source and every attestation are on-chain and owner-signed.

**2. Observer demotion instead of latent jailing.** During end-of-epoch processing (`ProcessEconomicsEndOfEpochV1/V2` → `updatePeerListStatus`), when a specific version is required for the new epoch, validators whose attested version does not satisfy it are moved to the (previously unused for this purpose) `observer` list:

- excluded from election (`computeNodesConfigFromList` ignores observers),
- **not** jailed, no rating loss, no unjail fee,
- restored to `eligible` automatically at the first end-of-epoch after attesting a satisfying version.

The operator loses rewards while outdated — nothing else. Jail semantics for actual misbehavior are untouched.

**3. Required version comes from the existing `versions.versionsByEpochs` config** — the same release-shipped table already consumed by `headerIntegrityVerifier`. A release that ships `{ startEpoch: E, version: "vX.Y.Z" }` thereby declares the mandatory version for epoch E; the shipped-by-default `"*"` wildcard keeps the whole mechanism dormant. Satisfaction is semver-aware (`attested >= required`), so early upgraders are never demoted; non-parsable tags fall back to exact match.

**4. Liveness guards.** Demotion is only enforced when BOTH hold, measured over the electable set (elected + eligible — exactly the population demotion removes from; waiting/inactive/jailed validators neither count nor get demoted):

- *supermajority*: ≥ 2/3 of the electable set already attested a satisfying version, bounding demotion to < 1/3 of the consensus-carrying validators, and
- *floor*: the attested validators alone still meet the nodes shuffler's minimum electable count (genesis `MinNumberOfNodes`, wired through `StateComponentsFactoryArgs`), so demotion can never trigger `ErrSmallElectedListSize`/`ErrListSizeZero`.

When the guards don't hold, nothing is demoted (log line only) — and already-demoted observers stay demoted until they attest, so enforcement cannot oscillate. The on-chain attestation share doubles as a rollout signal before the epoch arrives.

**4b. Attestation freshness.** An attestation only counts if it was made at or after the start epoch of the *previous* `versionsByEpochs` entry — i.e. validators must re-attest once per release cycle. A one-time inflated attestation (`99.9.9`) cannot grant a permanent exemption, and `AttestedEpoch` is consensus-relevant, not decorative. Pre-release tags (`1.9.0-rc1`) do not satisfy a release requirement (`1.9.0`).

**5. Fork-gated.** Everything is behind a new `versionAttestation` enable-epoch flag (`enableEpochs.yaml`). Attestations are not even written to state before the flag epoch, so pre-fork state stays byte-identical for old binaries (proto3 unknown fields are ignored by old nodes; new nodes ignore the field until the flag).

## What this changes for operators

After updating, run one `ValidatorConfig` tx with `nodeVersion` set (tooling/SDK can automate this on node startup). Not attesting after a mandatory release ⇒ demoted to observer at the mandatory epoch (if 2/3 attested) ⇒ no rewards until attested, then automatic return. Attesting falsely (new version string, old binary) ⇒ same risk profile as today: jailed when the fork path triggers. Incentives favor honesty; honest-but-slow operators are protected from jail.

## Adversarial review before submission

Two independent audit passes (correctness/logic and security/consensus-safety) were run against the initial implementation; both confirmed determinism, epoch alignment, replay protection, jail-flow isolation and proto wire-compat, and surfaced four real issues that are fixed in this PR's second commit:

1. **(critical)** the original supermajority guard counted waiting validators, decoupling the guarded population from the demotable one — cheap waiting attesters could flip enforcement on and demote a majority of the real consensus set, and even a legal ≤1/3 demotion could fall below the shuffler minimum and halt epoch preparation. Fixed by measuring over elected+eligible only plus the `MinNumberOfNodes` floor guard.
2. **(major)** `attested >= required` with no freshness check made a one-time `99.9.9` attestation a permanent exemption. Fixed by the per-release-cycle freshness rule.
3. **(minor)** a guard lapse used to silently restore still-outdated observers, allowing demote/restore oscillation. Fixed with hysteresis (restore requires attestation; disabling the mechanism restores everyone).
4. **(minor)** `1.9.0-rc1` satisfied `1.9.0`. Fixed (semver pre-release ordering).

## Determinism & safety notes

- Enforcement decision and demotions use only consensus state (kapp trie + peer lists) plus release-shipped config (`versionsByEpochs`, `enableEpochs`) — the same trust model as gas schedule configs. Order-independent: the supermajority is counted in a read-only pre-pass; per-validator demotion depends only on that validator's own data.
- Attestations sent before the fork epoch are deliberately ignored rather than rejected: old binaries accept the unknown proto field as `Transaction_Ok`, so any observable difference (error code or receipt) would fork state pre-activation. Operator tooling should attest at/after the fork epoch.
- The genesis path (`ProcessEconomicsEndOfEpoch(0, …)`) is safe: nobody has attested at genesis, so the guard keeps enforcement off.
- Timing: demotion runs in the epoch-start block *before* `ResetValidatorStatisticsAtNewEpoch` → `SetEpochValidatorsInfo`, so a version required at epoch E excludes non-attested validators from epoch E's election exactly at the fork boundary.
- Restoration cannot strand validators: observers stay in the peer trie and are iterated every end-of-epoch; `updatePeerListStatus` lifts them back to eligible as soon as they attest a satisfying version, or immediately if the mechanism is disabled (fork off / wildcard). A guard lapse alone deliberately does not restore unattested observers (hysteresis).

## Changes

- `data/transaction/proto/contracts.proto`: `ValidatorConfig.NodeVersion` (9)
- `core/kapp/validators/proto/validatorData.proto`: `AttestedVersion` (27), `AttestedEpoch` (28)
- `core/kapp/validators/versionAttestation.go` (new): required-version lookup with freshness epoch, semver compare (pre-release aware), supermajority + floor guards
- `core/kapp/validators/validators.go`: attestation handling in `UpdateValidator` (fork-gated, validated: utf8, ≤ `MaxSoftwareVersionLengthInBytes`)
- `core/kapp/validators/peersUpdate.go`: version branch in `updatePeerListStatus`, enforcement threaded through V1/V2 epoch processing
- `config/enableEpochs.go`, `config/node/enableEpochs.yaml`, `core/fork/forks.go`, `core/interface.go` + fork controller stubs: new `versionAttestation` flag
- `core/kapp/kappController/kapp.go`, `core/kapp/factory/validators.go`, `factory/stateComponents.go`, `cmd/node/startup.go`: wire `versions.versionsByEpochs` and genesis `MinNumberOfNodes` into the validators kapp
- `core/kapp/validators/versionAttestation_test.go` (new): 40+ cases covering parsing, comparison (incl. pre-release), epoch resolution and freshness, demotion/restore, hysteresis, supermajority boundary (incl. exact 2/3), waiting-attester exclusion, floor guard, stale attestations, fork gating, validation, and full `ProcessEconomicsEndOfEpoch` round-trips

## Known residual risks (accepted by design, documented for review)

- Attestation remains self-reported: an operator can attest a version they do not run. This is unchanged from the status quo (nothing on-chain can prove a binary), it is now bounded (must be repeated every release cycle, each one owner-signed on-chain), and a false attester keeps today's risk profile — jailed when the fork path triggers.
- Version demotion stacks with same-epoch jailing/stake-drops; the ≤1/3 bound covers only the version mechanism. The floor guard limits the combined worst case, and jail waves are a pre-existing phenomenon with the existing `numToStay` backfill.
- `versions.versionsByEpochs` becomes a state-relevant config (same trust model as gas schedule configs, and already network-uniform whenever non-wildcard because of the header `SoftwareVersion` equality check). `requiredVersionForEpoch` must keep agreeing with `headerIntegrityVerifier.getMatchingVersion`; both rely on the startup validation in `prepareVersions`.

## Complementary follow-ups (not in this PR)

1. Ship real versions in `versionsByEpochs` for mandatory releases (activates both this mechanism and the existing dormant header-version enforcement).
2. Longer term: unconditional state divergence at activation epochs (Cosmos-style upgrade heights) would remove the latent-fork window entirely; this PR's observer demotion then becomes the lenient handling for nodes that drop out at the scheduled moment.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary

Adds fork-gated validator node-version attestation and an end-of-epoch version enforcement mechanism in the validators KApp, affecting consensus-critical validator peer-list selection and persisting new attestation state.

- **Transaction processing / state integrity (KVM & receipts):** Extends `ValidatorConfig` with `nodeVersion` (`ValidatorConfig.NodeVersion`). When `VersionAttestation` fork is **inactive**, `UpdateValidator` intentionally **ignores** `NodeVersion` (not rejected, not recorded) to avoid pre-fork receipt/state divergence. When active, `UpdateValidator` validates UTF-8 and length (<= `core.MaxSoftwareVersionLengthInBytes`) and records `ValidatorData.attestedVersion` and `ValidatorData.attestedEpoch`. Adds receipt tracking constant `ErrFieldInvalidNodeVersion` for invalid node version inputs.
- **State management (validator persistence):** Extends `ValidatorData` with persisted `AttestedVersion` and `AttestedEpoch` fields.
- **Consensus-critical epoch behavior (peer-list enforcement):** Implements `versionAttestation.go` plus peer-list routing via `core/kapp/validators/peersUpdate.go`:
  - Enforcement is active only when the **`VersionAttestation` fork flag** is enabled for the epoch and `versionsByEpochs` specifies a **non-wildcard** required version for that epoch; wildcard/empty/missing => no requirement.
  - At epoch boundaries, validators in **elected/eligible** that do **not** meet the required version and attestation **freshness** window are moved to **`observer`** (not jailed/penalized); they return to **eligible** once they submit a suitable attested version.
  - Demotion is guarded to prevent destabilizing transitions: requires a **2/3 electable-set supermajority** of satisfied elected+eligible validators and a **minimum electable-node floor** (`MinElectableNodes`) so electable coverage cannot drop too far.
  - If the configuration is active but demotion is not allowed by the guards, the mechanism preserves existing membership to avoid oscillation (stake-threshold-driven transitions still take precedence).
- **Fork/config wiring & state-component determinism:** Adds `enableEpochs.versionAttestation` and a fork feature flag (`ForkController.VersionAttestation()`), and wires `VersionsByEpochs` plus `MinElectableNodes` through:
  - KApp/controller construction (`NewValidatorKApp`, `NewKappController`, state components factory),
  - explicit node startup state component initialization (`cmd/node/startup.go` sets `MinElectableNodes` from genesis),
  - and fork-controller toggling on `EpochConfirmed`.
- **Cross-cutting error handling / fail-closed behavior:** During enforcement guard computation, validator record loading failures (`getValidator`) are logged and treated as **unsatisfied** to ensure the mechanism errs on the side of not demoting incorrectly.
- **Tests:** Adds extensive unit coverage for semver parsing/satisfaction, epoch-to-version resolution and `minAttestedEpoch` freshness logic, enforcement guard activation (supermajority + floor), demotion/restoration behavior through `updatePeerListStatus`, `UpdateValidator` attestation persistence rules (including fork-active vs inactive), and end-of-epoch economics demotion/restore scenarios (including negative/guard-lapse cases).


Test0rMaik:
Addressed the automated review in 8958841:

- **`infos[2]` brittleness** (versionAttestation_test.go): fixed — the restore round-trip now locates `owner3` by `OwnerAddress` instead of relying on map-derived slice order.
- **Silent `continue` on `getValidator` failure** (versionAttestation.go guard loop): fixed — failures are now logged with the owner address. The fail-closed behavior is kept deliberately: an undercounted `satisfied` can only keep demotion *off*, never demote anyone spuriously.
- **Nested enforcement branch** (updatePeerListStatus): extracted into `resolveVersionEnforcedList`, single `SetList` call, semantics unchanged (covered by the existing demote/hold/restore test cases).

**Declined (with reasoning): caching `*ValidatorData` between `computeVersionEnforcement` and `processValidatorEpochV1/V2`.** The duplicate read is one extra kapp-storage lookup per electable validator, once per epoch, through the already-cached accounts layer. Sharing mutable `ValidatorData` records across the guard pre-pass and the mutating epoch loop would couple two consensus-critical code paths through shared state for a negligible saving — in this code I'd rather pay the read than the coupling. Happy to revisit if the team prefers otherwise.

Test0rMaik:
Rewrote the 3 commits to drop the `Co-Authored-By` trailer — noticed our house style doesn't use it in this repo, in line with the note on a recent PR. Commit content/diff is unchanged (also picked up the latest develop, including the x/crypto,x/net,quic-go dependency bump from #87). No action needed on the review comments above; they still apply to the equivalent lines.

Test0rMaik:
@coderabbitai review

core/kapp/validators/versionAttestation.go
```diff
@@ -0,0 +1,232 @@
...
+			continue
+		}
+		if enforcement.isSatisfiedBy(val) {
+			satisfied++
+		}
+	}
+
+	hasSupermajority := electable > 0 &&
+		satisfied*versionEnforcementDenominator >= electable*versionEnforcementNumerator
+	holdsFloor := v.minElectableNodes == 0 || uint32(satisfied) >= v.minElectableNodes // #nosec G115
```
MathijsBok: 🔴 **Critical** (F1): the guards cannot prevent an empty elected list, which permanently stalls the network

This is open issue #132, filed on 2026-08-13 and still unmitigated on this head.

Both guards reason about the **combined** electable set: `electable` and `satisfied` count `ElectedList` and `EligibleList` into the same counters (lines 183-207), and this floor compares that combined `satisfied` against `minElectableNodes`. Nothing constrains the **elected sublist on its own**.

`sharding/nodesCoordinator.go:942-944` aborts the entire epoch computation when the elected list alone is empty:

```go
if len(electedList) == 0 {
    return nil, fmt.Errorf("%w elected list size is zero. No validators found", ErrListSizeZero)
}
```

**Failure scenario that passes both guards.** 10 elected validators, all outdated; 30 eligible validators, all attested. `electable = 40`, `satisfied = 30`. Supermajority holds (`30*3 = 90 >= 40*2 = 80`), and the floor holds for any `minElectableNodes <= 30`. Demotion proceeds, all 10 elected validators become observers, and the elected list is empty.

**Why this does not recover.** `EpochStartPrepare` logs `could not compute nodes config from list` and returns at `sharding/nodesCoordinator.go:762-766`, so `nodesConfig[newEpoch]` is never built. Every subsequent `ComputeConsensusGroup` for that epoch then fails with `ErrEpochNodesConfigDoesNotExist` (`:511-513`), so no header of the new epoch validates on any intake path. Nothing ever closes that window, on any node.

This is not an exotic configuration: the elected set is the long-running incumbent population, which is exactly the group most likely to lag a mandatory upgrade.

**Also worth correcting:** the doc comment on this function states that the floor means demotion "can never reduce elected+eligible below what EpochStartPrepare needs". `EpochStartPrepare` needs the elected list specifically to be non-empty, which no combined count can guarantee.

**Suggested direction.** Either count the elected sublist separately here and refuse to demote when doing so would empty it, or change the coordinator to tolerate an empty elected list when the eligible list can still fill `consensusGroupSize` (the shuffler runs immediately afterwards at `:768-781` and repopulates elected from eligible). Issue #132 deliberately leaves that ownership question open, so please say which side you would rather fix; we are happy to take the coordinator half in a separate PR after this one lands. Whichever side is chosen, the abort branch needs a test.

*Ref: CWE-691 (insufficient control flow management)*

core/kapp/validators/versionAttestation.go
```diff
@@ -0,0 +1,232 @@
...
+		return false
+	}
+
+	for i := 0; i < len(attestedParts); i++ {
+		if attestedParts[i] != requiredParts[i] {
+			return attestedParts[i] > requiredParts[i]
+		}
+	}
+
+	return !attestedPre || requiredPre
```
MathijsBok: 🟡 **Warning** (F2): prerelease identifiers are never compared, so an older release candidate satisfies a newer one

When the three numeric components are equal, this line decides the outcome purely on the presence of a prerelease suffix on either side. The identifiers themselves (`rc1`, `rc2`, `alpha`) are dropped by `parseSemver`, so any prerelease satisfies any other prerelease of the same numeric version.

Verified against this head:

```
versionSatisfies("v1.9.0-rc1", "v1.9.0-rc2") == true
versionSatisfies("v1.9.0-alpha", "v1.9.0-rc1") == true
```

Both should be `false`: per semver, `1.9.0-alpha < 1.9.0-rc1 < 1.9.0-rc2`. This contradicts the documented contract of the function ("semantic compare (attested >= required)").

**Why it is reachable.** Required entries are capped at `core.MaxSoftwareVersionLengthInBytes` (10 bytes), and `v1.9.0-rc2` is exactly 10 bytes, so a prerelease requirement is a valid config entry, and our release tags do use rc suffixes. The effect is deterministic across nodes, so this is an enforcement gap rather than a divergence risk: validators still on an older rc of the same version keep their slots when a newer rc is required.

**Suggested fix.** Either compare prerelease identifiers per semver 11.4, or, simpler and adequate for a 10-byte domain, fall back to exact string match whenever the *required* version carries a prerelease suffix. The current test table covers prerelease against release in both directions but has no prerelease against prerelease case, which is why this slipped through; worth pinning `("v1.9.0-rc1", "v1.9.0-rc2") == false` and `("v1.9.0-rc2", "v1.9.0-rc1") == true` once fixed.

core/kapp/validators/validators.go
```diff
@@ -528,6 +540,22 @@ func (v *validatorsKApp) UpdateValidator(sender []byte, tc *transaction.Validato
 		val.Name = tc.GetConfig().GetName()
 	}
 
+	// NodeVersion is intentionally ignored (not rejected) before the versionAttestation
+	// fork: old binaries accept the unknown field as Transaction_Ok, so any observable
+	// difference (error code or receipt) would fork state pre-activation. Operators must
+	// attest at or after the fork epoch for the attestation to be recorded.
+	if len(tc.GetConfig().GetNodeVersion()) > 0 && v.forkController.VersionAttestation() {
```
MathijsBok: 🟡 **Warning** (F5): attestation-only transactions silently reset commission and delegation settings

This is about the interaction between the new block and the three unconditional assignments just below it (lines 559-561), which are pre-existing:

```go
val.CanDelegate = tc.GetConfig().GetCanDelegate()
val.Commission = tc.GetConfig().GetCommission()
val.MaxDelegation = tc.GetConfig().GetMaxDelegationAmount()
```

Unlike `RewardAddress`, `Logo`, `Name` and `URIs`, which are all guarded by `len(...) > 0` checks, these three are overwritten unconditionally, so an absent field means zero.

Before this PR that was mostly harmless, because `ValidatorConfig` transactions were occasional and deliberate. This PR makes them a **recurring operational requirement**: every validator has to submit one at least once per release cycle to stay electable. An operator, or a script, that sends the minimal attestation transaction (`BLSPublicKey` plus `NodeVersion`) will silently set commission to 0, set `CanDelegate` to false and clear `MaxDelegation`. There is no error and no receipt indicating it happened, and the economic consequences are real.

**Suggested fix.** At minimum, document that an attestation transaction must echo the validator's full current config, and make sure the official tooling does so. A cleaner long-term option, behind a future fork, is to treat zero values for these three fields as "no change", matching how the string fields already behave, or to give attestation its own contract type.

**Separately, on this same block:** it adds 5 points of cognitive complexity to `UpdateValidator`, taking it from 59 to 64 against this repo's threshold of 15. The added lines are themselves fully covered by the new tests, so this is purely about compounding an already oversized function. Extracting the block into a small helper (for example `applyNodeVersionAttestation`) removes the whole delta at zero behavior change and gives the attestation logic a directly testable unit. There is recent precedent on `develop` for exactly this kind of split (`cf516ee`, done to bring `createNodesCoordinator` under the threshold).

data/transaction/contracts.pb.go
```diff
@@ -1,8 +1,8 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.36.6
-// 	protoc        v5.29.3
-// source: contracts.proto
+// 	protoc        v6.33.1
+// source: data/transaction/proto/contracts.proto
```
MathijsBok: 🟡 **Warning** (F6): generated files were produced with a non-canonical protoc invocation

Both regenerated files (`data/transaction/contracts.pb.go` and `core/kapp/validators/validatorData.pb.go`) embed repo-root-relative source paths and were built with protoc v6.33.1, whereas the checked-in generation directives register the descriptors under bare filenames with protoc v5.29.3:

- `data/transaction/transaction.go:2`: `//go:generate protoc -I=proto ... contracts.proto`
- `core/kapp/validators/validatorData.go:1`: same pattern for `validatorData.proto`

Two consequences:

1. Every internal symbol was renamed (`file_contracts_proto_*` becomes `file_data_transaction_proto_contracts_proto_*`), which inflates this diff by roughly 2000 mechanical lines and makes the genuine changes hard to find. The next person who runs the canonical `go:generate` will revert all of it, producing churn and merge conflicts.
2. The protobuf registry file path changes. That path is a public identifier for anything resolving these descriptors by name through `protoregistry` or reflection. Nothing in this repo does today, but it is an unnecessary change to make silently.

The field additions themselves are correct and wire-compatible: fresh field numbers (`ValidatorConfig.NodeVersion = 9`, `ValidatorData.AttestedVersion = 27`, `AttestedEpoch = 28`), consistent `json_name` values, and accessors matching the `.proto` sources.

**Suggested fix.** Regenerate both files using the existing `//go:generate` directives, with protoc invoked as `-I=proto` plus the bare filename, and keep the protoc version aligned with the rest of the team. The diff should then shrink to just the new fields and their accessors.

core/kapp/validators/versionAttestation.go
```diff
@@ -0,0 +1,232 @@
...
+
+// requiredVersionForEpoch returns the version required for the given epoch based on the
+// versions.versionsByEpochs configuration, the minimum epoch an attestation must have
+// been made in to count for it, and whether a specific version is required at all.
+// A wildcard or missing entry means no requirement (mechanism stays dormant).
+//
+// NOTE: this table is the same release-shipped config validated and consumed by
+// headerCheck.headerIntegrityVerifier; both lookups must keep agreeing on which
+// entry covers an epoch.
+func (v *validatorsKApp) requiredVersionForEpoch(epoch uint32) (string, uint32, bool) {
```
MathijsBok, line 100: 💡 **Suggestion** (F10): the NOTE above this function states an invariant that nothing enforces

The comment correctly flags that this lookup and `headerCheck.headerIntegrityVerifier` must keep agreeing on which `versionsByEpochs` entry covers an epoch. They are, however, two independent implementations with different assumptions:

- `headerCheck.getMatchingVersion` (`core/process/headerCheck/headerIntegrityVerifier.go:103-120`) operates on input that `prepareVersions` has already sorted and validated (first entry at epoch 0, strictly ascending, length checked).
- `requiredVersionForEpoch` tolerates unsorted input and a missing epoch-0 entry, and additionally tracks the previous entry.

Token-level duplication is zero, so no static analysis will catch drift between them. A single agreement test would turn the comment into an enforced property: feed the same non-trivial `versionsByEpochs` fixture to both lookups and assert they select the same entry for a range of epochs.

For context, the semver logic itself duplicates nothing: no numeric version comparator existed before this PR (`core/statistics/softwareVersion` does plain string comparison, `headerCheck` does byte equality), so `parseSemver` and `versionSatisfies` are genuinely new ground.

core/kapp/validators/versionAttestation_test.go
```diff
@@ -0,0 +1,750 @@
...
+		}
+
+		resultCode, err := v.UpdateValidator(ownerAddress, tc)
+		require.NoError(t, err)
+		require.Equal(t, transaction.Transaction_Ok, resultCode)
+
+		app, _ := v.getKApp()
+		val, err := v.getValidator(app, ownerAddress)
+		require.NoError(t, err)
+		assert.Equal(t, "v1.9.0", val.AttestedVersion)
```
MathijsBok, line 748: 💡 **Suggestion** (F7): nothing pins that `AttestedEpoch` is recorded from the block header

This test asserts `AttestedVersion` only. No test in the suite covers the other half of what `UpdateValidator` writes at `validators.go:556`:

```go
val.AttestedEpoch = ctx.Block().GetHeader().GetEpoch()
```

That field is what the entire freshness mechanism depends on. The stale-attestation tests construct `ValidatorData.AttestedEpoch` by hand, so they exercise the comparison but not the recording. A regression in the recording path (storing 0, or reading the wrong header field) would leave the whole suite green while silently disabling per-release re-attestation in production: every attestation would look either permanently fresh or permanently stale, depending on the table.

**Suggested fix.** Configure a nonzero epoch on the mocked block header in this test's context and assert it round-trips:

```suggestion
		assert.Equal(t, "v1.9.0", val.AttestedVersion)
		assert.Equal(t, expectedEpoch, val.AttestedEpoch)
```

core/kapp/validators/versionAttestation_test.go
```diff
@@ -0,0 +1,750 @@
...
+				BLSPublicKey: []byte("blspubkey"),
+				NodeVersion:  "v1.9.0",
+			},
+		}
+
+		resultCode, err := v.UpdateValidator(ownerAddress, tc)
+		require.NoError(t, err)
+		require.Equal(t, transaction.Transaction_Ok, resultCode)
+
+		app, _ := v.getKApp()
```
MathijsBok: 💡 **Suggestion** (F8): `getKApp` error discarded here and at line 548

Every other `getKApp` call in this file checks its error (the restore flow later in the file uses `require.NoError`), and the project convention is that errors are never silently discarded, test setup included. If `getKApp` failed here, the following `getValidator` would fail with a confusing nil-handler error rather than pointing at the real cause.

```suggestion
		app, err := v.getKApp()
		require.NoError(t, err)
```

The same applies to the identical line in the "attestation is ignored before the fork" subtest at line 548.

core/kapp/validators/peersUpdate.go
```diff
@@ -687,8 +708,10 @@ func (v *validatorsKApp) ProcessEconomicsEndOfEpochV1(currentEpoch uint32, valid
 		return err
 	}
 
+	enforcement := v.computeVersionEnforcement(app, validatorInfos, currentEpoch)
+
 	for _, validatorInfo := range validatorInfos {
-		if err := v.processValidatorEpochV1(app, validatorInfo, currentEpoch, minSelfDelegated, minTotalDelegated, totalDelegations); err != nil {
+		if err := v.processValidatorEpochV1(app, validatorInfo, currentEpoch, minSelfDelegated, minTotalDelegated, totalDelegations, enforcement); err != nil {
```
MathijsBok, line 714: 💡 **Suggestion** (F9): these are the only lines this PR touches that no test covers

The `return err` bodies inside both epoch loops (lines 714-716 here, and 743-745 in `ProcessEconomicsEndOfEpochV2`) have zero coverage. The gap predates this PR, but because the call lines change here (the new `enforcement` argument), they now land in the new-code coverage denominator that the quality gate measures. Every other executable line this PR touches in non-generated code is covered.

The volume is small (two statements), so this is a cheap fix rather than an important one: one test per version that makes `processValidatorEpochV1`/`V2` fail, for example through a peer account load error, and asserts the error surfaces from `ProcessEconomicsEndOfEpochV1`/`V2`. There is a recent precedent for this shape of test on `develop` (`94bc4b1`, pinning `SetNodes` error propagation).

MathijsBok (changes requested):
## Review of the rebased branch (`85f95ca`)

Thanks for the rebase, it landed cleanly on the current `develop` tip and the branch is mergeable again. This is a full review of the feature as it now stands: every changed file read in full, the package tests and complexity metrics run against both this head and the merge base.

**Overall:** the mechanism is well built at the unit level. The attestation write path is correctly bound to the authenticated sender, `AttestedEpoch` is taken from the block header rather than the submitter, the pre-fork "ignore, do not reject" reasoning for `NodeVersion` is sound and avoids a pre-activation state fork, the guards fail closed on unreadable records, the fork wiring is complete across all four `ForkController` implementers, and the epoch alignment is exact. Test coverage of the new logic is 100% line coverage on every new function, and package coverage rises from 69.2% to 71.4%.

**Requesting changes** for one blocking item and a small number of correctness and hygiene points below.

### Blocking

- **F1** (critical): the demotion guards can empty the elected list, which permanently stalls epoch preparation. This is the open issue #132, and it is still unmitigated on this head. Details inline.

### Non-blocking findings

| ID | Severity | Location | Finding |
|----|----------|----------|---------|
| F2 | 🟡 Warning | `versionAttestation.go:145` | `versionSatisfies` ignores prerelease identifiers when the numeric parts are equal |
| F5 | 🟡 Warning | `validators.go:547` | Attestation-only transactions silently reset commission and delegation settings |
| F6 | 🟡 Warning | `contracts.pb.go:5` | Generated files produced with a non-canonical protoc invocation (~2000 lines of churn) |
| F7 | 💡 Suggestion | `versionAttestation_test.go:531` | No test pins that `AttestedEpoch` is recorded from the block header |
| F8 | 💡 Suggestion | `versionAttestation_test.go:528` | `getKApp` error discarded in two places |
| F9 | 💡 Suggestion | `peersUpdate.go:714` | The only uncovered lines this PR touches: error propagation in both epoch loops |
| F10 | 💡 Suggestion | `versionAttestation.go:56` | The NOTE about agreeing with `headerCheck` is not pinned by a test |

### Additional notes (no action required now)

- **Complexity:** `requiredVersionForEpoch` and `computeVersionEnforcement` both measure cognitive complexity 13 against this repo's threshold of 15, so each is two increments away from a quality gate finding. Given that both are consensus-critical and likely to gain guards later, extracting the tally loop out of `computeVersionEnforcement` (roughly 6 of its 13 points) would buy useful headroom. `UpdateValidator` is covered separately in F5.
- **Parameter counts:** `processValidatorEpochV1`/`V2` now take 7 parameters and `updatePeerListStatus` 6. Nothing exceeds a hard limit, but the next per-epoch input will force the issue; bundling the per-epoch invariants into a single struct would be the natural move at that point (not now).
- **Log levels:** the `version demotion skipped: safety guards not met` and `version demotion active` lines are at Debug. These fire at most once per epoch and describe a state operators will want to see in an incident; Info would fit them better.
- **Integration coverage:** `integrationTest/processorNode` and `kvm/mock/world` construct `ArgsNewKApp` without the new fields, so enforcement is dormant there. That is fine for correctness (zero values disable the feature) but it does mean demotion has no integration-level coverage.

Separately, I am sending you a short document by email covering a few further points that we do not discuss in public per our security policy. Nothing there blocks this PR on its own, but one of the items is closely related to F1 and worth reading before you rework the guards.

core/kapp/validators/versionAttestation.go
Test0rMaik: Fixed in b5a3212: `computeVersionEnforcement` now carries a third guard, `preservesElectedList`, that refuses demotion whenever it would leave the elected sublist with zero satisfying members. The guard is evaluated over a live re-derivation of each validator's current list membership (`peerAcc.GetList()`) rather than the epoch-start `validatorInfos` snapshot, so a validator reclassified earlier in the same epoch boundary (e.g. jailed by the ratings pass) can no longer inflate the guards either. Covered by new subtests in `TestComputeVersionEnforcement` ("demotion refused when it would empty the elected list", "demotion proceeds when at least one elected validator satisfies", "stale snapshot: a since-jailed validator does not inflate the guards"), each mutation-tested.

A follow-up commit, bf65553, further hardens the same function: unreadable validator/peer-account records were being excluded from the guard denominators entirely rather than counted as unsatisfied-but-electable, which could both shrink the supermajority guard's denominator and let the elected-list guard pass vacuously if the sole elected validator's record happened to be unreadable. Both are now fail-closed, pinned by two more subtests.

core/kapp/validators/versionAttestation.go
Test0rMaik: Fixed in 088195c: `parseSemver` now returns the prerelease identifier string itself (not just a presence flag), and a new `comparePrerelease` implements semver.org section 11.4 ordering — numeric fields compare numerically, alphanumeric fields as ASCII strings, numeric always outranks lower than alphanumeric, and a longer field list outranks an equal-prefix shorter one when both sides carry a prerelease. `versionSatisfies` now calls into it instead of just checking presence. Covered by a new `TestComparePrerelease` table and additional `TestVersionSatisfies" cases including your exact pinned examples (rc1/rc2/alpha).

core/kapp/validators/validators.go
Test0rMaik: Fixed in f1d5148: the inline NodeVersion-attestation block was extracted into its own `applyNodeVersionAttestation` method (bringing `UpdateValidator` back under the complexity threshold), and a comment now documents that `CanDelegate`/`Commission`/`MaxDelegation` are unconditionally overwritten on every `UpdateValidator` call, so callers (including attestation-only transactions) must echo the full config or those fields get zeroed.

data/transaction/contracts.pb.go
Test0rMaik: Fixed in 3c0eb2d: regenerated both files using the canonical invocation (`protoc -I=proto ... <bare-filename>.proto`, run from each proto's own package directory, matching the checked-in `//go:generate` directives), which brought the diffs down to just the intended new fields — 43 lines for validatorData.pb.go, 15 for contracts.pb.go. The protoc version used locally (v33.1) still differs from the original generation's (v5.29.3) in the header comment, which is an accepted cosmetic difference; the registry-path/symbol-renaming churn you flagged is gone.

core/kapp/validators/versionAttestation.go
Test0rMaik: Fixed in a555277: added `TestRequiredVersionForEpoch_AgreesWithHeaderIntegrityVerifier`, which feeds the same (deliberately unsorted) `versionsByEpochs` table to both `requiredVersionForEpoch` and `headerIntegrityVerifier.GetVersion` and asserts they select the same entry across a representative epoch range, turning the NOTE comment into an enforced property.

core/kapp/validators/versionAttestation_test.go
Test0rMaik: Fixed in 16cba7d: the "attestation is stored when fork is active" subtest now builds its own KappContext with a nonzero block epoch (rather than relying on the shared `addContext` helper's zero default) and asserts `val.AttestedEpoch` equals it — with the shared helper's zero epoch, a broken recording and a correct one were indistinguishable, which this fixes.

core/kapp/validators/versionAttestation_test.go
Test0rMaik: Fixed in 16cba7d: both call sites now use `app, err := v.getKApp()` with `require.NoError(t, err)` instead of discarding the error.

core/kapp/validators/peersUpdate.go
Test0rMaik: Fixed in 7c71910: added `TestProcessEconomicsEndOfEpoch_PropagatesPerValidatorError`, table-driven over V1/V2, which forces `LoadPeerCalled` to return an error and asserts it surfaces via `require.ErrorIs` from `ProcessEconomicsEndOfEpoch`. Mutation-tested by swallowing the loop error in each version independently to confirm the test actually catches the regression.

Test0rMaik:
Pushed a round of fixes for the rebased-branch review (`85f95ca` → `bf65553`), replied to each inline finding individually. Summary:

**F1 (blocking)**: `computeVersionEnforcement` now has a third guard, `preservesElectedList`, that refuses demotion whenever it would leave the elected sublist with zero satisfying members, evaluated against each validator's live peer-account list rather than the epoch-start snapshot (which closes a related staleness gap: a validator reclassified — e.g. jailed — earlier in the same epoch boundary no longer inflates the guards). The tally logic was extracted into its own `tallyElectableVersions` function, which incidentally addresses the complexity headroom note for `computeVersionEnforcement` as well.

**F2, F5, F6, F7, F8, F9, F10**: all fixed individually, see inline replies for details on each.

**Log levels**: `version demotion skipped`/`version demotion active` are now at Info.

**Parameter counts / integration coverage**: left as-is per your own note that these aren't blocking now.

While re-testing the F1 fix I found (and fixed, with new mutation-tested coverage) a related gap: `tallyElectableVersions` was excluding validators with unreadable records from the guard denominators entirely instead of counting them as unsatisfied-but-electable as the code's own comment claimed — which could both shrink the supermajority guard's denominator and let the elected-list guard pass vacuously if the sole elected validator's record was the unreadable one. Both are fail-closed now.

Full test suite is green (`go test ./core/kapp/validators/... ./core/process/headerCheck/...`), and `go build ./...` passes.

Issue #93, opened 2026-07-20
Consider a coordinated bump of github.com/herumi/bls-go-binary past v1.28.2


Test0rMaik:
`go.mod` currently pins `github.com/herumi/bls-go-binary v1.28.2` (direct dependency, used only by `crypto/signing/mcl/*` — the BLS12-381 signing suite for validator block/attestation signatures). `go list -m -versions` shows 12 releases since, up to v1.37.0.

This isn't urgent, but worth tracking deliberately rather than drifting further behind, because:
- It's a consensus-critical crypto dependency — any bump needs an audit of call-site behavior across `crypto/signing/mcl`, `crypto/signing/mcl/singlesig`, and `crypto/signing/mcl/multisig` for signature-determinism, plus a coordinated network rollout (all validators need matching behavior), not a routine dependency-update PR.
- Some incidental fixes have landed upstream in the vendored precompiled archives between v1.28.2 and v1.31.0 (unrelated to crypto logic — build/packaging level), which is a nice-to-have but not itself a reason to rush a bump.

Proposing this as a discussion/scoping issue: does the team want to schedule an audit + staged bump at some point, or intentionally stay pinned until there's a specific need? Happy to help scope the call-site audit if useful.

*(Note: a related low-severity build-hardening finding tied to the same vendored archive was reported privately per SECURITY.md — see that report for detail if relevant to this decision.)*

PR #95, opened 2026-07-21
fix: harden GNU_STACK on Linux builds (validator/operator/keygenerator)

Test0rMaik:
## Problem

`make build-validator`, `build-operator`, and `build-keygenerator` link a vendored precompiled static archive from `github.com/herumi/bls-go-binary v1.28.2` (`bls/lib/linux/amd64/libbls384_256.a`) whose hand-assembled object members lack an ELF `.note.GNU-stack` section. When GNU ld can't find that marker on every linked object, it conservatively marks the whole binary's `GNU_STACK` program header executable instead of read/write.

Confirmed via `readelf -lW`: `bin/validator`, `bin/operator`, and `bin/keygenerator` currently ship with `GNU_STACK ... RWE`. `bin/seednode` and `bin/benchmark` (no herumi link) are correctly `RW`.

Reported privately first via GHSA-3rg6-pgrq-qc3x per SECURITY.md, since it's a finding in currently-shipped release binaries. @fbsobreira independently verified the repro and additionally found `static_code.o` in the same archive missing the marker too (same fix covers it, since the flag is linker-level, not per-object), and confirmed the vendored `linux/arm64` archive already carries the note on all members, so arm64 builds are unaffected — this fix is specifically an amd64 gap. Agreed as Low severity (defense-in-depth gap, not independently exploitable) and cleared for a public PR rather than a private fork.

## Fix

Add `-Wl,-z,noexecstack` to the shared `GOBUILD` linker flags, gated to Linux only via the existing `UNAME_S` variable — `ld64` on Darwin has no ELF/`GNU_STACK` concept and doesn't accept this flag. Verified `make UNAME_S=Darwin -n build-validator` produces byte-identical flags to before this change, so the macOS legs of `release.yaml` are unaffected.

Docker builds need no separate change — they all invoke this same root `Makefile` target inside a Linux container, so they inherit the fix automatically.

## Testing

```
make clean && make build
for b in validator seednode operator keygenerator benchmark; do
  readelf -lW bin/$b | grep GNU_STACK
done
```
All 5 binaries now show `RW`; previously `validator`/`operator`/`keygenerator` showed `RWE`. Also re-verified with `RELEASE=1` (matches what CI/Docker actually ship) — stripping doesn't affect the `GNU_STACK` program header.

Also ran this change through this project's usual pre-submission audit pipeline (independent correctness + security review of the diff): confirmed correct Make variable ordering/expansion semantics, correct quoting/escaping under the added token, no other place in the repo reconstructs `-extldflags` independently, `-z noexecstack` is authoritative over ld's per-object inference (not a heuristic that could silently stop applying), and no code path in `kvm/wasmer2` or `crypto/signing/mcl` relies on an executable stack (wasmer2's JIT lives in a separately dlopen'd `.so` with its own program headers, unrelated to this flag).

## Out of scope

`herumi/bls-go-binary` ships the fix at the source as of v1.31.0 (object renamed to `bint64.o`, has a proper `.note.GNU-stack`), but bumping the dependency is intentionally not part of this PR — it's a consensus-critical BLS12-381 signing library spanning several minor releases since our v1.28.2 pin, and would need its own audit for signature-determinism/network-coordination before touching it. Tracked separately in #93.

## Summary

- Hardened **Linux amd64** builds for **validator**, **operator**, and **keygenerator** (and `make newkey` via shared linker flag plumbing) by adding `-Wl,-z,noexecstack` to the shared Go linker `extldflags` for these binaries.
- Kept the existing platform behavior by gating the change on the **target** `GOOS` (Darwin excluded) and preserving rpath handling (`$ORIGIN` / `@executable_path`).
- Prevents affected binaries from inheriting an **executable stack** (`GNU_STACK` → `RWE`) caused by missing `.note.GNU-stack` sections in objects from the **vendored Herumi BLS** archive; **seednode and benchmark binaries remain unchanged**.
- No changes to consensus, transaction processing, state management, KVM, networking, concurrency, error handling, node runtime behavior, or data integrity. **Vendored BLS dependency was not upgraded**.
- Linux Docker builds inherit the same fix through the root Makefile; linker flags are protected from being overridden via `override`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,8 +92,19 @@ LOG=*:INFO
 endif
 
 GOCMD=go
-GORUN=$(ENV_FLAG) $(GOCMD) run -ldflags="$(ldflags)"
-GOBUILD=$(GOCMD) build $(GO_BUILD_FLAGS) -ldflags="$(ldflags) -extldflags '-Wl,-rpath,\$$ORIGIN,-rpath,@executable_path'"
+
+# Some vendored cgo static archives lack .note.GNU-stack markers, which makes
+# GNU ld mark the binaries' stack executable. Force noexecstack on Linux;
+# Darwin's ld64 has no GNU_STACK concept and doesn't accept the flag. Gated on
+# the Go target OS (not UNAME_S, the build host) so a cross-compiled
+# GOOS=linux build from a non-Linux host still gets it.
+override EXTLDFLAGS := -Wl,-rpath,\$$ORIGIN,-rpath,@executable_path
+ifeq ($(shell $(GOCMD) env GOOS),linux)
+override EXTLDFLAGS += -Wl,-z,noexecstack
+endif
+
+GORUN=$(ENV_FLAG) $(GOCMD) run -ldflags="$(ldflags) -extldflags '$(EXTLDFLAGS)'"
+GOBUILD=$(GOCMD) build $(GO_BUILD_FLAGS) -ldflags="$(ldflags) -extldflags '$(EXTLDFLAGS)'"
 
 .DEFAULT_GOAL := help
 
--- a/Makefile
+++ b/Makefile
@@ -92,14 +92,21 @@ LOG=*:INFO
 endif
 
 GOCMD=go
+override GOOS_TARGET := $(shell $(GOCMD) env GOOS)
 
 # Some vendored cgo static archives lack .note.GNU-stack markers, which makes
 # GNU ld mark the binaries' stack executable. Force noexecstack on Linux;
-# Darwin's ld64 has no GNU_STACK concept and doesn't accept the flag. Gated on
-# the Go target OS (not UNAME_S, the build host) so a cross-compiled
-# GOOS=linux build from a non-Linux host still gets it.
-override EXTLDFLAGS := -Wl,-rpath,\$$ORIGIN,-rpath,@executable_path
-ifeq ($(shell $(GOCMD) env GOOS),linux)
+# Darwin's ld64 has no GNU_STACK concept and doesn't accept the flag. rpath
+# tokens are platform-specific too ($ORIGIN is GNU ld/ELF, @executable_path is
+# ld64/Mach-O), so pick exactly one instead of passing both on every build.
+# Gated on GOOS_TARGET (the Go target, not UNAME_S the build host) so a
+# cross-compiled GOOS=linux build from a non-Linux host still gets it.
+ifeq ($(GOOS_TARGET),darwin)
+override EXTLDFLAGS := -Wl,-rpath,@executable_path
+else
+override EXTLDFLAGS := -Wl,-rpath,\$$ORIGIN
+endif
+ifeq ($(GOOS_TARGET),linux)
 override EXTLDFLAGS += -Wl,-z,noexecstack
 endif
 


Makefile
```diff
@@ -93,7 +93,19 @@ endif
...
 GOCMD=go
 GORUN=$(ENV_FLAG) $(GOCMD) run -ldflags="$(ldflags)"
-GOBUILD=$(GOCMD) build $(GO_BUILD_FLAGS) -ldflags="$(ldflags) -extldflags '-Wl,-rpath,\$$ORIGIN,-rpath,@executable_path'"
+
+# Some vendored cgo static archives (notably herumi/bls-go-binary v1.28.2's
+# libbls384_256.a) ship object members without a .note.GNU-stack section,
+# which makes GNU ld mark the resulting binary's GNU_STACK segment
+# executable instead of read/write. -z noexecstack forces the correct flag
+# regardless of the vendored object. ld64 on Darwin has no ELF/GNU_STACK
+# concept and does not accept this flag, so it's Linux-only.
```
fbsobreira: nit: this can be trimmed — the ld inference mechanics are more detail than a Makefile reader needs, and the v1.28.2 pin goes stale once #93 bumps the dep. Something like:

```suggestion
# The vendored herumi BLS amd64 archive lacks .note.GNU-stack markers, which
# makes GNU ld mark the binaries' stack executable. Force noexecstack on Linux;
# Darwin's ld64 has no GNU_STACK concept and doesn't accept the flag.
```

Makefile
```diff
@@ -93,7 +93,19 @@ endif
...
+EXTLDFLAGS := -Wl,-rpath,\$$ORIGIN,-rpath,@executable_path
+ifeq ($(UNAME_S),Linux)
```
nickgs1337: Small thing on the commit message: the linux/arm64 archives already had the .note.GNU-stack section, so only amd64 builds were shipping RWE. Right now it reads like every Linux artifact was affected. Worth adding the amd64 qualifier.

nickgs1337: This keys off `uname -s`, which is the build host and not the target. So `GOOS=linux make build-validator` from a Mac quietly produces an unhardened binary, no warning, nothing in the log. CI never cross-compiles today so it's latent, but a security flag that fails open silently makes me nervous. Can we gate on `$(shell $(GOCMD) env GOOS)` instead? GOCMD is already defined a few lines up, so it's a drop-in.

Makefile
```diff
@@ -93,7 +93,19 @@ endif
...
+EXTLDFLAGS += -Wl,-z,noexecstack
```
nickgs1337: Nothing asserts this in CI, so if the flag ever gets dropped, reordered, or ignored (external linking going away, a herumi bump, etc) the binary still builds fine and nobody notices. A `readelf -lW bin/validator | grep -q 'GNU_STACK.*RW '` step after the build in release.yaml/push.yaml would cover it. The trailing space in `RW ` matters, that's what keeps RWE from matching. Not blocking. While you're in the linker flags, `-Wl,-z,relro -Wl,-z,now` would close the partial-RELRO gap too, but that's a separate PR.

Makefile
```diff
@@ -93,7 +93,19 @@ endif
...
 GORUN=$(ENV_FLAG) $(GOCMD) run -ldflags="$(ldflags)"
-GOBUILD=$(GOCMD) build $(GO_BUILD_FLAGS) -ldflags="$(ldflags) -extldflags '-Wl,-rpath,\$$ORIGIN,-rpath,@executable_path'"
+
+# Some vendored cgo static archives (notably herumi/bls-go-binary v1.28.2's
+# libbls384_256.a) ship object members without a .note.GNU-stack section,
+# which makes GNU ld mark the resulting binary's GNU_STACK segment
+# executable instead of read/write. -z noexecstack forces the correct flag
+# regardless of the vendored object. ld64 on Darwin has no ELF/GNU_STACK
+# concept and does not accept this flag, so it's Linux-only.
+EXTLDFLAGS := -Wl,-rpath,\$$ORIGIN,-rpath,@executable_path
```
nickgs1337: nit: pulling this out of the literal means `make build-validator EXTLDFLAGS=whatever` on the command line now silently wipes both the rpath and the hardening flag. Nobody's setting that by accident, but `override EXTLDFLAGS :=` fixes it for one word and makes the intent obvious.

Makefile
nickgs1337: Did anyone actually run a linux/amd64 build with this, or was it just inspected? `-z noexecstack` changes runtime behavior and not only the header, so if any of the linked asm genuinely needed an exec stack you'd get a SIGSEGV inside a BLS call, not a link error. I doubt it does (the two offenders are hand-written bignum asm), but readelf can't tell us. A sign+verify round trip or a validator producing a block would settle it, plus a line in the description.

Makefile
```diff
@@ -93,7 +93,19 @@ endif
 
 GOCMD=go
 GORUN=$(ENV_FLAG) $(GOCMD) run -ldflags="$(ldflags)"
```
nickgs1337: `make newkey` goes through GORUN, which carries no `-extldflags` at all, so the temp binary go run builds still has the exec stack. `build-keygenerator` is covered fine. Since the target's help text sells newkey as the way to generate validator keys, either extend GORUN the same way or narrow the commit subject to the `build-*` targets.

nickgs1337:
Couple of housekeeping things before this merges. The commit still carries a Co-Authored-By trailer, which we don't do in this repo, so it needs an amend or a squash-merge with a cleaned message. It's the only commit here so it lands on develop as-is otherwise.

Also for whoever picks this up next: `git diff develop...HEAD` shows ~133 files and 9k lines, but that's just develop being behind the branch point. The actual change is Makefile only, +13/-1.

Test0rMaik:
Thanks both for the thorough review — pushed an amended commit (`017b20e`) addressing everything:

- **Co-Authored-By trailer**: removed, amended the single commit.
- **fbsobreira's comment nit**: trimmed to the shorter version suggested, no more v1.28.2 pin reference that'd go stale after #93.
- **Commit message scope**: corrected — this is an amd64-specific issue (linux/arm64 archive already had the marker), not "every Linux artifact."
- **`UNAME_S` vs target OS (real bug, thank you for catching it)**: switched the gate to `$(shell $(GOCMD) env GOOS)`. Confirmed with a `GOOS=darwin make -n build-validator` dry-run from this (Linux) host that it now correctly follows the *target*, not the build host.
- **`override` on `EXTLDFLAGS`**: added — and this actually caught a real bug of mine in testing: I'd initially added `override` only on the `:=` line, not the `+=` line inside the `ifeq`. Turns out GNU Make silently drops a plain `+=` on a variable that's already `override`-flagged — no warning, `make -n` output looks fine, the appended flag is just gone. Confirmed by bisecting a minimal reproduction against the real Makefile. Fixed by repeating `override` on both lines; verified the flag is present *and* that `make EXTLDFLAGS=hijacked build-validator` still can't clobber it.
- **Runtime verification, not just readelf**: fair challenge. Built `crypto/signing/mcl/singlesig` and `crypto/signing/mcl/multisig` test binaries with the exact same `-extldflags`, confirmed via `readelf -lW` that the *test binaries themselves* carry `GNU_STACK RW` (so the flag was actually applied, not just assumed), then ran them: full sign/verify suite passes (`TestBLSSigner_SignOK`, `VerifyOK`, `SignVerifyWithReconstructedPubKeyOK`) plus multisig signature aggregation (`TestMultiSignerBLS_AggregateSignaturesOK`). No crash — the hand-written bignum asm doesn't need an executable stack.
- **`GORUN`/`newkey` gap**: extended `GORUN` to carry the same `-extldflags`, so `make newkey` is covered too, not just the `build-*` targets.
- **CI assertion / RELRO**: agreed both are good ideas, leaving as fast-follows rather than expanding this PR's diff into `release.yaml`/`push.yaml` — happy to pick either up separately if useful.
- **Title format**: don't have a KLC ticket for this one (came in via the private security report rather than JIRA) — happy to rename if someone assigns a key, or if you'd rather rename at merge time that works too.

Re-verified full build afterward: all 5 binaries rebuilt clean, `readelf -lW` shows `GNU_STACK RW` on validator/operator/keygenerator (previously RWE), `RW` unchanged on seednode/benchmark.

Makefile
nickgs1337: Subject and body both call out amd64 now (arm64 already had the note), reads right.

nickgs1337: Gate's on go env GOOS now, so a GOOS=linux cross build from a Mac still gets hardened. That closes the fail-open, thanks for switching it.

Makefile
nickgs1337: Yeah, both are fine as fast-follows (they were non-blockers anyway). Happy to pick up the readelf CI check separately.

Makefile
nickgs1337: override's on both the := and += lines now, so a command-line EXTLDFLAGS can't wipe it. Nice catch on the += getting silently dropped too.

Makefile
nickgs1337: You actually ran the BLS sign/verify + aggregation with the flag applied (confirmed via readelf on the test binary itself), so the asm clearly does not need an exec stack. That settles it.

Makefile
nickgs1337: GORUN carries the same -extldflags now, so newkey's covered too and not just the build-* targets. Sorted.

Makefile
```diff
@@ -92,8 +92,19 @@ LOG=*:INFO
...
+
+# Some vendored cgo static archives lack .note.GNU-stack markers, which makes
+# GNU ld mark the binaries' stack executable. Force noexecstack on Linux;
+# Darwin's ld64 has no GNU_STACK concept and doesn't accept the flag. Gated on
+# the Go target OS (not UNAME_S, the build host) so a cross-compiled
+# GOOS=linux build from a non-Linux host still gets it.
+override EXTLDFLAGS := -Wl,-rpath,\$$ORIGIN,-rpath,@executable_path
+ifeq ($(shell $(GOCMD) env GOOS),linux)
+override EXTLDFLAGS += -Wl,-z,noexecstack
+endif
```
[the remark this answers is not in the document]
nickgs1337: This dual rpath isn't introduced here, it's been on GOBUILD in develop all along. This PR only hoists that string into a variable and adds noexecstack on Linux, and the darwin flags come out byte identical to before. Splitting the rpath by OS is a fair cleanup but it changes the macOS build output, so I'd rather keep it out of a Linux stack hardening change and do it as a separate PR.

nickgs1337 (dismissed):

nickgs1337:
@Test0rMaik all good, we are ready to merge your PR. But, per security definition, we can only merge verified signatures. Could you please sign all your commits so we can merge this?
image

Test0rMaik:
@nickgs1337 Done — all commits on this branch are now signed and showing as Verified. Ready to merge whenever you are.

nickgs1337 (approved):

Beroni (approved):

PR #96, opened 2026-07-21
fix: int32 overflow in managed-buffer bounds checks (fork-gated)

Test0rMaik:
## Summary

Three managed-buffer VM host functions computed their bounds check as `int(startingPosition+length) > len(buf)`, with the addition happening in **int32** before the cast to `int`. Since `startingPosition`/`sliceLength`/`dataLength` are contract-controlled `int32` arguments, a crafted call (e.g. `startingPosition` and `sliceLength` both near `math.MaxInt32`) overflows the addition. Both operands are already checked non-negative first, and two non-negative int32s can only overflow to a *negative* wrapped value, so this never produced a wrong-but-in-bounds slice — it always fell through to Go's own runtime slice-bounds panic on the subsequent slice expression, which is recovered upstream in `RunSmartContractCall`/`RunSmartContractCreate` as `ErrExecutionPanicked` (an irrecoverable execution failure). So this was never exploitable as memory corruption, but it relied on that panic/recover as an implicit safety net instead of failing gracefully, unlike the equivalent overflow-safe check in `kvm/executor/executorMemory.go`.

Affected functions in `kvm/vmhost/vmhooks/manBufOps.go`:
- `MBufferGetByteSlice`
- `ManagedBufferCopyByteSliceWithHost` (`MBufferCopyByteSlice`)
- `ManagedBufferSetByteSliceWithTypedArgs` (`MBufferSetByteSlice`)

## Why fork-gated

Changing this from "irrecoverable tx failure" to "graceful out-of-bounds error, execution may continue" is contract-visible, consensus-relevant behavior — a validator on the old binary and one on the new binary would compute different outcomes for the identical crafted-overflow input. So the fix is gated behind `FixAuditChangesV3`, matching the existing pattern used for this class of change elsewhere (e.g. `BigIntPow` in `bigIntOps.go`): pre-fork the code path is byte-identical to today (including the panic/recover), post-fork it uses `math.AddInt32WithErr` for an explicit, graceful check. This keeps historical block replay unaffected and leaves the actual activation epoch to the team, same as other pending V3 items.

## Testing

- `go build ./...` and full `go test ./kvm/vmhost/...` pass.
- Added 3 regression tests in `kvm/vmhost/hosttest/managedops_test.go` covering the overflow input for all three functions, asserting the graceful post-fork outcome (`result == 1`, no panic).
- Manually verified (then reverted) that forcing the fork inactive via `WithEnableEpochs` reproduces the original panic on the same input, confirming the tests exercise a real behavior difference and aren't tautological.

## Test plan
- [ ] CI green
- [ ] Team confirms `FixAuditChangesV3` is the right bucket for this (vs. a dedicated flag) and assigns/keeps the activation epoch

## Summary

- Fixed `int32` overflow in KVM managed-buffer bounds checks.
- Updated `MBufferGetByteSlice`, `ManagedBufferCopyByteSliceWithHost`, `ManagedBufferSetByteSliceWithTypedArgs`, and `ManagedTypesContext.SetByteSlice`.
- After `FixAuditChangesV3`, overflow returns a graceful bounds error instead of causing a slice-bounds panic.
- Pre-fork behavior remains unchanged for compatibility.
- Added fork-aware regression tests for all affected paths.
- The fix improves KVM execution stability and reduces denial-of-service risk from crafted parameters. It does not directly change consensus, transaction processing, state management, or networking.


kvm/vmhost/vmhooks/manBufOps.go
```diff
@@ -130,12 +130,27 @@ func (context *VMHooksImpl) MBufferGetByteSlice(
...
 
-	if startingPosition < 0 || sliceLength < 0 || int(startingPosition+sliceLength) > len(sourceBytes) {
-		// does not fail execution if slice exceeds bounds
-		return 1
+	// Pre-fork, startingPosition+sliceLength was added as int32 before the bounds
+	// check, so a crafted overflow slipped past it and panicked on the slice
+	// expression below; that panic was recovered upstream as an irrecoverable
+	// execution failure. Changing that outcome for historical blocks would break
+	// replay, so the fixed, non-panicking bounds check is fork-gated.
+	var slice []byte
```
nickgs1337: Any reason we do not just fold the gate into the existing condition instead of branching? Adding `(addErr != nil && context.host.ForkController().FixAuditChangesV3())` to the current `if` and leaving the slice expression alone keeps pre-fork behaviour identical (the sum still wraps negative, still fails `int(sum) > len`, still panics), and it drops ~42 lines across the three sites. That is also the shape bigIntOps.go:657 already uses for V3 gating. At the third site keep the legacy `int(start)+int(length)` widening so the replay argument stays trivial to audit.

nickgs1337: nit: the same 14 line block is now copy pasted three times, and two of the copies carry a comment admitting they have to stay in lockstep with the first. Next time someone tweaks the bounds semantics they will touch two of three sites and that divergence is consensus relevant. Could be one unexported helper, something like `sliceBounds(host, start, length, bufLen) (int32, int32, bool)`. Less pressing if you take the fold suggestion, since each site shrinks to a few lines anyway.

kvm/vmhost/vmhooks/manBufOps.go
```diff
@@ -130,12 +130,27 @@ func (context *VMHooksImpl) MBufferGetByteSlice(
...
+	if context.host.ForkController().FixAuditChangesV3() {
+		endPosition, addErr := math.AddInt32WithErr(startingPosition, sliceLength)
+		if startingPosition < 0 || sliceLength < 0 || addErr != nil || int(endPosition) > len(sourceBytes) {
+			// does not fail execution if slice exceeds bounds
+			return 1
+		}
+		slice = sourceBytes[startingPosition:endPosition]
+	} else {
+		if startingPosition < 0 || sliceLength < 0 || int(startingPosition+sliceLength) > len(sourceBytes) {
```
nickgs1337: Not an issue with the patch, more a question for whoever owns the fork schedule. Until V3 turns on, this branch still panics, and every hit writes an ERROR level stack dump and then drops the warm instance via CleanInstance(), so repeated calls are noisy and force recompiles. Nothing dangerous, it is recovered deterministically into ErrExecutionPanicked and the caller pays full gas. Can someone confirm the mainnet fixAuditChangesV3 epoch is near term? In-repo default is 0, but a distant override would keep that window open for a while.

kvm/vmhost/hosttest/managedops_test.go
```diff
@@ -197,3 +197,106 @@ func TestManaged_StorageStoreKeyIsolatedFromBufferMutation(t *testing.T) {
 
 	require.NoError(t, err)
 }
+
+// overflowStartingPosition + a small positive length overflows int32, which
+// used to slip past the bounds check and panic on the slice expression itself.
+const overflowStartingPosition = int32(2147483642) // math.MaxInt32 - 5
```
nickgs1337: Any reason not to write `math.MaxInt32 - 5` here? Then the decoder comment can go, and the value explains itself.

kvm/vmhost/hosttest/managedops_test.go
```diff
@@ -197,3 +197,106 @@ func TestManaged_StorageStoreKeyIsolatedFromBufferMutation(t *testing.T) {
...
+
+func TestManaged_GetByteSlice_OverflowRejectedGracefully(t *testing.T) {
```
nickgs1337: Nothing here pins the legacy branch, which is the whole reason the gate exists. TestHostBuilder leaves FixAuditChangesV3 out of its EnableEpochs literal, so it lands on 0 and the flag is already on, meaning all three tests only cover the post-fork path and the else branches have zero coverage. Can we add a counterpart per hook with `.WithEnableEpochs(...)` pushing FixAuditChangesV3 far out and assert ErrExecutionPanicked? preForkEnableEpochs() in readonly_delete_test.go already does this. Without it a future cleanup deletes those else branches, tests stay green, and replaying a pre-activation block diverges.

nickgs1337: nit: that is a lot of wasm harness for asserting three `int32(1)` returns. The vmhooks package already has the lighter fork-aware builder these audit fixes use (`makeHostParametersWithFork(newForkStub(...))`, bigFloatOps_test.go:42), and sibling V3 fixes drive both fork states table-driven through it in about a third of the lines. Happy either way though, adding the pre-fork rows here works too.

kvm/vmhost/hosttest/managedops_test.go
```diff
@@ -197,3 +197,106 @@ func TestManaged_StorageStoreKeyIsolatedFromBufferMutation(t *testing.T) {
...
+				WithBalance(1000).
+				WithMethods(func(parentInstance *mock.InstanceMock, config interface{}) {
+					parentInstance.AddMockMethod("testFunction", func() *mock.InstanceMock {
+						host := parentInstance.Host
+						managedType := host.ManagedTypes()
+						hooks := vmhooks.NewVMHooksImpl(host)
+						mBuffer := managedType.NewManagedBufferFromBytes([]byte(data))
+
+						result := hooks.MBufferGetByteSlice(mBuffer, overflowStartingPosition, 10, 0)
+						require.Equal(t, int32(1), result)
```
nickgs1337: nit: this runs on the goroutine RunSmartContractCall spawns, not the test goroutine, and FailNow is not allowed from there. It still fails on a regression, just as a confusing "Provided VMOutput is nil" instead of the actual mismatch. Cheap fix is to Finish the return code and assert it in AndAssertResults. Same at 252 and 285.

nickgs1337:
The comment above fixAuditChangesV3 in config/node/enableEpochs.yaml (line 31) still describes the flag as just the kdaFeesPool ratio rejection, the swap price range check and the upgrade hook dispatch block. Can you append the managed buffer bounds change there? That comment is the only readable inventory of what flipping the flag does on a live chain, and operators use it to size the activation risk, so it drifting understates what is riding on it. The rest of the drift on the branch can be a follow-up.

Test0rMaik:
Addressed in bf50cbc:

- Folded the fork gate into each function's existing bounds condition (`... || (addErr != nil && ForkController().FixAuditChangesV3())`) instead of if/else duplication — same shape as `BigIntPow` in bigIntOps.go:657. Pre-fork path is now byte-identical to the original by construction. Site 3 keeps the legacy `int(start)+int(length)` widening untouched, as requested.
- Replaced the three tests (which asserted with `require` inside the goroutine `RunSmartContractCall` spawns — invalid `FailNow` usage from a non-test goroutine) with a single table-driven test in `kvm/vmhost/vmhooks/manBufOps_test.go`, using the lighter `makeHostParametersWithFork(newForkStub(...))` harness from bigFloatOps_test.go, calling the hooks directly with no execution goroutine involved. It now also pins the pre-fork legacy path (`require.Panics`), which had zero coverage before.
- Appended a line to the `fixAuditChangesV3` comment in `config/node/enableEpochs.yaml` documenting this change, per your request — left the rest of that comment's drift as a follow-up as you suggested.

On the mainnet epoch timing question: that's for the team/fork-schedule owner to answer, not something I can speak to from the code.

Re-ran the correctness + security audit pipeline on this revision (including `-race`) after the changes — no issues found. One unrelated pre-existing thing surfaced while running `-race` across the full `kvm/vmhost/...` tree: a `checkptr` panic in the wasmer2 cgo bridge (`getVMHooksFromContextRawPtr`), reproducible on the unmodified base commit too, unrelated to this PR — flagging in case it's not already a known issue.

Test0rMaik:
@coderabbitai review

kvm/vmhost/vmhooks/manBufOps.go
Test0rMaik: Done in 5bbaaa7 — folded into each site's existing bounds condition (`... || (addErr != nil && ForkController().FixAuditChangesV3())`), same shape as `bigIntOps.go:657`. Pre-fork path is now byte-identical to the original by construction, and site 3 keeps the legacy `int(start)+int(length)` widening untouched, as you called out.

Test0rMaik: Left the three sites separately duplicated rather than extracting a helper, per your own framing of this as less pressing once the fold-into-existing-condition change landed — each site shrank enough that a shared `sliceBounds` helper felt like it'd cost more indirection than it saves for three call sites. Happy to extract it if you'd still rather have it.

kvm/vmhost/vmhooks/manBufOps.go
Test0rMaik: That's for the team/fork-schedule owner to answer, not something I can speak to from the code — flagging for them rather than guessing at the mainnet epoch timing.

kvm/vmhost/hosttest/managedops_test.go
Test0rMaik: This test file's additions from the original patch were removed entirely — replaced with a single table-driven test in `kvm/vmhost/vmhooks/manBufOps_test.go` (commit 5bbaaa7) that calls the hooks directly instead of going through this decoder path, so this specific `math.MaxInt32` nit no longer applies to the new code.

kvm/vmhost/hosttest/managedops_test.go
Test0rMaik: Same as above — this file's test additions were replaced by the new table-driven test in `manBufOps_test.go`, which now does cover the pre-fork legacy branch: each hook has a `fixAuditV3: false` case asserting `require.Panics` on the same crafted-overflow input, driven through `newForkStub(fixAuditV3)` so a future cleanup can't silently drop it without a red test.

Test0rMaik: Took the suggestion — the new table-driven test in `manBufOps_test.go` drives both fork states through `makeHostParametersWithFork(newForkStub(...))` per hook, no wasm execution harness involved, about a third of the lines of the old approach.

kvm/vmhost/hosttest/managedops_test.go
Test0rMaik: Fixed by construction, not patched — the new tests in `manBufOps_test.go` call the hooks directly (no `RunSmartContractCall` goroutine involved), so there's no wrong-goroutine `FailNow`/`require` risk in the replacement tests.

Test0rMaik:
Rebased onto latest \`develop\`, and it surfaced a real, currently-crashable regression along the way (commit dae6ebe):

An unrelated upstream commit refactored \`ManagedBufferSetByteSliceWithTypedArgs\` (\`kvm/vmhost/vmhooks/manBufOps.go\`) to delegate its bounds-check-and-copy logic to a new \`ManagedTypesContext.SetByteSlice\` (\`kvm/vmhost/contexts/managedType.go\`) — but that new method still computes \`startingPosition+dataLength\` as \`int32\` before the bounds check, reintroducing the exact overflow this PR fixes at the other two call sites. This branch's own existing regression test caught it immediately as a live panic on the \`FixAuditChangesV3=true\` path during the rebase.

Ported the same fork-gated \`math.AddInt32WithErr\` pattern to \`SetByteSlice\`, added \`TestManagedTypesContext_SetByteSlice_OverflowIsForkGated\` (mutation-tested against the original naive check), and ran it through the same correctness+security audit pipeline as the rest of this PR — both came back clean.

fbsobreira:
Closing this — the fix landed independently, and it's now released and disclosed.

## What shipped

`729f8317` ([KLC-2579][KLR-50]) consolidated every call site behind one helper,
`vmhost.SliceIsOutOfBounds` in `kvm/vmhost/helpers.go`, using the same
`math.AddInt32WithErr` pattern this PR proposed:

```go
func SliceIsOutOfBounds(fixAuditChangesV5 bool, startingPosition int32, sliceLength int32, bufferLength int) bool {
      if startingPosition < 0 || sliceLength < 0 {
              return true
      }

      if !fixAuditChangesV5 {
              return int(startingPosition+sliceLength) > bufferLength
      }

      endPosition, err := math.AddInt32WithErr(startingPosition, sliceLength)

      return err != nil || int(endPosition) > bufferLength
}
```

It covers `MBufferGetByteSlice`, `ManagedBufferCopyByteSliceWithHost`,
`ManagedBufferSetByteSliceWithTypedArgs` and `ManagedTypesContext.SetByteSlice` — including the
refactor reintroduction you caught on the August rebase.

Two differences from this branch:

- **Fork flag.** Gated on `FixAuditChangesV5`, not `FixAuditChangesV3`. The audit-fix buckets are
  strictly ordered `V3 < V4 < V5`, and V5 was the bucket still open when this landed.
- **Shared helper.** The check lives in one place instead of being repeated per call site, which is
  what let the fourth site get picked up along with the other three.

Released in **1.7.21-rc3**. `FixAuditChangesV5` activated on mainnet on 2026-09-18.

## Advisories

Published today, one per host function:

- [GHSA-vxf6-vgpv-822h](https://github.com/klever-io/klever-go/security/advisories/GHSA-vxf6-vgpv-822h) — `MBufferGetByteSlice`
- [GHSA-2m22-3m34-2qxf](https://github.com/klever-io/klever-go/security/advisories/GHSA-2m22-3m34-2qxf) — `MBufferCopyByteSlice`
- [GHSA-wjrx-39g9-j97q](https://github.com/klever-io/klever-go/security/advisories/GHSA-wjrx-39g9-j97q) — `MBufferSetByteSlice`

You're credited as a finder on all three — this PR is the earliest dated identification of all
three sites on our record.

## One ask for next time

Per [`SECURITY.md`](https://github.com/klever-io/klever-go/blob/develop/SECURITY.md#reporting-a-vulnerability),
please don't send vulnerability findings as public pull requests — use **Security → Report a
vulnerability** instead. This one sat in a public branch with the overflow and its trigger values
spelled out for two months before we had an advisory ready. It worked out here, but it removes our
ability to choose the disclosure timing. Credit is identical either way; the private channel just
lets the fix land first.

Thanks for the find, and for the rebase catch in particular — the `SetByteSlice` delegation would
have slipped through otherwise.

[KLC-2579]: https://klever-io.atlassian.net/browse/KLC-2579?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #97, opened 2026-07-23
fix: bound postWSConnection mirror concurrency with a worker pool

Test0rMaik:
## Summary

- `SocketHub.asyncPost` previously spawned an unbounded goroutine per account/tx event to mirror it via HTTP POST to the operator-configured `postConnectionURL`. Under high chain throughput (many txs/account changes per block) with a slow or unresponsive mirror endpoint, this let goroutines and outbound sockets grow without bound — a self-inflicted resource-exhaustion risk scaling with chain activity rather than with the mirror's actual capacity.
- Replaces it with a fixed pool of `postWorkerCount` (8) workers draining a bounded `postQueue` (1000 slots), started once from `StartServer(ctx)` and stopped on `ctx` cancellation.
- `asyncPost` now does a non-blocking send and drops (with a rate-limited warning, mirroring the existing `indexer.trySendEvent` drop-log pattern) when the queue is full, instead of spawning more goroutines.
- The queue is never allocated at all when no `postConnectionURL`/`postConnectionAPIKey` is configured, so the feature stays a true no-op when the mirror is disabled (previously a goroutine was still spawned per event just to check and return).
- `postConnectionURL`/`postConnectionAPIKey` remain operator-only config (`network/api/api.go`); no websocket client input can reach them, and the events that feed the mirror are unchanged — only the concurrency/backpressure handling after the fact changed.

Reviewed with a two-agent correctness + security audit pass against the diff; no findings from either.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./websocket/...`
- [x] `gofmt -l` clean on changed files
- [x] `go test ./websocket/... ./network/api/websocket/... -race` — all pass, including new tests covering: no-op when unconfigured, queue allocated only when a URL/API key is set, delivery through a real `httptest` server, drop-on-full behavior, and worker shutdown on context cancellation

## Summary

- **Networking (websocket mirror delivery):** Adds a bounded “post mirror” pipeline for `postWSConnection` in `websocket`:
  - Replaces per-event goroutine spawning with a fixed **8-worker** pool draining a bounded **1,000-event** queue.
  - `asyncPost` becomes a **non-blocking enqueue** (or **no-op** when mirror config isn’t present) and **drops** mirror events when the queue is full, with **rate-limited** warning logs to highlight backpressure.
  - Mirror HTTP posting is made **context-cancellable** via `PostURLWithContext`, and workers **shut down cleanly** on server context cancellation (waitgroup coordination).
- **Blockchain impact:** **No changes** to consensus, transaction processing, state management, or KVM. The change affects only outbound mirror/observability traffic; **mirrored payloads are unchanged**, with the only new functional behavior being **event drops under sustained backpressure**.
- **Reliability / stability (cross-cutting concurrency & error handling):**
  - Prevents **unbounded goroutine growth** from per-event posting by enforcing bounded concurrency and queueing.
  - Improves HTTP POST robustness with stage-specific error wrapping and ensures **failed error-body reads** are surfaced (wrapped) instead of being ignored.
- **Testing:** Extends coverage for mirror disabled behavior (no queue allocated; `asyncPost` no-op; no workers started), queue allocation gating, successful delivery via `httptest`, deterministic queue-full dropping, worker stop-on-cancellation, and improved error handling via a regression test for truncated error responses.

### DIFF
--- a/cmd/operator/utils/http.go
+++ b/cmd/operator/utils/http.go
@@ -60,7 +60,15 @@ func PostURLWithContext(ctx context.Context, url, body string, headers []string,
 	if err != nil {
 		return fmt.Errorf("send POST request: %w", err)
 	}
-	defer func() { _ = r.Body.Close() }()
+	// Draining before Close lets the transport reuse the underlying HTTP/1.x connection;
+	// otherwise it's forced to redial (and re-handshake, for TLS) on the next request. This
+	// matters most when target == nil (the caller doesn't want the body at all, e.g. the
+	// websocket mirror at websocket.go — every 2xx response with a non-empty body would
+	// otherwise pay for a fresh connection).
+	defer func() {
+		_, _ = io.Copy(io.Discard, io.LimitReader(r.Body, maxResponseBody))
+		_ = r.Body.Close()
+	}()
 
 	if err := checkStatus(http.MethodPost, url, r); err != nil {
 		return err
@@ -88,11 +96,20 @@ func checkStatus(method, url string, r *http.Response) error {
 	}
 
 	status := sanitizeServerText(r.Status)
+	// io.ReadAll returns the bytes it did manage to read alongside a non-nil error on a
+	// truncated/hijacked body (e.g. a declared Content-Length the server never delivered) —
+	// keep that partial snippet instead of discarding it. readErr's own message can embed
+	// server-supplied text (net/http's trailer parsing may fold a server header into it), so
+	// sanitize it the same as status/body rather than let it skip the same treatment.
 	snippet, readErr := io.ReadAll(io.LimitReader(r.Body, maxErrorBodySnippet))
-	if readErr != nil {
-		return fmt.Errorf("%s %s: unexpected HTTP status %s (failed to read response body: %w)", method, url, status, readErr)
-	}
 	body := strings.TrimSpace(sanitizeServerText(string(snippet)))
+	if readErr != nil {
+		sanitizedReadErr := sanitizeServerText(readErr.Error())
+		if body == "" {
+			return fmt.Errorf("%s %s: unexpected HTTP status %s (failed to read response body: %s)", method, url, status, sanitizedReadErr)
+		}
+		return fmt.Errorf("%s %s: unexpected HTTP status %s: %s (failed to read response body: %s)", method, url, status, body, sanitizedReadErr)
+	}
 	if body == "" {
 		return fmt.Errorf("%s %s: unexpected HTTP status %s", method, url, status)
 	}
--- a/cmd/operator/utils/http.go
+++ b/cmd/operator/utils/http.go
@@ -49,7 +49,7 @@ func PostURLWithContext(ctx context.Context, url, body string, headers []string,
 	reqBody := strings.NewReader(body)
 	req, errNewReq := http.NewRequestWithContext(ctx, http.MethodPost, url, reqBody)
 	if errNewReq != nil {
-		return errNewReq
+		return fmt.Errorf("create POST request: %w", errNewReq)
 	}
 	req.Header.Add("Content-type", "application/json; charset=UTF-8")
 	for i := 0; i < len(headers); i += 2 {
@@ -58,7 +58,7 @@ func PostURLWithContext(ctx context.Context, url, body string, headers []string,
 
 	r, err := httpClient.Do(req)
 	if err != nil {
-		return err
+		return fmt.Errorf("send POST request: %w", err)
 	}
 	defer func() { _ = r.Body.Close() }()
 
@@ -69,7 +69,7 @@ func PostURLWithContext(ctx context.Context, url, body string, headers []string,
 	if target != nil {
 		data, errRead := io.ReadAll(io.LimitReader(r.Body, maxResponseBody+1))
 		if errRead != nil {
-			return errRead
+			return fmt.Errorf("read POST response body: %w", errRead)
 		}
 		if int64(len(data)) > maxResponseBody {
 			return fmt.Errorf("%s %s: response body exceeds %d bytes", http.MethodPost, url, maxResponseBody)
--- a/cmd/operator/utils/http.go
+++ b/cmd/operator/utils/http.go
@@ -87,8 +87,11 @@ func checkStatus(method, url string, r *http.Response) error {
 		return nil
 	}
 
-	snippet, _ := io.ReadAll(io.LimitReader(r.Body, maxErrorBodySnippet))
 	status := sanitizeServerText(r.Status)
+	snippet, readErr := io.ReadAll(io.LimitReader(r.Body, maxErrorBodySnippet))
+	if readErr != nil {
+		return fmt.Errorf("%s %s: unexpected HTTP status %s (failed to read response body: %w)", method, url, status, readErr)
+	}
 	body := strings.TrimSpace(sanitizeServerText(string(snippet)))
 	if body == "" {
 		return fmt.Errorf("%s %s: unexpected HTTP status %s", method, url, status)
--- a/cmd/operator/utils/http.go
+++ b/cmd/operator/utils/http.go
@@ -1,6 +1,7 @@
 package utils
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -38,8 +39,15 @@ func GetURL(url string, target interface{}) error {
 
 // PostURL provides a post using a json string
 func PostURL(url, body string, headers []string, target interface{}) error {
+	return PostURLWithContext(context.Background(), url, body, headers, target)
+}
+
+// PostURLWithContext behaves like PostURL but ties the request to ctx, so a caller can
+// abort an in-flight POST (e.g. on shutdown) instead of waiting out httpClient's fixed
+// timeout.
+func PostURLWithContext(ctx context.Context, url, body string, headers []string, target interface{}) error {
 	reqBody := strings.NewReader(body)
-	req, errNewReq := http.NewRequest(http.MethodPost, url, reqBody)
+	req, errNewReq := http.NewRequestWithContext(ctx, http.MethodPost, url, reqBody)
 	if errNewReq != nil {
 		return errNewReq
 	}
--- a/websocket/config.go
+++ b/websocket/config.go
@@ -26,6 +26,18 @@ const (
 	// string can never match a real address and would only feed a per-connection
 	// memory-amplification path (GHSA-4fwh-wrm6-97xm).
 	maxEncodedAddressLength = 62
+
+	// postWorkerCount bounds how many postWSConnection mirror requests run concurrently.
+	// Without a cap, a slow or unresponsive mirror endpoint turns every account/tx event
+	// into an unbounded pile of in-flight goroutines and sockets, scaling with chain
+	// throughput rather than with the mirror's actual capacity.
+	postWorkerCount = 8
+
+	// postQueueSize bounds how many pending mirror sends can queue behind postWorkerCount
+	// before new ones are dropped (mirroring the EventQueue drop-on-full pattern in
+	// indexer/events.go), so a stalled mirror endpoint sheds load instead of growing without
+	// bound.
+	postQueueSize = 1000
 )
 
 // pingPeriod and pongWait are the /subscribe keepalive timings. They are vars only so
--- a/websocket/config.go
+++ b/websocket/config.go
@@ -27,17 +27,18 @@ const (
 	// memory-amplification path (GHSA-4fwh-wrm6-97xm).
 	maxEncodedAddressLength = 62
 
-	// postWorkerCount bounds how many postWSConnection mirror requests run concurrently.
-	// Without a cap, a slow or unresponsive mirror endpoint turns every account/tx event
-	// into an unbounded pile of in-flight goroutines and sockets, scaling with chain
-	// throughput rather than with the mirror's actual capacity.
-	postWorkerCount = 8
+	// defaultPostWorkerCount bounds how many postWSConnection mirror requests run
+	// concurrently, when PostWorkers is unset. Without a cap, a slow or unresponsive
+	// mirror endpoint turns every account/tx event into an unbounded pile of in-flight
+	// goroutines and sockets, scaling with chain throughput rather than with the mirror's
+	// actual capacity.
+	defaultPostWorkerCount = 8
 
-	// postQueueSize bounds how many pending mirror sends can queue behind postWorkerCount
-	// before new ones are dropped (mirroring the EventQueue drop-on-full pattern in
-	// indexer/events.go), so a stalled mirror endpoint sheds load instead of growing without
-	// bound.
-	postQueueSize = 1000
+	// defaultPostQueueSize bounds how many pending mirror sends can queue behind
+	// postWorkerCount before new ones are dropped (mirroring the EventQueue drop-on-full
+	// pattern in indexer/events.go), when PostQueueSize is unset — so a stalled mirror
+	// endpoint sheds load instead of growing without bound.
+	defaultPostQueueSize = 1000
 )
 
 // pingPeriod and pongWait are the /subscribe keepalive timings. They are vars only so
@@ -55,6 +56,12 @@ var (
 type Limits struct {
 	MaxAddressesPerSubscribe int
 	MaxAddressesPerClient    int
+	// PostWorkers bounds concurrent postWSConnection mirror requests; 0 uses
+	// defaultPostWorkerCount.
+	PostWorkers int
+	// PostQueueSize bounds pending mirror sends queued behind PostWorkers; 0 uses
+	// defaultPostQueueSize.
+	PostQueueSize int
 }
 
 // resolvedLimits holds the effective limits after defaults are applied.
@@ -62,6 +69,8 @@ type resolvedLimits struct {
 	maxAddressesPerSubscribe int
 	maxAddressesPerClient    int
 	maxMessageSize           int64
+	postWorkers              int
+	postQueueSize            int
 }
 
 func (l Limits) resolve() resolvedLimits {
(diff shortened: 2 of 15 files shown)


Test0rMaik:
@coderabbitai review

Test0rMaik:
@coderabbitai review

websocket/websocket.go
```diff
@@ -25,8 +27,32 @@ const (
...
+
+var (
+	droppedPostCount   int64
+	lastPostDropLogged int64
 )
 
+// logPostQueueDrop rate-limits the "mirror queue full" warning so a sustained stall
+// against a slow postConnectionURL logs a periodic summary instead of one line per
+// dropped event.
+func logPostQueueDrop() {
```
nickgs1337: nit: this counter/CAS/swap dance is the same one inlined in `indexer.trySendEvent` (events.go:36-49), 10s window and all. I wouldn't export that one though, its rate-limit state is package-global and that's the opposite of where we want to go here. If we dedupe, probably a small value type holding a `tools/atomic.Counter` plus the window timestamp, with each owner keeping its own instance.

websocket/websocket_test.go
```diff
@@ -691,6 +693,141 @@ func TestPostWSConnection_WithServer(t *testing.T) {
 	}
 }
 
+func TestAsyncPost_NoopWithoutConfig(t *testing.T) {
```
nickgs1337, line 710: nit: the goroutine + `close(done)` + 2s timeout harness is copy-pasted a bunch of times in this file now, might be worth a small `assertReturnsQuickly(t, fn)` helper. Also this one can't fail the way the comment says: a send on a nil channel falls straight into the `default:` arm, so it'll never block. Asserting the drop counter didn't move would give it some teeth.

cmd/operator/utils/http.go
```diff
@@ -38,10 +39,17 @@ func GetURL(url string, target interface{}) error {
 
 // PostURL provides a post using a json string
 func PostURL(url, body string, headers []string, target interface{}) error {
```
nickgs1337, line 41: nit: `PostURLWithContext` has exactly one caller and `GetURL` didn't get the same twin, so we could just add ctx to `PostURL` and pass `context.Background()` at the 4 CLI sites. Fine to leave as-is too, it's the same shape as `http.NewRequest`/`NewRequestWithContext`. Your call.

websocket/websocket.go
```diff
@@ -221,7 +302,7 @@ func (h *SocketHub) handleBroadcastEvent(event indexer.Event, subscription map[*
 	h.broadcastToSubscription(parsed, subscription)
 }
 
-func (h *SocketHub) postWSConnection(message *Send) error {
+func (h *SocketHub) postWSConnection(ctx context.Context, message *Send) error {
 	if h.postConnectionURL == "" && h.postConnectionAPIKey == "" {
```
nickgs1337: "Is the mirror on?" is asked in three places now with two different spellings, and this one is unreachable from the production path since `asyncPost` can't enqueue on a nil queue. Careful just deleting it though: `NewHub` uses `url != "" || key != ""`, so setting only the API key allocates the queue and starts 8 workers, and this `&&` lets every send through to an empty URL that fails forever. Requiring a non-empty URL in `NewHub` makes the nil queue the single source of truth and makes this guard provably dead, so it can go in the same change.

cmd/operator/utils/http.go
```diff
@@ -50,7 +58,7 @@ func PostURL(url, body string, headers []string, target interface{}) error {
 
 	r, err := httpClient.Do(req)
 	if err != nil {
-		return err
+		return fmt.Errorf("send POST request: %w", err)
 	}
 	defer func() { _ = r.Body.Close() }()
```
nickgs1337: When `target == nil` we close the body without ever reading it, and that's exactly how the mirror calls this (websocket.go:315 passes nil). Go's transport won't reuse an HTTP/1.x connection unless the body is read to completion and closed, so any mirror returning a non-empty 2xx body makes every event POST pay for a fresh dial plus TLS handshake. An `io.Copy(io.Discard, io.LimitReader(r.Body, maxResponseBody))` before the early return covers it. Also, `httpClient` rides `http.DefaultTransport`, so `MaxIdleConnsPerHost` is 2 while we run 8 workers against one host, meaning 6 of 8 churn even after draining. That client is shared with the operator CLI so cloning the transport is optional, just call it out if you do.

websocket/websocket_test.go
```diff
@@ -691,6 +693,141 @@ func TestPostWSConnection_WithServer(t *testing.T) {
...
+	}
+}
+
+func TestNewHub_PostQueueAllocatedOnlyWhenConfigured(t *testing.T) {
+	assert.Nil(t, NewHub("", "", nil).postQueue, "no URL or API key: mirror disabled")
+	assert.NotNil(t, NewHub("http://example.com", "", nil).postQueue, "URL set: mirror enabled")
+	assert.NotNil(t, NewHub("", "api-key", nil).postQueue, "API key set: mirror enabled")
+}
+
+func TestAsyncPost_DeliversToWorker(t *testing.T) {
```
nickgs1337, line 796: The invariant this PR exists for, at most `postWorkerCount` mirror requests in flight, isn't tested anywhere. `postWorkerCount` doesn't show up in a single `_test.go` file, and every test here would stay green if we reverted to `go func(){ h.postWSConnection(...) }()`, so the suite blesses the shape of the fix without proving it. Same gap on ctx propagation: `StopOnContextCancel` points at `http://example.invalid`, which fails DNS instantly, so the workers exit because they were idle and not because a live request got cut short. A blocking httptest handler covers both cheaply: queue `postWorkerCount*4` sends, assert in-flight peaks at `postWorkerCount`, then cancel while a request is parked in the handler and assert `postWorkersWG.Wait()` returns well inside the 10s client timeout. Can we get that in before merge?

websocket/websocket.go
```diff
@@ -25,8 +27,32 @@ const (
 	errMissingNonceHash = "must provide nonce or hash"
 	errTxNotFound       = "transaction not found"
 	errBlockNotFound    = "block not found"
+
+	postQueueDropLogIntervalSeconds = 10
+)
+
+var (
```
nickgs1337: These describe per-hub state but live at package level, while every other mirror field (`postQueue`, `postConnectionURL`, `postWorkersWG`) is already on `SocketHub`. Cross-hub interference is latent in prod since there's one hub, but the test pollution is live: `TestAsyncPost_DropsWhenQueueFull` leaves `lastPostDropLogged` set to now for whatever runs next in the package, which is the kind of order-dependent flake that shows up on a loaded CI box months later. Moving both onto the struct and making `logPostQueueDrop` a method is mechanical. Separate half: the tail count never gets flushed, so a burst that drops 899 events and stops sits in the counter until some unrelated later drop reports it against the wrong window. Flushing on worker exit would fix that.

websocket/config.go
```diff
@@ -26,6 +26,18 @@ const (
 	// string can never match a real address and would only feed a per-connection
 	// memory-amplification path (GHSA-4fwh-wrm6-97xm).
 	maxEncodedAddressLength = 62
+
+	// postWorkerCount bounds how many postWSConnection mirror requests run concurrently.
+	// Without a cap, a slow or unresponsive mirror endpoint turns every account/tx event
+	// into an unbounded pile of in-flight goroutines and sockets, scaling with chain
+	// throughput rather than with the mirror's actual capacity.
+	postWorkerCount = 8
```
nickgs1337: These two are hardcoded while every sibling /subscribe limit is operator-tunable through the facade (`WSMaxConnections` and friends). The bigger question isn't the plumbing though: this flips mirror delivery from unbounded-but-lossless to bounded-and-lossy, throughput capped at roughly `postWorkerCount / latency`, with no retry or replay. A 100ms mirror plus a few hundred account sends per block drops during perfectly normal operation and the mirror's chain view just diverges, with a 10s WARN as the only signal. Can we get a call from whoever owns the mirror consumer on whether 8/1000 fit our real endpoints, and whether drops should emit a metric rather than only that WARN? If we want it tunable, extending the existing `Limits` struct with `PostWorkers`/`PostQueueSize` keeps today's values as the zero-value fallback.

websocket/websocket.go
```diff
@@ -74,12 +113,52 @@ func (h *SocketHub) MaxAddressesPerSubscribe() int {
...
+			for {
+				select {
+				case <-ctx.Done():
+					return
+				case parsed := <-h.postQueue:
+					// select has no case priority: ctx.Done() and postQueue can both be
+					// ready at once, so a cancelled shutdown can still dequeue one more
+					// item. Check ctx.Err() before posting instead of starting (and
+					// immediately failing) a doomed request.
+					if ctx.Err() != nil {
```
nickgs1337, line 206: nit: this drops the message we just dequeued plus everything left in the queue with no record, while `StartServer` logs "close gracefully" right before blocking on the WG. Reads like we drained the backlog when it really means 8 goroutines agreed to abandon it. One WARN here carrying `len(h.postQueue)` would at least make a gap in mirrored events after a restart explainable.

websocket/websocket.go
```diff
@@ -74,12 +113,52 @@ func (h *SocketHub) MaxAddressesPerSubscribe() int {
...
+						return
+					}
+					if err := h.postWSConnection(ctx, parsed); err != nil {
+						log.Warn("ws.EventReceive.postWSConnection", "failed to post", err.Error())
```
nickgs1337: We rate-limit the drop warning two lines up but this one fires per failure, and a dead mirror is by far the louder path (drops only happen once the queue is already full). A refused connection fails instantly, so 8 workers chew through the 1000-entry backlog and emit 1000 WARNs in a few milliseconds, then one per event forever after. Folding it into the same helper, one representative error plus a count per window, would be consistent. It's also reachable by misconfig alone: key set with an empty URL starts 8 workers that can never succeed, each failing inside `httpClient.Do` with "no Host in request URL" wrapped as "send POST request: ...". Rejecting that combo in `NewHub` is the better fix and lines up with the other thread on the `postWSConnection` guard.

cmd/operator/utils/http.go
```diff
@@ -79,8 +87,11 @@ func checkStatus(method, url string, r *http.Response) error {
 		return nil
 	}
 
-	snippet, _ := io.ReadAll(io.LimitReader(r.Body, maxErrorBodySnippet))
 	status := sanitizeServerText(r.Status)
+	snippet, readErr := io.ReadAll(io.LimitReader(r.Body, maxErrorBodySnippet))
+	if readErr != nil {
+		return fmt.Errorf("%s %s: unexpected HTTP status %s (failed to read response body: %w)", method, url, status, readErr)
```
nickgs1337: Two things in these three lines. `io.ReadAll` hands back the bytes it did manage to read alongside the error, so this early return throws away a partial body that the old `snippet, _ :=` kept. `TestErrorBodyReadFailureIsWrapped` actually demonstrates it: the server writes "short" before hijacking, and that never reaches the message. Second, `readErr` is the one server-influenced string here that skips `sanitizeServerText` while `status` and `body` both go through it, and net/http's trailer parsing can embed a server-supplied key in that error. Keeping the sanitized snippet and appending the read failure satisfies the earlier round without the regression, and since `GetURL` shares `checkStatus` the test should widen to cover it too.

websocket/websocket.go
Test0rMaik: Added a small per-hub `dropWarner` type (count + window timestamp) instead — used for both this queue-full path and the post-failure path below, so the two can't share or reset each other's window. Left `indexer.trySendEvent`'s copy alone per your note (package-global state there is intentionally a different shape). Pushed in bdf4026.

websocket/websocket_test.go
Test0rMaik: Replaced the four copy-pasted goroutine+timeout blocks in this file with a shared `assertReturnsQuickly(t, timeout, msg, fn)` helper. And you're right that this one couldn't actually fail the way its old comment claimed — added an assertion that the drop-warn counter didn't move, so a regression that deleted `asyncPost`'s `postQueue == nil` early-return would be caught (it'd fall through to the counted path instead of the early return, even though the select itself still wouldn't block). Pushed in bdf4026.

cmd/operator/utils/http.go
Test0rMaik: Left as-is per your "fine to leave as-is too" — didn't want to touch `PostURL`'s shape in a round that's already changing a lot around it.

websocket/websocket.go
Test0rMaik: Requiring a URL was the right fix — NewHub now only allocates postQueue when postConnectionURL != "" (API key alone logs a warning and leaves the mirror disabled instead of starting doomed workers), which makes this guard provably dead, so it's removed. Pushed in bdf4026.

cmd/operator/utils/http.go
Test0rMaik: Added a drain (io.Copy(io.Discard, io.LimitReader(r.Body, maxResponseBody))) before Close in PostURLWithContext, so the connection can be reused on the target == nil path the mirror actually takes. Pushed in bdf4026.

websocket/websocket_test.go
Test0rMaik: Added TestStartPostWorkers_BoundedInFlightAndCtxCancelAbortsRequest — a blocking httptest handler that parks postWorkerCount*4 sends, asserts peak concurrency never exceeds postWorkerCount, then cancels while requests are actually parked in the handler and asserts postWorkersWG.Wait() returns well inside httpClient's 10s timeout. Mutation-tested by reverting startPostWorkers to one-goroutine-per-item first — failed with "32 is not <= 8", confirming it actually catches the regression. Pushed in bdf4026.

websocket/websocket.go
Test0rMaik: Moved droppedPostCount/lastPostDropLogged onto the SocketHub struct (via the new dropWarner type, one instance for queue-full drops and a separate one for post failures) and made the log function a method. Also flush any non-zero tail count once after postWorkersWG.Wait() in StartServer's shutdown path, so a burst that never reaches another occurrence to trigger the next window doesn't sit unreported. Pushed in bdf4026.

websocket/config.go
Test0rMaik: Went with making it tunable — Limits gained PostWorkers/PostQueueSize (0 = today's 8/1000 defaults), threaded through a new antiflood config section, nodeFacade, and MainAPIHandler the same way the existing address-cap limits already are. Also wired the drop/failure counts into core.AppStatusHandler (two new metrics: klv_ws_mirror_queue_dropped_total, klv_ws_mirror_post_failures_total) alongside the existing rate-limited WARN logs, so there's a proper cumulative signal beyond the periodic log line. Both were explicit calls from the repo owner rather than my own default recommendation (I'd have flagged both as follow-ups otherwise) — happy to adjust the metric names/shape if they don't match whatever convention the rest of the mirror-consumer tooling expects. Pushed in bdf4026.

websocket/websocket.go
Test0rMaik: Added a WARN here with len(h.postQueue)+1 (the just-dequeued item plus whatever's still queued) when shutdown abandons it. Pushed in bdf4026.

websocket/websocket.go
Test0rMaik: Folded this into the same dropWarner (a separate postFailWarn instance), so a dead mirror now logs one representative error plus a count per window instead of one line per failure. Also closed the misconfig case directly: NewHub now requires postConnectionURL to be set to enable the mirror at all, so an API-key-only config no longer starts workers that can never succeed in the first place (see the other thread on the postWSConnection guard). Pushed in bdf4026.

cmd/operator/utils/http.go
Test0rMaik: Fixed: kept the partial snippet (io.ReadAll still returns what it did read alongside the error) and sanitized readErr.Error() the same way status/body already are, since it can carry server-controlled text via net/http's trailer parsing. Widened TestErrorBodyReadFailureIsWrapped into GetURL/PostURL subtests, both now asserting the partial "short" body survives into the message. Pushed in bdf4026.

websocket/websocket.go
nickgs1337: dropWarner is what I had in mind, and giving the two paths their own instance is better than what I suggested. Leaving indexer's copy alone is right too.

websocket/websocket_test.go
nickgs1337: helper kills the copy-paste, and the drop counter assertion gives that test actual teeth. good.

cmd/operator/utils/http.go
nickgs1337: yep, fine by me. it was your call and the twin matches net/http's own shape anyway.

websocket/websocket.go
nickgs1337: keying the mirror off the URL alone makes the nil queue the single source of truth, so dropping the guard is safe now. warning on api-key-only instead of starting 8 doomed workers is the right behavior.

cmd/operator/utils/http.go
nickgs1337: drain is in the right place (deferred, before Close), so the nil-target mirror posts keep their connection. the MaxIdleConnsPerHost=2 vs 8 workers thing is still there, but I flagged that half as optional so not holding this up.

websocket/websocket_test.go
nickgs1337: this is exactly what I was after. the blocking handler proves the bound, cancellation happens with requests genuinely parked, and mutation testing it against the one-goroutine-per-item revert is the part that convinces me.

websocket/websocket.go
nickgs1337: state is on the hub now so the cross-test pollution is gone, and the flush after postWorkersWG.Wait() covers the tail count. both halves handled.

websocket/config.go
nickgs1337: tunable via Limits with today's values as the zero fallback is the shape I had in mind, and the two counters give us the cumulative signal the WARN never could. if the owner made the call on 8/1000 that is good enough for me.

websocket/websocket.go
nickgs1337: pending count in the shutdown warn is all I wanted. a gap in mirrored events after a restart is explainable now.

websocket/websocket.go
nickgs1337: separate postFailWarn instance stops the log flood on a dead mirror, and requiring the URL in NewHub closes the misconfig that made it reachable. both covered.

cmd/operator/utils/http.go
nickgs1337: partial snippet survives now and readErr goes through sanitizeServerText like status and body do, plus the test covers both callers. all good.

nickgs1337 (approved):

PR #99, opened 2026-07-23
fix: eliminate data race on websocket keepalive timing vars

Test0rMaik:
## Summary

- `pingPeriod`/`pongWait` (`websocket/config.go`) were plain package-level vars, documented as "vars only so tests can shorten them ... nothing in production mutates them." In practice, `TestClient_IdleConnectionReclaimedAtPongWait` (`websocket/hardening_test.go`) does mutate them via `setKeepaliveForTest`, concurrently with a live client's `loopIn`/`loopOut` goroutines (`websocket/client.go`) still reading the previous value on their way out — the test only waits for `Done()` (ctx cancellation observed by a separate goroutine), not for `loopIn`/`loopOut` to have actually returned, so the test's deferred restore could race with `loopOut`'s ticker-branch read.
- Reproduced directly: `go test ./websocket/... -race -run TestClient_IdleConnectionReclaimedAtPongWait -count=1` failed with a genuine `DATA RACE` report roughly 1 in 5-10 runs (not a timing-assertion flake — an actual unsynchronized concurrent read/write caught by the race detector).
- Fixed at the root: `pingPeriod`/`pongWait` are now `atomic.Int64`-backed (`pingPeriodNs`/`pongWaitNs`), read via `getPingPeriod()`/`getPongWait()` accessors at every call site in `client.go`, and mutated via `Swap`/`Store` in the test helper (using `Swap` to atomically capture-and-replace in one step, avoiding a `Load`+`Store` race on the restore path).
- No behavior change for production code — same values, same semantics, just synchronized access.

Before starting this fix, I searched klever-io/klever-go's open and closed PRs/issues (`gh search prs`/`gh search issues` scoped to this repo, plus a full title grep) to confirm nothing already reports or fixes this race — nothing did.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./websocket/...`
- [x] `gofmt -l` clean on changed files
- [x] `go test ./websocket/... -race -run TestClient_IdleConnectionReclaimedAtPongWait -count=30` — 30 consecutive clean passes (previously ~1-in-5-10 failure rate with a DATA RACE report)
- [x] `go test ./websocket/... -race -count=1` (full package, several runs) — all clean

## Summary

- **Blockchain-critical components affected:** None. Changes are confined to the **websocket** package’s **/subscribe** keepalive/timeout behavior used by networking tests and WebSocket clients; there are no modifications to consensus, transaction processing, state management, or KVM.
- **Node stability / data integrity:** No intended production behavior change. The fix primarily targets **test/clients concurrency safety** by eliminating a timing-related **data race** when test helpers update and restore keepalive parameters while websocket goroutines are running.
- **Concurrency (race prevention):** The package-level **ping/pong** timing configuration is now stored as an **atomic, immutable pair** (via an atomic pointer) and read through **`getPingPeriod()` / `getPongWait()`** accessors. Test helpers update/restore the paired timings **atomically**, preventing torn/partially-updated values from being observed concurrently.
- **Networking timeout handling hardening:** The websocket client’s inbound/outbound logic now derives **read deadlines** and **ping scheduling/write/control deadlines** from the atomic accessors rather than fixed globals. Failures to refresh read deadlines are treated as fatal to the inbound loop (log + terminate loop), improving deterministic timeout behavior.
- **Error handling improvements (tests):** The idle-connection reclamation test now surfaces websocket **upgrade errors** through a buffered channel so the test fails deterministically rather than relying solely on timing-based behavior.

### DIFF
--- a/websocket/client.go
+++ b/websocket/client.go
@@ -84,7 +84,10 @@ func (c *client) loopIn() {
 	// handler and each read. loopOut's pings keep a live client warm while a dead/idle one
 	// is reclaimed at pongWait (GHSA-4fwh-wrm6-97xm).
 	c.conn.SetReadLimit(c.hub.limits.maxMessageSize)
-	_ = c.conn.SetReadDeadline(time.Now().Add(getPongWait()))
+	if err := c.conn.SetReadDeadline(time.Now().Add(getPongWait())); err != nil {
+		log.Warn("ws.loopIn", "err", err.Error())
+		return
+	}
 	c.conn.SetPongHandler(func(string) error {
 		return c.conn.SetReadDeadline(time.Now().Add(getPongWait()))
 	})
@@ -97,7 +100,10 @@ func (c *client) loopIn() {
 			}
 			break
 		}
-		_ = c.conn.SetReadDeadline(time.Now().Add(getPongWait()))
+		if err := c.conn.SetReadDeadline(time.Now().Add(getPongWait())); err != nil {
+			log.Warn("ws.loopIn", "err", err.Error())
+			break
+		}
 
 		// ReadMessage only yields data frames; gorilla answers client ping and close
 		// control frames through its default handlers (close surfaces as a read error
--- a/websocket/client.go
+++ b/websocket/client.go
@@ -84,9 +84,9 @@ func (c *client) loopIn() {
 	// handler and each read. loopOut's pings keep a live client warm while a dead/idle one
 	// is reclaimed at pongWait (GHSA-4fwh-wrm6-97xm).
 	c.conn.SetReadLimit(c.hub.limits.maxMessageSize)
-	_ = c.conn.SetReadDeadline(time.Now().Add(pongWait))
+	_ = c.conn.SetReadDeadline(time.Now().Add(getPongWait()))
 	c.conn.SetPongHandler(func(string) error {
-		return c.conn.SetReadDeadline(time.Now().Add(pongWait))
+		return c.conn.SetReadDeadline(time.Now().Add(getPongWait()))
 	})
 
 	for {
@@ -97,7 +97,7 @@ func (c *client) loopIn() {
 			}
 			break
 		}
-		_ = c.conn.SetReadDeadline(time.Now().Add(pongWait))
+		_ = c.conn.SetReadDeadline(time.Now().Add(getPongWait()))
 
 		// ReadMessage only yields data frames; gorilla answers client ping and close
 		// control frames through its default handlers (close surfaces as a read error
@@ -135,7 +135,7 @@ func (c *client) loopOut() {
 	// Ping the client periodically; its pong refreshes loopIn's read deadline, keeping
 	// passive-but-live subscribers up while dead ones time out. WriteControl is safe to
 	// call concurrently with the other writer.
-	ticker := time.NewTicker(pingPeriod)
+	ticker := time.NewTicker(getPingPeriod())
 	defer ticker.Stop()
 
 	for {
@@ -148,7 +148,7 @@ func (c *client) loopOut() {
 			}
 			// Bound the write: a client that re-arms its read deadline but never drains its
 			// socket would otherwise park this goroutine here indefinitely (GHSA-4fwh-wrm6-97xm).
-			if err := c.conn.SetWriteDeadline(time.Now().Add(pingPeriod)); err != nil {
+			if err := c.conn.SetWriteDeadline(time.Now().Add(getPingPeriod())); err != nil {
 				log.Warn("ws.loopOut", "err", err.Error())
 				c.close()
 				return
@@ -159,7 +159,7 @@ func (c *client) loopOut() {
 				return
 			}
 		case <-ticker.C:
-			if err := c.conn.WriteControl(ws.PingMessage, nil, time.Now().Add(pingPeriod)); err != nil {
+			if err := c.conn.WriteControl(ws.PingMessage, nil, time.Now().Add(getPingPeriod())); err != nil {
 				log.Warn("ws.loopOut.ping", "err", err.Error())
 				c.close()
 				return
--- a/websocket/client.go
+++ b/websocket/client.go
@@ -84,12 +84,12 @@ func (c *client) loopIn() {
 	// handler and each read. loopOut's pings keep a live client warm while a dead/idle one
 	// is reclaimed at pongWait (GHSA-4fwh-wrm6-97xm).
 	c.conn.SetReadLimit(c.hub.limits.maxMessageSize)
-	if err := c.conn.SetReadDeadline(time.Now().Add(getPongWait())); err != nil {
+	if err := c.conn.SetReadDeadline(time.Now().Add(c.hub.limits.pongWait)); err != nil {
 		log.Warn("ws.loopIn", "err", err.Error())
 		return
 	}
 	c.conn.SetPongHandler(func(string) error {
-		return c.conn.SetReadDeadline(time.Now().Add(getPongWait()))
+		return c.conn.SetReadDeadline(time.Now().Add(c.hub.limits.pongWait))
 	})
 
 	for {
@@ -100,8 +100,10 @@ func (c *client) loopIn() {
 			}
 			break
 		}
-		if err := c.conn.SetReadDeadline(time.Now().Add(getPongWait())); err != nil {
-			log.Warn("ws.loopIn", "err", err.Error())
+		if err := c.conn.SetReadDeadline(time.Now().Add(c.hub.limits.pongWait)); err != nil {
+			// Debug, not Warn: this only fires when a concurrent c.close() lands between a
+			// successful read and this call — normal teardown, not an anomaly.
+			log.Debug("ws.loopIn", "err", err.Error())
 			break
 		}
 
@@ -141,7 +143,7 @@ func (c *client) loopOut() {
 	// Ping the client periodically; its pong refreshes loopIn's read deadline, keeping
 	// passive-but-live subscribers up while dead ones time out. WriteControl is safe to
 	// call concurrently with the other writer.
-	ticker := time.NewTicker(getPingPeriod())
+	ticker := time.NewTicker(c.hub.limits.pingPeriod)
 	defer ticker.Stop()
 
 	for {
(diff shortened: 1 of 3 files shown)


Test0rMaik:
@coderabbitai review

websocket/config.go
```diff
@@ -28,14 +31,30 @@ const (
...
+	pong time.Duration
+}
+
+// keepalive holds the current *keepaliveTimings. Only tests mutate it, to shorten the
+// timings and exercise idle-client read-deadline reclamation quickly — but they mutate
+// concurrently with a live client's loopIn/loopOut goroutines still reading the previous
+// value on their way out, so plain vars here are a genuine data race, not just a
+// theoretical one (caught by `go test -race`: a test's deferred restore racing loopOut's
+// read at client.go's ticker branch).
+var keepalive atomic.Pointer[keepaliveTimings]
```
nickgs1337: All of this (the type, the atomic pointer, the init, both getters, 6 call-site rewrites) exists so one test can shorten the timings, and the race only shows up because that test's deferred restore writes while goroutines are still reading. Any reason we don't just put `pingPeriod`/`pongWait` on `resolvedLimits` and set them in `resolve()`? The hub is built once per test before any client starts, so there's no concurrent write to synchronize, and the client already reads `c.hub.limits.maxMessageSize` right there. That kills the race by construction and the timings go back to being consts.

websocket/config.go
```diff
@@ -28,14 +31,30 @@ const (
...
-var (
-	pingPeriod = 15 * time.Second
-	pongWait   = 30 * time.Second
-)
+// keepaliveTimings holds the /subscribe keepalive timing pair. pongWait (the lifetime
+// read deadline) must exceed pingPeriod so a live client can answer a server ping before
+// it elapses — the two are stored and swapped together as one immutable value (never
+// mutated in place) so a reader can never observe a torn combination of an old ping with
+// a new pong (or vice versa) that could violate that invariant.
+type keepaliveTimings struct {
```
nickgs1337: The comment says a reader can never see an old ping paired with a new pong, but no caller ever loads both fields from one snapshot. `getPingPeriod()` and `getPongWait()` each do their own `Load()`, and the two values are consumed by different goroutines anyway (loopIn only reads pong, loopOut only ping), so a `Swap` landing between those reads produces exactly the mixed pair the comment rules out. The data race itself is genuinely fixed, but the struct isn't buying anything over two `atomic.Int64`s. Moving the pair onto `resolvedLimits` so each connection snapshots it once would make the claim true, otherwise the comment should say what the code actually does.

websocket/config.go
```diff
@@ -28,14 +31,30 @@ const (
...
 
-// pingPeriod and pongWait are the /subscribe keepalive timings. They are vars only so
-// tests can shorten them to exercise the idle-client read-deadline reclamation quickly;
-// nothing in production mutates them. pongWait (the lifetime read deadline) must exceed
-// pingPeriod so a live client can answer a server ping before it elapses.
-var (
-	pingPeriod = 15 * time.Second
-	pongWait   = 30 * time.Second
-)
+// keepaliveTimings holds the /subscribe keepalive timing pair. pongWait (the lifetime
```
nickgs1337: 23 lines of comment for a two-field struct is a lot, and the tearing argument gets restated almost word for word at hardening_test.go:206-212. It also pins production docs to a specific test and to one branch in client.go, which will rot the first time either of those moves. If the timings move onto the hub this deletes itself, otherwise one line about swapping the pair together covers it.

websocket/config.go
```diff
@@ -28,14 +31,30 @@ const (
...
+
+// keepalive holds the current *keepaliveTimings. Only tests mutate it, to shorten the
+// timings and exercise idle-client read-deadline reclamation quickly — but they mutate
+// concurrently with a live client's loopIn/loopOut goroutines still reading the previous
+// value on their way out, so plain vars here are a genuine data race, not just a
+// theoretical one (caught by `go test -race`: a test's deferred restore racing loopOut's
+// read at client.go's ticker branch).
+var keepalive atomic.Pointer[keepaliveTimings]
+
+func init() {
```
nickgs1337: Go runs every package-level var initializer before any `init()`, so a future global in this package whose initializer reaches `getPingPeriod()` would nil-deref the `Load()` before main. Nothing does that today (only `log` and the const block), so it's purely defensive. Storing the default in the var initializer instead of `init()` removes the ordering dependency, and it's moot if the timings move to the hub anyway.

websocket/client.go
```diff
@@ -97,7 +100,10 @@ func (c *client) loopIn() {
 			}
 			break
 		}
-		_ = c.conn.SetReadDeadline(time.Now().Add(pongWait))
+		if err := c.conn.SetReadDeadline(time.Now().Add(getPongWait())); err != nil {
```
nickgs1337: nit: the read error five lines up deliberately filters ordinary disconnects through `IsUnexpectedCloseError`, but this one warns unconditionally. It only fires when a concurrent `c.close()` lands between a successful read and this call, which is normal teardown rather than an anomaly, so `Debug` feels like the right level. Handling the error at all is a real improvement over the old `_ =`.

websocket/hardening_test.go
```diff
@@ -222,9 +228,14 @@ func TestClient_IdleConnectionReclaimedAtPongWait(t *testing.T) {
 	upgrader := ws.Upgrader{CheckOrigin: func(*http.Request) bool { return true }}
 
 	released := make(chan struct{})
+	// Buffered so the handler goroutine (not the test goroutine — calling t.Fatal there
+	// would violate testing's FailNow-must-run-on-the-test's-own-goroutine contract)
+	// never blocks reporting an upgrade failure back to the test.
+	upgradeErrCh := make(chan error, 1)
```
nickgs1337: nit: `t.Errorf` is safe from any goroutine, it's only `FailNow`/`Fatal` that have to run on the test goroutine. So this could collapse to `t.Errorf("upgrade: %v", err)` plus a return in the handler, dropping the channel, the select case and the comment. The 2s timeout still fails the test and you still get the real cause printed.

nickgs1337:
Other shape worth weighing, over in `NewClient` (client.go:24): freeze the two timings into the `client` struct there and read them as plain fields in loopIn/loopOut. No global, no atomics, no init, since they're written once before the goroutines start, and the pong > ping invariant then actually holds for the whole life of the connection. I'd still lean toward the `resolvedLimits` version though, the timings are config and that way the test injects through `NewHub(..., Limits{...})` instead of poking package state. Either way, pick one, not both.

nickgs1337:
The Summary at the top of the description still says the timings are `atomic.Int64`-backed via `pingPeriodNs`/`pongWaitNs`, but neither identifier exists and the code uses `atomic.Pointer[keepaliveTimings]`. Looks like it went stale when the approach changed, and it now contradicts the generated section further down. Worth a quick edit before merge since the body is the durable record for why this looks the way it does.

nickgs1337:
Unrelated to this change, just recording it so nobody redoes the digging. `processSubscription` (network/api/websocket/routes.go:132) calls `NewClient`, which already starts loopIn/loopOut, before `HandleClientInsertion` registers the client, so on paper a client that fails fast could run its `handleClientDelete` defer before it is ever registered and leave entries stuck in the hub maps until shutdown. I tried to reproduce it against this branch (200 connections RST right after the handshake, then 300 connections x 4 buffered subscribe requests with a clean close) and every counter came back zero, so today the caller always wins on scheduling alone. There is no guard behind it though, so it could invert quietly if registration ever gets slower. Cheap hardening if we want it: refuse insertion when the client ctx is already done, or register before starting the goroutines.

websocket/config.go
Test0rMaik: No good reason — you're right. Moved pingPeriod/pongWait onto resolvedLimits, set once in resolve() at NewHub time; the hub is built once before any client starts and never mutates them afterward, so this kills the race by construction. Deleted the atomic.Pointer/keepaliveTimings/getter machinery entirely — the const block is back. Pushed in ffcef54.

websocket/config.go
Test0rMaik: You're right, and that's a real gap in what I claimed — getPingPeriod()/getPongWait() were independent Loads consumed by different goroutines, so a Swap between them could produce exactly the mixed pair the comment said couldn't happen. Rather than patch the claim, moved the timings onto resolvedLimits (set once at construction, never mutated after), which sidesteps the question entirely instead of trying to make it true. Pushed in ffcef54.

websocket/config.go
Test0rMaik: Agreed on all counts. Moved to resolvedLimits, so this whole type/comment deletes itself — one line in the new comment covers the pongWait>pingPeriod invariant now, no duplication with the test. Pushed in ffcef54.

websocket/config.go
Test0rMaik: Moot now — timings live on resolvedLimits, set in resolve() at NewHub time, no package-level var or init() left to worry about ordering for. Pushed in ffcef54.

websocket/client.go
Test0rMaik: Downgraded to Debug. Pushed in ffcef54.

websocket/hardening_test.go
Test0rMaik: Good catch, and good to know — collapsed it to a direct t.Errorf plus return, dropped the channel, the select case, and the comment. Pushed in ffcef54.

websocket/config.go
nickgs1337: Yep, this is exactly it. Timings sit on resolvedLimits now, set once in resolve(), and the const block is back, so the race is gone by construction rather than synchronized away.

websocket/config.go
nickgs1337: Sorted. The struct and the claim both went away, and each connection now reads the pair off the hub, so the pong > ping invariant actually holds for the whole connection instead of just being asserted.

websocket/config.go
nickgs1337: Deleted itself like we figured. Two short comments left, no restatement in the test, and nothing pinned to a specific branch in client.go anymore.

websocket/config.go
nickgs1337: Moot now, agreed. No package level var and no init() left, so there is no initialization order to depend on.

websocket/client.go
nickgs1337: Debug now, and the inline comment says why it fires. Good by me.

websocket/hardening_test.go
nickgs1337: Yep, straight t.Errorf plus return in the handler now, channel and select case gone. The 2s timeout still catches it either way.

nickgs1337:
@Test0rMaik Everything fixed, please, rebase the PR so we can merge it.

Test0rMaik:
Rebased onto develop (commit 2925313) to pick up #97's changes — had to hand-merge the additive conflicts in `websocket/config.go`/`hardening_test.go` since both PRs touched `Limits`/`resolvedLimits` (this PR's `PingPeriod`/`PongWait` fields alongside #97's `PostWorkers`/`PostQueueSize`). Build, vet, and `go test ./websocket/... -race -count=1` all clean afterward. Ready for merge from my side.

nickgs1337 (approved):

Beroni (approved):

PR #100, opened 2026-08-04
[KLC-2574] perf(indexer): skip ES force-refresh during import-db

fbsobreira:
## Summary
Disable Elasticsearch force-refresh on single-document index writes while the node is in import-db mode, so from-scratch ES reindex is much faster without hurting live searchability.

## Problem
Full-chain Elasticsearch backfill via import-db can take on the order of 10–15 days. Block and epoch paths hard-coded `Refresh: "true"` on every `esapi.IndexRequest`, forcing a Lucene refresh on each write. That is correct for live indexing (immediate search visibility) but is a major bottleneck during import-db reindex.

## Solution
Gate the refresh policy on `IsInImportDBMode` (already plumbed into the indexer as `isInImportMode`):

- **Live:** `Refresh: "true"` (unchanged behavior)
- **Import-db:** `Refresh: "false"` so ES uses normal refresh behavior instead of per-write force-refresh

A small `indexRefresh()` helper is used by both `SaveHeader` (`blockIndex`) and `indexEpochInfo` (`epochIndex`). Bulk index paths already do not force refresh.

## Key Changes
- **Updated:** `indexer/elasticProcessor.go` — `indexRefresh()`; block and epoch `IndexRequest` call sites
- **Updated:** `indexer/elasticProcessor_test.go` — unit + `SaveHeader` / `SaveEpochInfo` coverage for live vs import-db

## Testing
- [x] Targeted: `go test ./indexer/ -run 'TestElasticProcessor_indexRefresh|TestElasticProcessor_SaveHeader_RefreshPolicy|TestElasticProcessor_SaveEpochInfo'`
- [ ] All existing tests pass (`make tests`)
- [ ] Manual testing performed: import-db reindex throughput comparison (optional post-merge)

## Configuration Changes
None. Uses existing `--import-db` / `IsInImportDBMode` plumbing.

## Breaking Changes
None. Live indexing behavior is unchanged. During import-db only, newly written docs may not be immediately searchable until ES’s normal refresh interval.

## Related Issues
Fixes KLC-2574
Related to KLC-28 (historical import-db slow sync investigation)

## Checklist
- [ ] Documentation updated (README, CLAUDE.md, code comments)

## Summary

- Updates Elasticsearch indexing for `import-db` mode.
- Disables per-document force refresh for block headers and epoch information during imports by using `Refresh: "false"`.
- Keeps `Refresh: "true"` for live indexing.
- Adds the shared `indexRefresh()` helper for both indexing paths.
- Adds tests for live and import-db refresh policies.

## Blockchain impact

- Consensus, transaction processing, state management, KVM, and networking are unchanged.
- Node stability and data integrity are unchanged.
- Import performance improves by avoiding forced Elasticsearch refreshes.
- Imported documents may remain unavailable for search until Elasticsearch performs its normal refresh.
- Bulk indexing paths, concurrency behavior, error handling, configuration, and public APIs are unchanged.

### DIFF
--- a/indexer/elasticProcessor.go
+++ b/indexer/elasticProcessor.go
@@ -320,7 +320,7 @@ func (ei *elasticProcessor) SaveHeader(
 		Index:      blockIndex,
 		DocumentID: hex.EncodeToString(headerHash),
 		Body:       bytes.NewReader(buff.Bytes()),
-		Refresh:    "true",
+		Refresh:    ei.indexRefresh(),
 	}
 
 	if err := ei.elasticClient.DoRequest(req); err != nil {
@@ -1848,7 +1848,7 @@ func (ei *elasticProcessor) indexEpochInfo(epoch uint32, epochInfo *data.EpochIn
 		Index:      epochIndex,
 		DocumentID: fmt.Sprint(epoch),
 		Body:       bytes.NewReader(buff.Bytes()),
-		Refresh:    "true",
+		Refresh:    ei.indexRefresh(),
 	}
 
 	return ei.elasticClient.DoRequest(req)
@@ -1857,6 +1857,16 @@ func (ei *elasticProcessor) indexEpochInfo(epoch uint32, epochInfo *data.EpochIn
 func (ei *elasticProcessor) isIndexEnabled(index string) bool {
 	_, isEnabled := ei.enabledIndexes[index]
 	return isEnabled
+}
+
+// indexRefresh returns the Elasticsearch IndexRequest refresh policy.
+// Live mode forces refresh so docs are immediately searchable; import-db
+// skips force-refresh to speed full-chain ES backfill.
+func (ei *elasticProcessor) indexRefresh() string {
+	if ei != nil && ei.txDatabaseProcessor != nil && ei.isInImportMode {
+		return "false"
+	}
+	return "true"
 }
 
 func getTemplateByName(templateName string, templateList map[string]*bytes.Buffer) *bytes.Buffer {
--- a/indexer/elasticProcessor_test.go
+++ b/indexer/elasticProcessor_test.go
@@ -1599,6 +1599,71 @@ func TestBuildAccountInfo(t *testing.T) {
 	})
 }
 
+func TestElasticProcessor_indexRefresh(t *testing.T) {
+	t.Parallel()
+
+	t.Run("live mode forces refresh", func(t *testing.T) {
+		t.Parallel()
+		args := createMockElasticProcessorArgs()
+		args.IsInImportDBMode = false
+		ep := newTestElasticSearchDatabase(&imock.DatabaseWriterStub{}, args)
+		require.Equal(t, "true", ep.indexRefresh())
+	})
+
+	t.Run("import-db mode disables force refresh", func(t *testing.T) {
+		t.Parallel()
+		args := createMockElasticProcessorArgs()
+		args.IsInImportDBMode = true
+		ep := newTestElasticSearchDatabase(&imock.DatabaseWriterStub{}, args)
+		require.Equal(t, "false", ep.indexRefresh())
+	})
+}
+
+func TestElasticProcessor_SaveHeader_RefreshPolicy(t *testing.T) {
+	t.Parallel()
+
+	header := &dataBlock.Block{
+		Header: &dataBlock.BlockHeader{Nonce: 1},
+	}
+	signer := []byte("signer")
+
+	t.Run("live mode sets refresh true", func(t *testing.T) {
+		t.Parallel()
+		var captured *esapi.IndexRequest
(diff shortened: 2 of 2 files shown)


Beroni (approved):

indexer/elasticProcessor.go
```diff
@@ -320,7 +320,7 @@ func (ei *elasticProcessor) SaveHeader(
 		Index:      blockIndex,
 		DocumentID: hex.EncodeToString(headerHash),
 		Body:       bytes.NewReader(buff.Bytes()),
-		Refresh:    "true",
+		Refresh:    ei.indexRefresh(),
```
nickgs1337, line 323: Dropping the forced refresh here breaks revert. RemoveHeader goes through DoBulkRemove, which is a _delete_by_query, and a search cannot see segments that have not refreshed yet, so a rollback landing inside the refresh window matches nothing and leaves the block doc in ES for good. It is silent too, DoBulkRemove never checks res.IsError(). Since we already write with a deterministic DocumentID, can we just delete by ID instead? By-ID reads hit the translog so refresh does not matter, and it is cheaper than delete_by_query. Setting refresh=wait_for here would also close the window but it only papers over it.

indexer/elasticProcessor.go
```diff
@@ -1859,6 +1859,16 @@ func (ei *elasticProcessor) isIndexEnabled(index string) bool {
 	return isEnabled
 }
 
+// indexRefresh returns the Elasticsearch IndexRequest refresh policy.
+// Live mode forces refresh so docs are immediately searchable; import-db
+// skips force-refresh to speed full-chain ES backfill.
+func (ei *elasticProcessor) indexRefresh() string {
+	if ei != nil && ei.txDatabaseProcessor != nil && ei.isInImportMode {
```
nickgs1337, line 1866: `ei != nil` can never be false here. SaveHeader already reads ei.enabledIndexes well before this, so a nil receiver has panicked long ago. The bigger thing is that isInImportMode comes from the embedded txDatabaseProcessor, which is the only reason the second guard exists at all, and the refresh policy for the blocks/epoch indexes has nothing to do with tx processing. We already get arguments.IsInImportDBMode directly in NewElasticProcessor, so can we store it on elasticProcessor itself (name it isImportDBMode so it does not shadow the promoted one) and drop both guards? newTestElasticSearchDatabase needs the same one-liner to stay in sync.

indexer/elasticProcessor.go
```diff
@@ -1848,7 +1848,7 @@ func (ei *elasticProcessor) indexEpochInfo(epoch uint32, epochInfo *data.EpochIn
 		Index:      epochIndex,
 		DocumentID: fmt.Sprint(epoch),
 		Body:       bytes.NewReader(buff.Bytes()),
-		Refresh:    "true",
+		Refresh:    ei.indexRefresh(),
```
nickgs1337, line 1851: nit: nothing forces a refresh once an import-db run finishes, so docs written near the tail end are not searchable until the next auto-refresh. Not data loss, but any tooling that counts docs right after import-db terminates could read low and look like it. An explicit _refresh on the block and epoch indices at end of import would remove the doubt cheaply, assuming there is a clean hook for it. Unrelated to this PR, but delete_by_query looks refresh-dependent in a few other spots in this package too, probably worth its own ticket rather than patching site by site.

indexer/elasticProcessor_test.go
```diff
@@ -1599,6 +1599,71 @@ func TestBuildAccountInfo(t *testing.T) {
 	})
 }
 
+func TestElasticProcessor_indexRefresh(t *testing.T) {
```
nickgs1337, line 1602: nit: TestElasticProcessor_SaveHeader_RefreshPolicy right below already hits both branches through the real SaveHeader path, so this one is mostly redundant. Happy either way, just flagging it if you want to trim.

indexer/elasticProcessor_test.go
```diff
@@ -1599,6 +1599,71 @@ func TestBuildAccountInfo(t *testing.T) {
...
+	t.Run("import-db mode disables force refresh", func(t *testing.T) {
+		t.Parallel()
+		args := createMockElasticProcessorArgs()
+		args.IsInImportDBMode = true
+		ep := newTestElasticSearchDatabase(&imock.DatabaseWriterStub{}, args)
+		require.Equal(t, "false", ep.indexRefresh())
+	})
+}
+
+func TestElasticProcessor_SaveHeader_RefreshPolicy(t *testing.T) {
```
nickgs1337, line 1622: nit: these two subtests only differ by one bool and the expected string, a tiny table over a single body would cut them in half. Worth keeping a t.Run per case so a failure still names which mode broke. Same treatment would apply to the SaveEpochInfo refresh subtests if you go that way.

nickgs1337 (approved):

PR #92, opened 2026-07-20
fix: restore validator sync across restarts and epoch transitions (#88, #90, #91)

MathijsBok:
## Summary

Any node that is not yet synchronized (fallback/redundancy node, a validator resyncing after a restart or network interruption, or a late-joining node) lost track of the validator set at an epoch transition or mid-epoch restart, stalling into bursty request-driven sync (missing blocks). This PR fixes the four cooperating causes and makes the failure loud when state restoration is incomplete. The base change is one commit on top of current develop, followed by the review fix-up commits.

## The four parts

**1. Nodes coordinator: rebuild `publicKeyToValidatorMap` on restore**
`LoadState` restores the per-epoch validator configs but never rebuilt the pubkey lookup, so after any mid-epoch restart the p2p validator-originator gate rejected every gossiped header until the next epoch boundary. The map is now rebuilt and published atomically with the configs under one write lock. `SetNodes` had the same defect on the fast-bootstrap path and now rebuilds the map before returning (the epoch config's write lock is scoped so the rebuild can re-acquire it as a read lock without deadlocking); the now-redundant rebuild in `EpochStartPrepare` is removed. The constructor's unconditional initial save also overwrote the previous run's registry on a restart within the install epoch (permanent BLS signature failures after `LoadState` read genesis data back); it is now gated behind `StartEpoch == 0`.

**2. Consensus worker: keep learning validators while not synchronized**
`ProcessReceivedMessage` learns the pid to pk mapping before the not-synchronized early return, after the full validity pipeline (including BLS verification), so only proven validator keys enter the `PeerShardMapper`. `EpochStartAction` refreshes the elected set via `GetConsensusWhitelistedNodes` so `IsNodeInConsensusGroup` accepts new-epoch validators while syncing. The redundancy inactivity reset runs only on fully valid messages; anything else would let an attacker replay the public validator key to block a backup takeover. Related hardening in the same area: the per-(pk, slot, msgType) message budget check-and-increment now shares one critical section (`tryAddMessageTypeToPublicKey`), closing a TOCTOU where concurrent duplicates via distinct peers could all pass while signature verification ran between check and add; the epoch-start notifier now sorts its handler slice under the write lock (it mutated the slice under a read lock) and its `EpochStartPrepare` guarded the prepare callback behind the wrong nil check.

**3. Startup: fail loudly when the coordinator is not restored**
[...]

PR #108, opened 2026-08-08
fix: rebuild publicKeyToValidatorMap in nodesCoordinator LoadState

MathijsBok:
## Problem

A validator node that is restarted with an existing database stops following the chain per slot. After the restart it commits blocks only in bursts of ~20 blocks every 20 slots (80 seconds with 4s slots), with fully silent windows in between, and every burst starts by re-committing the previous tip block. The behavior persists until the next epoch boundary, after which the node follows per slot again. This was observed in production on fallback/redundancy nodes (`--redundancy-level=1`) but affects any restart from local storage.

## Root cause

`indexHashedNodesCoordinator.publicKeyToValidatorMap` is only filled in two places: the constructor and `EpochStartPrepare`. `LoadState` (the restart-from-storage path, called by the storage bootstrapper) restores the per-epoch validator configs into `nodesConfig` but never rebuilds the map.

On a restart, the epoch-start bootstrap returns `Parameters{NodesConfig: nil}` (it is only populated on the network start-in-epoch path), so the constructor falls back to the genesis nodes file and seeds the map with genesis-era validators only. `LoadState` then restores the correct per-epoch configs (so header signature verification works), but the map stays genesis-only until the next `EpochStartPrepare`.

### Full cascade (traced through code)

1. `LoadState` (`nodesCoordinator.go:252`) sets `nodesConfig` but does not rebuild `publicKeyToValidatorMap`
2. `GetValidatorWithPublicKey` (`nodesCoordinator.go:418`) looks up the stale (genesis-only) map and returns `ErrValidatorNotFound` for all current validators
3. `PeerShardMapper.getPeerInfoWithNodesCoordinator` (`peerShardMapper.go:127`) classifies all current validators as `core.UnknownPeer`
4. `p2pAntiflood.IsOriginatorElectedForTopic` (`p2pAntiflood.go:106`) rejects them because `PeerType != ValidatorPeer`
5. `SingleDataInterceptor.ProcessReceivedMessage` (`singleDataInterceptor.go:137`) drops every gossip header that is not on the request whitelist
6. The fork detector's `ProbableHighestNonce` never advances (no gossip headers reach the pool)
7. The node considers itself synchronized (`hasLastBlock = ProbableHighestNonce <= currentNonce`)
8. Only the 20-slot `requestHeadersIfSyncIsStuck` cycle (`SlotModulusTriggerWhenSyncIsStuck = 20`) brings headers in via explicit requests (which bypass the validator gate through the request whitelist)
9. Each request batch brings `MaxHeadersToRequestInAdvance = 20` headers, producing the observed burst-of-20 pattern every 80 seconds

### Why it self-heals at epoch boundary

`EpochStartPrepare` (line 781) calls `fillPublicKeyToValidatorMap()`, which rebuilds the map with the correct validators. After that, gossip headers are accepted and the node follows per slot.

### Second bug: constructor saveState overwrites real data

The constructor unconditionally calls `saveState(hash(selfPubKey))` (line 130). On a restart within the install epoch (`StartEpoch > 0`), this overwrites the previously saved real validator state with genesis data. A subsequent `LoadState` then reads back genesis entries instead of the correct validators.

### 1. Rebuild publicKeyToValidatorMap in LoadState

`computePublicKeyToValidatorMap` (extracted helper) merges the elected and eligible validators of all restored epoch configs into one lookup map. `LoadState` computes this map from the freshly restored `nodesConfig` *before* acquiring the write lock, then publishes both `nodesConfig` and the rebuilt map atomically under one `mutNodesConfig.Lock`, so `GetValidatorWithPublicKey` can never observe a state where configs are updated but the map is stale.

### 2. Guard initial saveState behind StartEpoch == 0

The constructor's `saveState` call persists the genesis-seeded `nodesConfig` under `hash(selfPubKey)`. On a restart within the install epoch (`StartEpoch > 0`), this would overwrite the previously saved real validator state with genesis data, causing `LoadState` to read back genesis entries. The constructor now only saves on a fresh genesis start (`StartEpoch == 0`).

## Validation

- Regression tests:
  - `TestNodesCoordinator_LoadStateRefillsPublicKeyToValidatorMap`: seeds a coordinator the way a restart does, verifies both elected and eligible validators resolve after `LoadState`, and genesis-seeded entries are gone.
  - `TestNodesCoordinator_ConstructorDoesNotOverwriteSavedStateOnRestart`: verifies a restart within the install epoch (`Epoch == StartEpoch > 0`, matching production shape) does not overwrite saved state.
  - `TestNodesCoordinator_ConstructorSavesStateOnGenesisStart`: verifies the constructor does save state when `StartEpoch == 0`, protecting the genesis-start path against a future inversion.
- `go test ./sharding/... -race` green; `go vet` clean.

## Notes

- No wire or storage format changes; fixed and unfixed nodes interoperate, so no coordinated rollout is needed.
- A related pre-existing gap exists on the network start-in-epoch path (`SetNodes` for the previous epoch also does not refill the map). It is a much smaller window, self-healing at the next boundary, and intentionally left out of this fix; it can be tracked separately.

## Test plan

- [ ] Verify `go test ./sharding/... -race` passes
- [ ] Verify `go vet ./sharding/...` is clean
- [ ] Validate on a testnet fallback node: mid-epoch restart on the fixed build should resume per-slot following immediately after catch-up, with zero silent windows

## Summary

- Rebuilds `publicKeyToValidatorMap` from restored validator configurations during `nodesCoordinator.LoadState`.
- Gives later epoch configurations precedence for duplicate public keys.
- Updates validator configurations and the lookup map under the same lock.
- Preserves the existing lookup map when restored data produces an empty map and logs a warning.
- Prevents current validators from being classified as unknown after restart.
- Prevents gossip headers from being dropped during validator-node synchronization.
- Prevents persisted validator state from being overwritten by genesis data when `StartEpoch > 0`.
- Saves initial state only when `StartEpoch == 0`.
- Improves node stability and validator-state integrity after restart.
- Limits the change to sharding state management and validator networking.
- Introduces no wire or storage format changes.
- Adds regression tests for map rebuilding, multi-epoch precedence, and constructor persistence.
- Reports passing race tests and `go vet` checks.

### DIFF
--- a/sharding/nodesCoordinator.go
+++ b/sharding/nodesCoordinator.go
@@ -127,10 +127,12 @@ func NewNodesCoordinator(arguments ArgNodesCoordinator) (*indexHashedNodesCoordi
 		indexHashedNodesCoordinator: ihgs,
 	}
 
-	err = ihgs.saveState(ihgs.savedStateKey)
-	if err != nil {
-		log.Error("saving initial nodes coordinator config failed",
-			"error", err.Error())
+	if arguments.StartEpoch == 0 {
+		err = ihgs.saveState(ihgs.savedStateKey)
+		if err != nil {
+			log.Error("saving initial nodes coordinator config failed",
+				"error", err.Error())
+		}
 	}
 
 	log.Info("new nodes config is set for epoch", "epoch", arguments.Epoch)
@@ -281,8 +283,12 @@ func (ihgs *indexHashedNodesCoordinator) LoadState(key []byte) error {
 	}
 
 	displayNodesConfigInfo(nodesConfig)
+
+	publicKeyToValidatorMap := ihgs.computePublicKeyToValidatorMap(nodesConfig)
+
 	ihgs.mutNodesConfig.Lock()
 	ihgs.nodesConfig = nodesConfig
+	ihgs.publicKeyToValidatorMap = publicKeyToValidatorMap
 	ihgs.mutNodesConfig.Unlock()
 
 	ihgs.stateReady.Store(true)
@@ -372,10 +378,18 @@ func (ihgs *indexHashedNodesCoordinator) fillPublicKeyToValidatorMap() {
 	ihgs.mutNodesConfig.Lock()
 	defer ihgs.mutNodesConfig.Unlock()
 
+	ihgs.publicKeyToValidatorMap = ihgs.computePublicKeyToValidatorMap(ihgs.nodesConfig)
+}
+
+// computePublicKeyToValidatorMap merges the elected and eligible validators of
+// all given epoch configs into one lookup map, later epochs taking precedence
+func (ihgs *indexHashedNodesCoordinator) computePublicKeyToValidatorMap(
+	nodesConfig map[uint32]*epochNodesConfig,
+) map[string]Validator {
 	index := 0
-	epochList := make([]uint32, len(ihgs.nodesConfig))
+	epochList := make([]uint32, len(nodesConfig))
 	mapAllValidators := make(map[uint32]map[string]Validator)
-	for epoch, epochConfig := range ihgs.nodesConfig {
+	for epoch, epochConfig := range nodesConfig {
 		epochConfig.mutNodesMaps.RLock()
 		mapAllValidators[epoch] = ihgs.createPublicKeyToValidatorMap(epochConfig.electedList, epochConfig.eligibleList)
 		epochConfig.mutNodesMaps.RUnlock()
@@ -388,13 +402,15 @@ func (ihgs *indexHashedNodesCoordinator) fillPublicKeyToValidatorMap() {
 		return epochList[i] < epochList[j]
 	})
 
-	ihgs.publicKeyToValidatorMap = make(map[string]Validator)
+	publicKeyToValidatorMap := make(map[string]Validator)
 	for _, epoch := range epochList {
 		validatorsForEpoch := mapAllValidators[epoch]
 		for pubKey, vInfo := range validatorsForEpoch {
-			ihgs.publicKeyToValidatorMap[pubKey] = vInfo
+			publicKeyToValidatorMap[pubKey] = vInfo
 		}
 	}
+
+	return publicKeyToValidatorMap
 }
 
 func (ihgs *indexHashedNodesCoordinator) createPublicKeyToValidatorMap(
--- a/sharding/nodesCoordinator.go
+++ b/sharding/nodesCoordinator.go
@@ -284,11 +284,14 @@ func (ihgs *indexHashedNodesCoordinator) LoadState(key []byte) error {
 
 	displayNodesConfigInfo(nodesConfig)
 
-	publicKeyToValidatorMap := ihgs.computePublicKeyToValidatorMap(nodesConfig)
-
 	ihgs.mutNodesConfig.Lock()
 	ihgs.nodesConfig = nodesConfig
-	ihgs.publicKeyToValidatorMap = publicKeyToValidatorMap
+	publicKeyToValidatorMap := ihgs.computePublicKeyToValidatorMap(ihgs.nodesConfig)
+	if len(publicKeyToValidatorMap) > 0 {
+		ihgs.publicKeyToValidatorMap = publicKeyToValidatorMap
+	} else {
+		log.Warn("LoadState: restored registry produced empty validator map, keeping existing lookup")
+	}
 	ihgs.mutNodesConfig.Unlock()
 
 	ihgs.stateReady.Store(true)
--- a/sharding/nodesCoordinator_test.go
+++ b/sharding/nodesCoordinator_test.go
@@ -809,37 +809,34 @@ func TestNodesCoordinator_LoadStateRefillsPublicKeyToValidatorMap(t *testing.T) 
 func TestNodesCoordinator_LoadStateMultiEpochPrecedence(t *testing.T) {
 	t.Parallel()
 
-	sharedPK := []byte("shared-pk-appears-in-both-epochs")
+	sharedPK := []byte("shared-pk-appears-in-all-epochs")
 	bootStorer := mock.NewStorerMock()
 
-	epoch4Elected := []Validator{
-		mock.NewValidatorMock(sharedPK, sharedPK, DefaultSelectionChances, 10),
-		mock.NewValidatorMock([]byte("pk4_1"), []byte("pk4_1"), DefaultSelectionChances, 1),
-		mock.NewValidatorMock([]byte("pk4_2"), []byte("pk4_2"), DefaultSelectionChances, 2),
-		mock.NewValidatorMock([]byte("pk4_3"), []byte("pk4_3"), DefaultSelectionChances, 3),
-	}
-	epoch5Elected := []Validator{
-		mock.NewValidatorMock(sharedPK, sharedPK, DefaultSelectionChances, 20),
-		mock.NewValidatorMock([]byte("pk5_1"), []byte("pk5_1"), DefaultSelectionChances, 1),
-		mock.NewValidatorMock([]byte("pk5_2"), []byte("pk5_2"), DefaultSelectionChances, 2),
-		mock.NewValidatorMock([]byte("pk5_3"), []byte("pk5_3"), DefaultSelectionChances, 3),
+	baseEpoch := uint32(4)
+	numEpochs := uint32(5)
+	highestEpoch := baseEpoch + numEpochs - 1
+
+	makeElected := func(epoch, index uint32) []Validator {
+		return []Validator{
+			mock.NewValidatorMock(sharedPK, sharedPK, DefaultSelectionChances, index),
+			mock.NewValidatorMock([]byte(fmt.Sprintf("pk%d_1", epoch)), []byte(fmt.Sprintf("pk%d_1", epoch)), DefaultSelectionChances, 1),
+			mock.NewValidatorMock([]byte(fmt.Sprintf("pk%d_2", epoch)), []byte(fmt.Sprintf("pk%d_2", epoch)), DefaultSelectionChances, 2),
+			mock.NewValidatorMock([]byte(fmt.Sprintf("pk%d_3", epoch)), []byte(fmt.Sprintf("pk%d_3", epoch)), DefaultSelectionChances, 3),
+		}
 	}
 
 	args := createArguments()
 	args.BootStorer = bootStorer
-	args.ElectedNodes = epoch4Elected
+	args.ElectedNodes = makeElected(baseEpoch, baseEpoch*10)
 	args.EligibleNodes = createDummyNodesList(0, "eligible")
-	args.Epoch = 4
+	args.Epoch = baseEpoch
+	args.StartEpoch = baseEpoch
 	coordinator, err := NewNodesCoordinator(args)
 	require.Nil(t, err)
 
-	selector5, err := coordinator.createSelector(&epochNodesConfig{electedList: epoch5Elected})
-	require.Nil(t, err)
-	coordinator.nodesConfig[5] = &epochNodesConfig{
-		electedList:  epoch5Elected,
-		eligibleList: []Validator{},
-		waitingList:  []Validator{},
-		selector:     selector5,
+	for ep := baseEpoch + 1; ep <= highestEpoch; ep++ {
+		err = coordinator.SetNodes(makeElected(ep, ep*10), []Validator{}, []Validator{}, ep)
+		require.Nil(t, err)
 	}
 
 	savedKey := []byte("multi-epoch-key")
@@ -858,8 +855,8 @@ func TestNodesCoordinator_LoadStateMultiEpochPrecedence(t *testing.T) {
 
 	v, err := coordinatorLoad.GetValidatorWithPublicKey(sharedPK)
 	require.Nil(t, err)
-	require.Equal(t, uint32(20), v.Index(),
-		"later epoch (5) should take precedence over earlier epoch (4)")
+	require.Equal(t, highestEpoch*10, v.Index(),
+		"later epoch should take precedence over earlier epochs")
 }
 
 func TestNodesCoordinator_ConstructorDoesNotOverwriteSavedStateOnRestart(t *testing.T) {
@@ -901,6 +898,7 @@ func TestNodesCoordinator_ConstructorDoesNotOverwriteSavedStateOnRestart(t *test
 	thirdArgs.Hasher = firstArgs.Hasher
 	thirdArgs.ElectedNodes = genesisElected
 	thirdArgs.EligibleNodes = genesisEligible
+	thirdArgs.Epoch = 5
 	thirdArgs.StartEpoch = 5
 	thirdCoordinator, err := NewNodesCoordinator(thirdArgs)
 	require.Nil(t, err)
--- a/sharding/nodesCoordinator_test.go
+++ b/sharding/nodesCoordinator_test.go
@@ -806,6 +806,62 @@ func TestNodesCoordinator_LoadStateRefillsPublicKeyToValidatorMap(t *testing.T) 
 	require.Equal(t, ErrValidatorNotFound, err)
 }
 
+func TestNodesCoordinator_LoadStateMultiEpochPrecedence(t *testing.T) {
+	t.Parallel()
+
+	sharedPK := []byte("shared-pk-appears-in-both-epochs")
+	bootStorer := mock.NewStorerMock()
+
+	epoch4Elected := []Validator{
+		mock.NewValidatorMock(sharedPK, sharedPK, DefaultSelectionChances, 10),
+		mock.NewValidatorMock([]byte("pk4_1"), []byte("pk4_1"), DefaultSelectionChances, 1),
+		mock.NewValidatorMock([]byte("pk4_2"), []byte("pk4_2"), DefaultSelectionChances, 2),
+		mock.NewValidatorMock([]byte("pk4_3"), []byte("pk4_3"), DefaultSelectionChances, 3),
+	}
+	epoch5Elected := []Validator{
+		mock.NewValidatorMock(sharedPK, sharedPK, DefaultSelectionChances, 20),
+		mock.NewValidatorMock([]byte("pk5_1"), []byte("pk5_1"), DefaultSelectionChances, 1),
+		mock.NewValidatorMock([]byte("pk5_2"), []byte("pk5_2"), DefaultSelectionChances, 2),
+		mock.NewValidatorMock([]byte("pk5_3"), []byte("pk5_3"), DefaultSelectionChances, 3),
+	}
+
+	args := createArguments()
+	args.BootStorer = bootStorer
+	args.ElectedNodes = epoch4Elected
+	args.EligibleNodes = createDummyNodesList(0, "eligible")
+	args.Epoch = 4
+	coordinator, err := NewNodesCoordinator(args)
+	require.Nil(t, err)
+
+	selector5, err := coordinator.createSelector(&epochNodesConfig{electedList: epoch5Elected})
+	require.Nil(t, err)
+	coordinator.nodesConfig[5] = &epochNodesConfig{
+		electedList:  epoch5Elected,
+		eligibleList: []Validator{},
+		waitingList:  []Validator{},
+		selector:     selector5,
+	}
+
+	savedKey := []byte("multi-epoch-key")
+	err = coordinator.saveState(savedKey)
+	require.Nil(t, err)
+
+	argsLoad := createArguments()
+	argsLoad.BootStorer = bootStorer
+	argsLoad.ElectedNodes = createDummyNodesList(4, "genesis")
+	argsLoad.EligibleNodes = createDummyNodesList(0, "eligible")
+	coordinatorLoad, err := NewNodesCoordinator(argsLoad)
+	require.Nil(t, err)
+
+	err = coordinatorLoad.LoadState(savedKey)
+	require.Nil(t, err)
+
+	v, err := coordinatorLoad.GetValidatorWithPublicKey(sharedPK)
+	require.Nil(t, err)
+	require.Equal(t, uint32(20), v.Index(),
+		"later epoch (5) should take precedence over earlier epoch (4)")
+}
+
 func TestNodesCoordinator_ConstructorDoesNotOverwriteSavedStateOnRestart(t *testing.T) {
 	t.Parallel()
 
--- a/sharding/nodesCoordinator_test.go
+++ b/sharding/nodesCoordinator_test.go
@@ -843,20 +843,21 @@ func TestNodesCoordinator_LoadStateMultiEpochPrecedence(t *testing.T) {
 	err = coordinator.saveState(savedKey)
 	require.Nil(t, err)
 
-	argsLoad := createArguments()
-	argsLoad.BootStorer = bootStorer
-	argsLoad.ElectedNodes = createDummyNodesList(4, "genesis")
-	argsLoad.EligibleNodes = createDummyNodesList(0, "eligible")
-	coordinatorLoad, err := NewNodesCoordinator(argsLoad)
-	require.Nil(t, err)
+	for attempt := 0; attempt < 20; attempt++ {
+		argsLoad := createArguments()
+		argsLoad.BootStorer = bootStorer
+		argsLoad.ElectedNodes = createDummyNodesList(4, "genesis")
+		argsLoad.EligibleNodes = createDummyNodesList(0, "eligible")
+		coordinatorLoad, err := NewNodesCoordinator(argsLoad)
+		require.Nil(t, err)
 
-	err = coordinatorLoad.LoadState(savedKey)
-	require.Nil(t, err)
+		err = coordinatorLoad.LoadState(savedKey)
+		require.Nil(t, err)
 
-	v, err := coordinatorLoad.GetValidatorWithPublicKey(sharedPK)
-	require.Nil(t, err)
-	require.Equal(t, highestEpoch*10, v.Index(),
-		"later epoch should take precedence over earlier epochs")
+		v, err := coordinatorLoad.GetValidatorWithPublicKey(sharedPK)
+		require.Nil(t, err)
+		require.Equal(t, highestEpoch*10, v.Index(), "attempt %d", attempt)
+	}
 }
 
 func TestNodesCoordinator_ConstructorDoesNotOverwriteSavedStateOnRestart(t *testing.T) {
--- a/sharding/nodesCoordinator_test.go
+++ b/sharding/nodesCoordinator_test.go
@@ -759,3 +759,137 @@ func TestNodesCoordinator_IsInterfaceNil(t *testing.T) {
 	require.Nil(t, err)
 	require.False(t, check.IfNil(ihgs3))
 }
+
+func TestNodesCoordinator_LoadStateRefillsPublicKeyToValidatorMap(t *testing.T) {
+	t.Parallel()
+
+	currentElected := createDummyNodesList(4, "currentElected")
+	currentEligible := createDummyNodesList(1, "currentEligible")
+	bootStorer := mock.NewStorerMock()
+
+	argsBefore := createArguments()
+	argsBefore.BootStorer = bootStorer
+	argsBefore.ElectedNodes = currentElected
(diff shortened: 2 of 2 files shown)


sharding/nodesCoordinator_test.go
```diff
@@ -759,3 +759,137 @@ func TestNodesCoordinator_IsInterfaceNil(t *testing.T) {
...
+	require.Nil(t, err)
+	require.Equal(t, currentElected[0].PubKey(), validator.PubKey())
+
+	eligibleValidator, err := coordinatorAfter.GetValidatorWithPublicKey(currentEligible[0].PubKey())
+	require.Nil(t, err)
+	require.Equal(t, currentEligible[0].PubKey(), eligibleValidator.PubKey())
+
+	_, err = coordinatorAfter.GetValidatorWithPublicKey(genesisElected[0].PubKey())
+	require.Equal(t, ErrValidatorNotFound, err)
+}
```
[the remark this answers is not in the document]
MathijsBok: Added in `cff88db`. The test persists two epoch configs (4 and 5) with the same public key at different indices and asserts the later epoch wins after LoadState.

sharding/nodesCoordinator.go
```diff
@@ -127,10 +127,12 @@ func NewNodesCoordinator(arguments ArgNodesCoordinator) (*indexHashedNodesCoordi
...
-	err = ihgs.saveState(ihgs.savedStateKey)
-	if err != nil {
-		log.Error("saving initial nodes coordinator config failed",
-			"error", err.Error())
+	if arguments.StartEpoch == 0 {
+		err = ihgs.saveState(ihgs.savedStateKey)
+		if err != nil {
+			log.Error("saving initial nodes coordinator config failed",
+				"error", err.Error())
+		}
```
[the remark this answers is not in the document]
MathijsBok: This is pre-existing behavior on develop (the log-and-continue pattern was there before this PR). Changing error propagation semantics is outside the scope of this fix. The log-and-continue is a deliberate choice: the node can operate with in-memory state even if the initial persistence fails, and making this fatal would prevent startup on nodes with transient storage issues. Can be evaluated separately.

sharding/nodesCoordinator.go
```diff
@@ -281,8 +283,12 @@ func (ihgs *indexHashedNodesCoordinator) LoadState(key []byte) error {
 	}
 
 	displayNodesConfigInfo(nodesConfig)
+
+	publicKeyToValidatorMap := ihgs.computePublicKeyToValidatorMap(nodesConfig)
+
 	ihgs.mutNodesConfig.Lock()
 	ihgs.nodesConfig = nodesConfig
+	ihgs.publicKeyToValidatorMap = publicKeyToValidatorMap
```
fbsobreira: **In scope — regression introduced here.** This assignment replaces the lookup unconditionally, with no check that the restored registry produced anything.

`registryToNodesCoordinator` returns an empty map and a **nil** error when `config.EpochsConfig` is empty or absent, and `json.Unmarshal` accepts `{}` or `{"epochConfigs":{},"currentEpoch":7}` without complaint. Verified on this head: after `LoadState` of such a payload, it returns `nil`, `len(nodesConfig) == 0`, `len(publicKeyToValidatorMap) == 0`, and a validator that resolved before the load now returns `ErrValidatorNotFound`.

Because `LoadState` returns nil, `applyBootInfos` (`baseStorageBootstrapper.go:348`) treats the bootstrap as successful and the node keeps running. `PeerShardMapper.getPeerInfoWithNodesCoordinator` is the only consumer of `GetValidatorWithPublicKey`, so every connected validator is classified `core.UnknownPeer` instead of `core.ValidatorPeer` for the rest of the run — losing validator prioritisation and eviction protection.

**Why this is in scope:** before this PR `LoadState` never touched the map, so a degenerate registry left the constructor's validator set intact and peer classification survived. This diff turns that path destructive.

**Important — do not fix it by returning an error.** PR #92 did exactly that (`ErrEmptyRestoredRegistry`), and it propagated through `loadAndApplyBlocks`' walk-back: every stored slot references the same `NodesCoordinatorConfigKey`, so all attempts fail identically and `handleBlockLoadingError` finally calls `restoreBlockChainToGenesis()` + `SaveLastSlot(0)`. The node discards its chain head and resyncs from genesis — days on mainnet.

The minimal fix keeps the pre-PR property with no new behaviour anywhere: **only replace the map when the newly computed one is non-empty.** `nodesConfig` still gets replaced as before; the lookup simply isn't wiped when there is nothing to replace it with. No new error path, no cascade, nothing outside this function.

sharding/nodesCoordinator.go
```diff
@@ -281,8 +283,12 @@ func (ihgs *indexHashedNodesCoordinator) LoadState(key []byte) error {
 	}
 
 	displayNodesConfigInfo(nodesConfig)
+
+	publicKeyToValidatorMap := ihgs.computePublicKeyToValidatorMap(nodesConfig)
```
fbsobreira: **In scope — regression introduced here.** The map is computed at this line, but `mutNodesConfig` is only taken on line 289. That compute-then-lock window is new in this diff.

If `EpochStartPrepare` (lines 792-797: `SetNodes` then `fillPublicKeyToValidatorMap`) completes between the two, `LoadState` writes back both the disk-derived `nodesConfig` **and** the disk-derived map, discarding the freshly shuffled epoch entirely.

Pre-PR only `nodesConfig` was clobbered — the newly rebuilt `publicKeyToValidatorMap` survived, so `GetValidatorWithPublicKey` still resolved the new epoch's validators. Now they are lost too, and `PeerShardMapper` reports them as `core.UnknownPeer` until the next epoch start rebuilds the map. So this diff strictly widens an existing race.

Fix is one line of movement: compute inside the same `Lock()` that publishes, or recompute from `ihgs.nodesConfig` after the assignment. Either closes the window without touching anything else.

sharding/nodesCoordinator_test.go
```diff
@@ -759,3 +759,193 @@ func TestNodesCoordinator_IsInterfaceNil(t *testing.T) {
...
+
+func TestNodesCoordinator_LoadStateMultiEpochPrecedence(t *testing.T) {
```
fbsobreira, line 809: **In scope — this is the regression test for this PR, and it mostly does not catch the regression.**

Two epochs and a single iteration means Go's randomized map iteration produces the ascending order by chance most of the time. Measured on this head:

1. Deleted the `sort.Slice` from `computePublicKeyToValidatorMap` — the exact ordering guarantee this test exists to protect
2. Confirmed the package still builds (`sort` is used elsewhere, so the import stays)
3. `go test -run TestNodesCoordinator_LoadStateMultiEpochPrecedence -count=40`

Result: **11 failures, 29 passes.** With the mutation in place the whole `sharding` package still reports `ok`, so no other test catches it either.

A future refactor that drops or inverts the epoch ordering therefore lands green roughly 7 times in 10, and when it does trip it reads as flake rather than regression. This is the same weakness the equivalent test had on #102; the form that held up on #92 was **5 epochs × 20 iterations**, which makes accidental ascending order negligible.

sharding/nodesCoordinator_test.go
```diff
@@ -759,3 +759,193 @@ func TestNodesCoordinator_IsInterfaceNil(t *testing.T) {
...
+		mock.NewValidatorMock([]byte("pk5_1"), []byte("pk5_1"), DefaultSelectionChances, 1),
+		mock.NewValidatorMock([]byte("pk5_2"), []byte("pk5_2"), DefaultSelectionChances, 2),
+		mock.NewValidatorMock([]byte("pk5_3"), []byte("pk5_3"), DefaultSelectionChances, 3),
+	}
+
+	args := createArguments()
+	args.BootStorer = bootStorer
+	args.ElectedNodes = epoch4Elected
+	args.EligibleNodes = createDummyNodesList(0, "eligible")
+	args.Epoch = 4
```
fbsobreira: **In scope — new fixture, unreachable state.** `args.Epoch = 4` with `StartEpoch` left at 0.

`saveState` persists a registry whose `CurrentEpoch` is 0 (`NodesCoordinatorToRegistry` fills it from the saving coordinator's `currentEpoch`, i.e. its `StartEpoch`) while its only epoch configs are 4 and 5. After `LoadState` at line 856 the coordinator sits at `currentEpoch == 0` with no config for epoch 0 — a state production cannot produce, since `cmd/node/node.go` sets `currentEpoch := startEpoch`.

The test asserts only on `GetValidatorWithPublicKey` and never notices, so the fixture cannot catch a `LoadState` regression that installs a config set with no entry for the epoch it just adopted — `ComputeConsensusGroup` at that epoch would return `ErrEpochNodesConfigDoesNotExist` while `stateReady` is true.

Same class at **line 905**: `thirdArgs` sets `StartEpoch` but never `.Epoch` (0 occurrences of `thirdArgs.Epoch` in the file), so the assertions there run against a combination `node.go` cannot produce either. That one was flagged on #102 and carried over.

Setting `Epoch` and `StartEpoch` together in both fixtures makes them match the real startup shape.

sharding/nodesCoordinator.go
```diff
@@ -372,10 +378,18 @@ func (ihgs *indexHashedNodesCoordinator) fillPublicKeyToValidatorMap() {
...
+
+// computePublicKeyToValidatorMap merges the elected and eligible validators of
+// all given epoch configs into one lookup map, later epochs taking precedence
+func (ihgs *indexHashedNodesCoordinator) computePublicKeyToValidatorMap(
+	nodesConfig map[uint32]*epochNodesConfig,
+) map[string]Validator {
 	index := 0
-	epochList := make([]uint32, len(ihgs.nodesConfig))
+	epochList := make([]uint32, len(nodesConfig))
 	mapAllValidators := make(map[uint32]map[string]Validator)
```
fbsobreira, line 394: **Out of scope — noted for a follow-up PR, not a merge condition.**

This builds `mapAllValidators` (a map-of-maps, one full `map[string]Validator` per stored epoch, up to `nodeCoordinatorStoredEpochs` = 10) and then copies every entry again into the result, discarding the first. Sorting the epoch list and merging `electedList`/`eligibleList` straight into the result gives identical output in one pass, and makes `createPublicKeyToValidatorMap` (line 416, no other caller in the repo) deletable.

The inefficiency is pre-existing — this diff only relocates the code into the extracted helper — so I would leave it alone here rather than grow the diff. Mentioning it because the extraction makes it newly visible, and it now runs on every `LoadState` as well as every rebuild.

sharding/nodesCoordinator_test.go
```diff
@@ -759,3 +759,193 @@ func TestNodesCoordinator_IsInterfaceNil(t *testing.T) {
...
+	coordinator, err := NewNodesCoordinator(args)
+	require.Nil(t, err)
+
+	selector5, err := coordinator.createSelector(&epochNodesConfig{electedList: epoch5Elected})
```
fbsobreira: Optional, your call — contained to a test this diff adds.

These lines hand-roll what `SetNodes(elected, eligible, waiting, epoch)` already does: build the `epochNodesConfig`, call `createSelector`, store it in `nodesConfig[epoch]` — and do it without holding `mutNodesConfig`. `coordinator.SetNodes(epoch5Elected, []Validator{}, []Validator{}, 5)` replaces all eight lines.

As written, any future field added to `epochNodesConfig` or setup step added to `SetNodes` is silently skipped here, so the test keeps passing while no longer representing a state the coordinator can actually reach.

sharding/nodesCoordinator_test.go
```diff
@@ -759,3 +759,193 @@ func TestNodesCoordinator_IsInterfaceNil(t *testing.T) {
...
+	secondArgs.SelfPublicKey = firstArgs.SelfPublicKey
+	secondArgs.Hasher = firstArgs.Hasher
+	secondArgs.ElectedNodes = genesisElected
+	secondArgs.EligibleNodes = genesisEligible
+	secondArgs.Epoch = 5
+	secondArgs.StartEpoch = 5
+	_, err = NewNodesCoordinator(secondArgs)
+	require.Nil(t, err)
+
+	thirdArgs := createArguments()
```
fbsobreira, line 896: Optional — `secondArgs` and `thirdArgs` differ only in `Epoch`, and the second coordinator is discarded with `_` immediately after construction.

Constructing `NewNodesCoordinator(secondArgs)` once and calling `LoadState` on that instance proves the same thing — that construction did not overwrite the saved state — with about ten fewer lines. As written a reader has to diff the two blocks character by character to discover the difference is not load-bearing, and any future edit has to be applied twice.

sharding/nodesCoordinator_test.go
```diff
@@ -759,3 +759,193 @@ func TestNodesCoordinator_IsInterfaceNil(t *testing.T) {
...
+	require.Nil(t, err)
+
+	args2 := createArguments()
+	args2.BootStorer = bootStorer
+	args2.SelfPublicKey = args.SelfPublicKey
+	args2.Hasher = args.Hasher
+	args2.ElectedNodes = createDummyNodesList(4, "otherElected")
+	args2.EligibleNodes = createDummyNodesList(1, "otherEligible")
+	args2.StartEpoch = 1
+	coordinator2, err := NewNodesCoordinator(args2)
```
fbsobreira, line 941: Optional — this asserts the constructor's `Put` indirectly, by building a second coordinator purely to have a receiver for `LoadState` and inferring the write from `GetValidatorWithPublicKey` succeeding.

`bootStorer.Get(append([]byte(core.NodesCoordinatorRegistryKeyPrefix), coordinator.savedStateKey...))` asserts the same fact directly. The indirection also means this test fails for unrelated reasons — a registry marshalling change, a `registryToNodesCoordinator` change, a `publicKeyToValidatorMap` change — pointing a future reader at the wrong subsystem instead of at the constructor's save.

fbsobreira (changes requested):
## Review at `cff88dbf` — the isolation worked

Scope is two files and nothing here touches consensus-critical code: no redundancy, no relay, no elected-set widening. That makes this a far easier change to reason about than the bundled version, and the `SetNodes` gap being **explicitly scoped out in the description** rather than silently missed is exactly right.

I've split the findings by one test: **does this diff introduce or worsen it?** The intent is to keep this to one focused round rather than letting it accumulate adjacent fixes.

### In scope — 4 items, all regressions this diff creates

| Where | What |
|---|---|
| `nodesCoordinator.go:291` | `LoadState` now wipes the lookup on a degenerate registry. Pre-PR it never touched the map, so the constructor's copy survived |
| `nodesCoordinator.go:287` | Map computed outside `mutNodesConfig` — new compute-then-lock window; pre-PR only `nodesConfig` was lost |
| `nodesCoordinator_test.go:809` | The regression test for this PR catches the regression 11 times in 40 (measured) |
| `nodesCoordinator_test.go:832` | New fixtures build `Epoch`/`StartEpoch` combinations production cannot reach |

Both production items fix **inside `LoadState` alone** — no new files, no new error paths, no new surface.

### Explicitly out of scope — for a separate PR

These are real, but they are pre-existing on `develop` and untouched by this diff. Raising them here only so they are recorded, **not** as merge conditions:

- **`SetNodes` does not rebuild the map** (`nodesCoordinator.go:826`, not in this diff). Your description already calls this out and defers it — I agree with that call. On the network start-in-epoch path `cmd/node/node.go:376` calls it right after construction, so the window is real but smaller and self-healing at the next boundary.
- **`mock.NewValidatorMock` silently discards its `owner` argument** (`sharding/mock/validatorMock.go:17`, not in this diff). Consequence: all four new round-trip tests would pass on a build that drops `OwnerAddress` entirely, since every fixture validator has a nil owner either way. Worth fixing in the mock eventually; doing it here would touch a file shared across the package for no benefit to this change.

### Note from PR #92

Two of these have known-bad "obvious" fixes, learned the expensive way:

1. **Do not make `LoadState` return an error on an empty registry.** #92 shipped exactly that, and it cascaded through the storage bootstrapper's walk-back into `restoreBlockChainToGenesis()` + `SaveLastSlot(0)` — a full resync from genesis. Details inline.
2. **The `StartEpoch == 0` guard in this PR is the flat form of a guard #92 later had to replace.** Not a blocker here and not something I'd change in this PR — flagging it so it is a known follow-up rather than a surprise.

Requesting changes on the four in-scope items only.

MathijsBok:
## Diff summary for `315d163` (addressing review at `cff88dbf`)

All four in-scope items from @fbsobreira's review are addressed in this commit. The two optional items addressed are noted at the end. Inline replies on each comment have the per-item detail.

---

### 1. Degenerate registry wipes validator lookup (`nodesCoordinator.go:291`)

**Problem:** `computePublicKeyToValidatorMap` returns an empty (but non-nil) map when `EpochsConfig` is empty or absent. The unconditional assignment `ihgs.publicKeyToValidatorMap = publicKeyToValidatorMap` then replaces the constructor's working map with an empty one. Pre-PR `LoadState` never touched the map, so the constructor's copy survived. This diff made that path destructive.

**Fix:** Only replace when the new map is non-empty. Added a `log.Warn` when the guard fires so the event is observable rather than silent.

```go
publicKeyToValidatorMap := ihgs.computePublicKeyToValidatorMap(ihgs.nodesConfig)
if len(publicKeyToValidatorMap) > 0 {
    ihgs.publicKeyToValidatorMap = publicKeyToValidatorMap
} else {
    log.Warn("LoadState: restored registry produced empty validator map, keeping existing lookup")
}
```

**Why not return an error instead?** As noted in the review, PR #92 did exactly that (`ErrEmptyRestoredRegistry`), and it cascaded through `loadAndApplyBlocks`' walk-back into `restoreBlockChainToGenesis()` + `SaveLastSlot(0)`, forcing a full resync from genesis. The guard preserves the pre-PR property (constructor map survives) without introducing a new error path. Normal validator removal flows through `EpochStartPrepare` -> `fillPublicKeyToValidatorMap`, which replaces the map unconditionally (no guard at line 382), so legitimate transitions are unaffected.

**Why `len > 0` and not `!= nil`?** `computePublicKeyToValidatorMap` always returns a non-nil map (`make(map[string]Validator)` at line 406), so a nil check would never trigger.

---

### 2. Compute-then-lock race window (`nodesCoordinator.go:287`)

**Problem:** The map was computed at line 287 but `mutNodesConfig` was only acquired at line 289. If `EpochStartPrepare` (which calls `SetNodes` then `fillPublicKeyToValidatorMap`) completed in that window, `LoadState` would overwrite both `nodesConfig` and the freshly rebuilt map with disk-derived data, discarding the new epoch entirely. Pre-PR only `nodesConfig` was clobbered (the map was never touched by `LoadState`), so this diff widened the race.

**Fix:** Moved `computePublicKeyToValidatorMap` inside the `mutNodesConfig.Lock()` block, computing on the just-assigned `ihgs.nodesConfig`:

```go
ihgs.mutNodesConfig.Lock()
ihgs.nodesConfig = nodesConfig
publicKeyToValidatorMap := ihgs.computePublicKeyToValidatorMap(ihgs.nodesConfig)
// ... guard + assign ...
ihgs.mutNodesConfig.Unlock()
```

**Deadlock safety:** `computePublicKeyToValidatorMap` internally takes `epochConfig.mutNodesMaps.RLock()` per epoch. This nests `mutNodesConfig` (outer) -> `mutNodesMaps` (inner), which is the same lock order used by `fillPublicKeyToValidatorMap` (line 378) and `SetNodes` (line 833/842). No call site in the codebase reverses this order. Additionally, the `epochNodesConfig` instances assigned to `ihgs.nodesConfig` here are freshly created by `registryToNodesCoordinator` and not yet visible to any other goroutine, so the inner locks are uncontended.

---

### 3. Flaky multi-epoch precedence test (`nodesCoordinator_test.go:809`)

**Problem:** The test used 2 epochs and 1 iteration. Go's randomized map iteration produced ascending order by chance ~50% of the time, meaning with `sort.Slice` removed from `computePublicKeyToValidatorMap`, the test still passed 29 out of 40 runs. A future refactor that drops or inverts the epoch ordering would land green roughly 7 times in 10.

**Fix:** Rewritten with 5 epochs (4 through 8), each added via `SetNodes()` instead of hand-rolled config. With 5! = 120 permutations, only 1 is ascending, so accidental pass probability drops from ~50% to ~0.8%.

Using `SetNodes()` also addresses the optional comment at line 836: the test now goes through the same code path that production uses in `EpochStartPrepare`, so any future field added to `epochNodesConfig` is automatically included.

**Verification:** `go test -run TestNodesCoordinator_LoadStateMultiEpochPrecedence -count=40` passes 40/40.

---

### 4. Unreachable `Epoch`/`StartEpoch` fixtures (`nodesCoordinator_test.go:832`, `:898`)

**Problem:** `args.Epoch = 4` with `StartEpoch` left at 0 is a state production cannot reach (`cmd/node/node.go` sets `currentEpoch := startEpoch`). After `LoadState` the coordinator would sit at `currentEpoch == 0` with no config for epoch 0. Similarly, `thirdArgs` set `StartEpoch = 5` but never `Epoch`, defaulting it to 0.

**Fix:**
- Precedence test: `args.Epoch = baseEpoch` and `args.StartEpoch = baseEpoch` (both 4).
- Constructor-overwrite test: added `thirdArgs.Epoch = 5` alongside the existing `thirdArgs.StartEpoch = 5`.

Both now match the production startup shape where a restarting node receives its current epoch as both `Epoch` and `StartEpoch`.

---

### Optional items addressed

| Comment | Status |
|---------|--------|
| Line 836: use `SetNodes()` instead of hand-rolled config | Adopted (part of fix 3) |
| Line 898: simplify `secondArgs`/`thirdArgs` | Left as-is: the second coordinator construction is the assertion target (proving `NewNodesCoordinator` with `StartEpoch > 0` does not overwrite); removing it would change what the test verifies |
| Line 942: direct `bootStorer.Get` assertion | Left as-is: changing assertion strategy would expand the diff beyond review feedback scope |

### Out-of-scope item

| Comment | Status |
|---------|--------|
| Line 391: double-pass allocation in `computePublicKeyToValidatorMap` | Tracked in #109 |

---

### Verification

- `go test ./sharding/... -race -count=1`: green
- `go test -run TestNodesCoordinator_LoadStateMultiEpochPrecedence -count=40`: 40/40 passes
- `go vet ./sharding/...`: clean
- SonarQube: no new issues on changed files (5 pre-existing issues on `nodesCoordinator.go`, all outside changed lines)

sharding/nodesCoordinator.go
MathijsBok: Fixed in `315d163`. The map is now only replaced when non-empty:

```go
if len(publicKeyToValidatorMap) > 0 {
    ihgs.publicKeyToValidatorMap = publicKeyToValidatorMap
} else {
    log.Warn("LoadState: restored registry produced empty validator map, keeping existing lookup")
}
```

Added a `log.Warn` so the guard is not silent. Normal validator removal flows through `EpochStartPrepare` -> `fillPublicKeyToValidatorMap`, which replaces unconditionally (no guard), so legitimate transitions are unaffected.

sharding/nodesCoordinator.go
MathijsBok: Fixed in `315d163`. `computePublicKeyToValidatorMap` is now called inside the `mutNodesConfig.Lock()` block, on the just-assigned `ihgs.nodesConfig`. This matches the existing pattern in `fillPublicKeyToValidatorMap` (line 378) which already nests `mutNodesConfig.Lock` -> `mutNodesMaps.RLock`. No deadlock risk (verified: consistent lock order across the codebase, `-race` clean).

sharding/nodesCoordinator_test.go
MathijsBok: Fixed in `315d163`. Rewritten with 5 epochs (4 through 8) instead of 2, each added via `SetNodes()`. With 5! = 120 permutations, accidental ascending order is ~0.8% vs ~50% before.

Verified: `go test -run TestNodesCoordinator_LoadStateMultiEpochPrecedence -count=40` passes 40/40.

sharding/nodesCoordinator_test.go
MathijsBok: Fixed in `315d163`. Both `Epoch` and `StartEpoch` are now set together: `args.Epoch = baseEpoch` / `args.StartEpoch = baseEpoch` in the precedence test, and `thirdArgs.Epoch = 5` added alongside `thirdArgs.StartEpoch = 5` in the constructor-overwrite test.

sharding/nodesCoordinator.go
MathijsBok: Agreed, leaving it for a follow-up. The extraction makes it visible, but the diff should stay focused.

Tracked in #109.

sharding/nodesCoordinator_test.go
MathijsBok: Adopted in `315d163`. The test now uses `coordinator.SetNodes(makeElected(ep, ep*10), []Validator{}, []Validator{}, ep)` for all additional epochs.

sharding/nodesCoordinator_test.go
MathijsBok: Left as-is for now. The second coordinator construction is the actual assertion target (proving that `NewNodesCoordinator` with `StartEpoch > 0` does not overwrite), so removing it would change what the test verifies. The third is needed as a fresh receiver for `LoadState`. Could revisit in a cleanup pass.

sharding/nodesCoordinator_test.go
MathijsBok: Good point. Left as-is in this round since changing the assertion strategy would expand the diff beyond the review feedback scope. Can be picked up in a cleanup pass.

sharding/nodesCoordinator_test.go
fbsobreira: Verified the two production fixes in code — the guard keeps the previous lookup with a `log.Warn` instead of erroring (no bootstrapper cascade), and the compute now runs inside the same `Lock()` on the just-assigned `ihgs.nodesConfig`. Lock order matches `fillPublicKeyToValidatorMap` and `SetNodes`, so no inversion. Both good.

On this test, one correction — the improvement is real but smaller than the reply states. I mutation-tested it the same way as before: removed the `sort.Slice`, confirmed the package still builds, then ran it.

| | catches a dropped `sort` |
|---|---|
| before (2 epochs) | 11 / 40 |
| now (5 epochs) | **25 / 40** |
| single `go test ./sharding/` run | **fails 7 / 10** |

So a regression still slips through a normal CI run about 3 times in 10.

The `5! = 120 → ~0.8%` reasoning does not apply here: the assertion only depends on **which epoch is visited last** for `sharedPK`, not on the full permutation. With 5 epochs that is roughly a 1-in-5 chance of landing correct by accident, which matches the ~62% catch rate measured.

Also worth noting `-count=40` passing with the sort **present** does not test this — the question is whether it fails when the sort is **absent**.

The piece that made this deterministic on the bundled PR was not more epochs but **repeating the assertion**: wrap the `LoadState` + `GetValidatorWithPublicKey` check in a loop of ~20, each iteration building a fresh map. At ~20% accidental-pass per iteration that compounds to negligible, and it is about three lines:

```go
for attempt := 0; attempt < 20; attempt++ {
    coordinatorLoad, err := NewNodesCoordinator(argsLoad)
    require.Nil(t, err)
    require.Nil(t, coordinatorLoad.LoadState(savedKey))

    v, err := coordinatorLoad.GetValidatorWithPublicKey(sharedPK)
    require.Nil(t, err)
    require.Equal(t, highestEpoch*10, v.Index(), "attempt %d", attempt)
}
```

Not a production risk and not something I would hold the merge on — the shipped code is correct either way. Flagging it because the reply cites a number that would make this look settled when it is not.

MathijsBok: You are right, the 5! reasoning was wrong. The assertion only depends on which epoch is visited last for the shared key, so with 5 epochs that is a ~1/5 accidental pass, not 1/120.

Fixed in `ebd2dbe`. The LoadState + GetValidatorWithPublicKey check now runs in a loop of 20 iterations, each creating a fresh coordinator. At ~20% accidental pass per iteration that compounds to ~10^-14 over 20 iterations.

```go
for attempt := 0; attempt < 20; attempt++ {
    coordinatorLoad, err := NewNodesCoordinator(argsLoad)
    require.Nil(t, err)
    require.Nil(t, coordinatorLoad.LoadState(savedKey))

    v, err := coordinatorLoad.GetValidatorWithPublicKey(sharedPK)
    require.Nil(t, err)
    require.Equal(t, highestEpoch*10, v.Index(), "attempt %d", attempt)
}
```

fbsobreira (approved):

PR #110, opened 2026-08-09
fix: refresh PeerTypeProvider cache after storage bootstrap and include waiting list

### LINKED ISSUE #88: klv_node_type / klv_peer_type metrics stay "observer" after mid-epoch restart until the next epoch switch
## Symptom

When a node that is elected or eligible in the current epoch restarts mid-epoch (crash, maintenance, temporary outage), the `klv_node_type` metric reports `observer` (and `klv_peer_type` reports the observer list) from startup until the next epoch switch. Only after the first epoch-start event does the node report its real status again. Observed in production; consensus itself is unaffected, this is a metrics/status issue.

## Root cause (traced)

The heartbeat sender derives both metrics from `PeerTypeProvider.ComputeForPubKey` (`node/heartbeat/process/sender.go:176-188`). On a cache miss that method returns `core.ObserverList` (`core/process/peer/peerTypeProvider.go:67`).

The cache is only (re)built in two places:

1. At construction: `NewPeerTypeProvider` calls `updateCache(arg.StartEpoch)` (`core/process/peer/peerTypeProvider.go:50`).
2. On the epoch-start event handler (`core/process/peer/peerTypeProvider.go:94`).

`createNewCache` reads validator lists via `nodesCoordinator.GetAllElectedValidatorsKeys` / `GetAllEligibleValidatorsKeys`, which read `nodesConfig` directly (`sharding/nodesCoordinator.go:619-668`).

The startup order makes the constructed cache permanently stale until the next epoch boundary:

- `createNode` (`cmd/node/startup.go:886`) calls `StartHeartbeat` (`cmd/node/node.go:193`), which constructs the `PeerTypeProvider` (`node/heartbeat/componentHandler/heartbeatHandler.go:107`) and immediately builds the cache. At this point `nodesConfig` only holds the genesis-seeded data from the nodesCoordinator constructor.
- Only afterwards does `ef.StartNode()` (`cmd/node/startup.go:1051`) reach `bootstrapper.LoadStorage()` (`node/node.go:315`), which calls `nodesCoordinator.LoadState` (`core/process/sync/storageBootstrap/baseStorageBootstrapper.go:348`) and restores the real per-epoch validator configs.

Nothing refreshes the `PeerTypeProvider` cache after `LoadState`, so a validator whose key is not in the genesis-seeded data resolves to observer on every heartbeat until the first epoch-start event fires `updateCache` again.

## Second gap in the same component

`createNewCache` (`core/process/peer/peerTypeProvider.go:111-129`) only fills the cache from the elected and eligible lists. The waiting list is never added, even though `GetAllWaitingValidatorsKeys` exists on the nodes coordinator. A node in the waiting list therefore always reports `observer`, restart or not.

## Suggested direction

Refresh the `PeerTypeProvider` cache after the storage bootstrap restores the nodes coordinator state (either an explicit refresh after `LoadFromStorage`, or let the provider observe the same restore path), and include the waiting list in `createNewCache`.

## Relation to PR #85

Independent of #85. That PR rebuilds `publicKeyToValidatorMap` (header interception path) in `LoadState`; the `PeerTypeProvider` cache reads `nodesConfig` through the getters above and is not touched by that fix.

MathijsBok:
## Context

Closes #88.

When a node restarts mid-epoch (crash, maintenance, temporary outage), the heartbeat metrics `klv_node_type` and `klv_peer_type` report every post-genesis validator as `observer` from startup until the next epoch boundary. Consensus is unaffected; only the metrics/status reporting is wrong.

### Root cause: stale cache after mid-epoch restart

The startup sequence builds the `PeerTypeProvider` cache before the real validator data is available:

1. `createNodesCoordinator` seeds the coordinator with genesis/bootstrap config only.
2. `createNode` -> `StartHeartbeat` -> `NewPeerTypeProvider` -> `updateCache()` populates the cache from the not-yet-loaded coordinator. At this point `nodesConfig` only holds genesis-era keys.
3. Much later, `StartConsensus` -> `LoadStorage` -> `nodesCoordinator.LoadState()` restores the real per-epoch validator configs from disk.

Nothing refreshes the `PeerTypeProvider` cache after step 3. The cache stays populated with genesis keys. The only existing refresh path is the epoch-start event handler, which only fires on a *new* epoch boundary, not on a mid-epoch restart.

The effect is wrong in both directions: post-genesis validators (eligible/elected in the current epoch) are downgraded to "observer", while genesis-era keys that are no longer active get promoted to "elected" because they still appear in the genesis `nodesSetup.json`.

### Cache refresh after storage bootstrap

- Export `updateCache` as `UpdateCache` on `PeerTypeProvider`, so the heartbeat subsystem can trigger a rebuild externally.
- Add `RefreshPeerTypeCache(epoch uint32)` to the `HeartbeatHandler` interface and its concrete implementation, delegating to `PeerTypeProvider.UpdateCache`.
- Call `RefreshPeerTypeCache(epoch)` from `Node.StartConsensus()` after `LoadStorage()` completes and the epoch has been determined from the chain. At this point the nodes coordinator has real validator data from `LoadState`, and the epoch is correctly derived from the restored block header.

### Two guards so the refresh can never make things worse

Both were raised in review and both are addressed:

- **Skip the refresh when no current block header exists.** When `LoadStorage` cannot restore a header (fresh sync, wiped storage, bootstrap failure), the epoch falls back to the genesis epoch. Refreshing with that epoch could replace a populated cache with wrong data. The call is guarded with `!check.IfNil(crtBlockHeader)`, so the refresh only runs when a restored header proves the coordinator has real data for that epoch.
- **Keep the previous cache when the coordinator cannot serve the epoch.** `createNewCache` now returns `(map, ok)` and fails on *any* getter error (the coordinator returns `ErrEpochNodesConfigDoesNotExist` for unknown epochs). `UpdateCache` keeps the previous cache and logs a Warn (with the error value) instead of installing an empty or partial map over valid data. This restores the keep-previous guard that PR #92 introduced for the same failure mode.

### Scope note: waiting-list support was split out

An earlier revision of this PR also seeded the cache with the waiting list. Review showed that change has consequences beyond this fix: `Monitor.shouldSkipValidator` and `heartbeatMessageInfo.GetIsValidator` only count elected/eligible, so waiting nodes would report contradictory metrics (`node_type=validator` on the node itself, absent from the network-wide validator view) and cause per-tick create/delete churn in the monitor under lock. That change needs the monitor side updated with it, so it was dropped here and will be a separate PR. This PR is a single behaviour change: the stale-cache fix.

## Why this approach (and not alternatives)

**Considered: fire `NotifyAll` on the epoch-start subscription handler after `LoadState`.** Rejected because it would trigger *all* epoch-start subscribers, not just the `PeerTypeProvider`. Notably, `nodesCoordinator.EpochStartAction` calls `saveState()` (writes to disk, prunes old configs) and `netStatistics` resets per-epoch byte counters to zero. A synthetic epoch event would cause real side effects in components that have nothing to do with this metrics bug.

**Considered: inject the epoch notifier into `storageBootstrapper` and fire `NotifyAll` from `applyBootInfos`.** Same blast radius problem, plus it requires changing `ArgsBaseStorageBootstrapper` (a widely-used struct) and threading a notifier through a low-level component that currently has no notion of epoch notifications.

**Chosen: direct, targeted refresh.** Only the `PeerTypeProvider` cache is rebuilt. No synthetic epoch events, no side effects on other components. The refresh call sits at the right lifecycle point in `StartConsensus` where both the restored validator data and the correct epoch are available.

Review also recorded two related pre-existing issues that this PR deliberately does not touch (both point at the same architectural follow-up: seed providers with the bootstrapped epoch instead of the pre-bootstrap one): the construction-time cache is still built from the pre-bootstrap epoch and serves reads until `StartConsensus`, and `validatorsProvider` has the same stale-epoch bug with no refresh. Worth its own issue; too big for this diff.

## Consensus safety

`PeerTypeProvider` is used exclusively by the heartbeat subsystem:
- `heartbeat/process/sender.go:191` (`ComputeForPubKey` for heartbeat message composition)
- `heartbeat/process/monitor.go:364` (`ComputeForPubKey` for peer type display/metrics)

It is not referenced anywhere in consensus group selection, block processing, or epoch transitions. This change has zero consensus impact.

## Thread safety

`UpdateCache` builds a complete new cache map outside the lock, then atomically swaps the map reference under `mutCache.Lock`. `ComputeForPubKey` reads under `mutCache.RLock`. This is the same `sync.RWMutex` pattern already in place. The constructor call runs before the sender goroutine starts, and the `StartConsensus` refresh runs before `StartSyncingBlocks`, so no epoch-start event can interleave with it. The heartbeat sender goroutine may read the stale cache for a few seconds during boot; this is a pre-existing condition that this change reduces from "remainder of epoch" to "seconds".

## Files changed

| File | Change |
|------|--------|
| `core/process/peer/peerTypeProvider.go` | Export `updateCache` -> `UpdateCache`; `createNewCache` returns `(map, ok)` and fails on any getter error; `UpdateCache` keeps the previous cache on failure; Warn logging with error values |
| `core/process/peer/peerTypeProvider_test.go` | Update `createNewCache`/`UpdateCache` callers; add keep-previous-on-failure and fail-fast-on-any-getter-error tests; `require.Len` before map dereferences |
| `common/mock/nodesCoordinatorMock.go` | Add epoch-aware hook variants (`...WithEpochCalled`) so tests can assert which epoch reaches the coordinator |
| `node/heartbeat/componentHandler/heartbeatHandler.go` | Add `RefreshPeerTypeCache` method |
| `node/heartbeat/componentHandler/heartbeatHandler_test.go` | Add 2 tests (refresh with epoch-forwarding assertion, nil-safety) |
| `node/interface.go` | Add `RefreshPeerTypeCache` to `HeartbeatHandler` interface |
| `node/node.go` | Call `RefreshPeerTypeCache` in `StartConsensus` after `LoadStorage`, guarded on a restored block header |

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./core/process/peer/... ./node/...` passes
- [x] `go test ./core/process/peer/...` passes (existing + new tests)
- [x] `go test ./node/heartbeat/componentHandler/...` passes (existing + new tests)
- [x] `go test ./node/...` passes
- [x] Three-agent audit (security, code review, adversarial full-diff), plus a second independent full-diff review after the review-feedback revision
- [ ] CI checks green
- [ ] Validated live on mainnet node (see #88 comment for reproduction procedure)

## Known gap

The `StartConsensus` wiring itself (call site, ordering after `LoadStorage`, nil-header guard) has no node-level test: exercising `StartConsensus` requires mocking 30+ `Node` dependencies. The behaviour on either side of the wiring is pinned by unit tests (`UpdateCache` semantics, keep-previous guard, epoch forwarding through `RefreshPeerTypeCache`). Accepted and recorded as a conscious trade-off.

### DIFF
--- a/core/process/peer/peerTypeProvider.go
+++ b/core/process/peer/peerTypeProvider.go
@@ -101,8 +101,14 @@ func (ptp *PeerTypeProvider) epochStartEventHandler() sharding.EpochStartActionH
 }
 
 // UpdateCache rebuilds the validator-type cache for the given epoch.
+// When the coordinator cannot provide validator lists (e.g. unknown epoch),
+// the previous cache is kept to avoid replacing valid data with an empty map.
 func (ptp *PeerTypeProvider) UpdateCache(epoch uint32) {
-	newCache := ptp.createNewCache(epoch)
+	newCache, ok := ptp.createNewCache(epoch)
+	if !ok {
+		log.Warn("peerTypeProvider - keeping previous cache, validator list unavailable", "epoch", epoch)
+		return
+	}
 
 	ptp.mutCache.Lock()
 	ptp.cache = newCache
@@ -111,28 +117,24 @@ func (ptp *PeerTypeProvider) UpdateCache(epoch uint32) {
 
 func (ptp *PeerTypeProvider) createNewCache(
 	epoch uint32,
-) map[string]*peerListAndShard {
+) (map[string]*peerListAndShard, bool) {
 	newCache := make(map[string]*peerListAndShard)
-
-	nodesMapWaiting, err := ptp.nodesCoordinator.GetAllWaitingValidatorsKeys(epoch, false)
-	if err != nil {
-		log.Debug("peerTypeProvider - GetAllWaitingValidatorsKeys failed", "epoch", epoch)
-	}
-	computePeerType(newCache, nodesMapWaiting, core.WaitingList)
 
 	nodesMapElected, err := ptp.nodesCoordinator.GetAllElectedValidatorsKeys(epoch, false)
 	if err != nil {
-		log.Debug("peerTypeProvider - GetAllElectedValidatorsKeys failed", "epoch", epoch)
+		log.Warn("peerTypeProvider - GetAllElectedValidatorsKeys failed", "epoch", epoch, "error", err)
+		return nil, false
 	}
 	computePeerType(newCache, nodesMapElected, core.ElectedList)
 
 	nodesMapEligible, err := ptp.nodesCoordinator.GetAllEligibleValidatorsKeys(epoch, false)
 	if err != nil {
-		log.Debug("peerTypeProvider - GetAllEligibleValidatorsKeys failed", "epoch", epoch)
+		log.Warn("peerTypeProvider - GetAllEligibleValidatorsKeys failed", "epoch", epoch, "error", err)
+		return nil, false
 	}
 	computePeerType(newCache, nodesMapEligible, core.EligibleList)
 
-	return newCache
+	return newCache, true
 }
 
 func computePeerType(
--- a/core/process/peer/peerTypeProvider.go
+++ b/core/process/peer/peerTypeProvider.go
@@ -47,7 +47,7 @@ func NewPeerTypeProvider(arg ArgPeerTypeProvider) (*PeerTypeProvider, error) {
 		mutCache:         sync.RWMutex{},
 	}
 
-	ptp.updateCache(arg.StartEpoch)
+	ptp.UpdateCache(arg.StartEpoch)
 
 	arg.EpochStartEventNotifier.RegisterHandler(ptp.epochStartEventHandler())
 
@@ -91,7 +91,7 @@ func (ptp *PeerTypeProvider) epochStartEventHandler() sharding.EpochStartActionH
 				"nonce", hdr.GetNonce(),
 				"slot", hdr.GetSlot(),
 				"epoch", hdr.GetEpoch())
-			ptp.updateCache(hdr.GetEpoch())
+			ptp.UpdateCache(hdr.GetEpoch())
 		},
 		func(_ data.HeaderHandler) {},
 		core.IndexerOrder,
@@ -100,7 +100,8 @@ func (ptp *PeerTypeProvider) epochStartEventHandler() sharding.EpochStartActionH
 	return subscribeHandler
 }
 
-func (ptp *PeerTypeProvider) updateCache(epoch uint32) {
+// UpdateCache rebuilds the validator-type cache for the given epoch.
+func (ptp *PeerTypeProvider) UpdateCache(epoch uint32) {
 	newCache := ptp.createNewCache(epoch)
 
 	ptp.mutCache.Lock()
@@ -112,6 +113,12 @@ func (ptp *PeerTypeProvider) createNewCache(
 	epoch uint32,
 ) map[string]*peerListAndShard {
 	newCache := make(map[string]*peerListAndShard)
+
+	nodesMapWaiting, err := ptp.nodesCoordinator.GetAllWaitingValidatorsKeys(epoch, false)
+	if err != nil {
+		log.Debug("peerTypeProvider - GetAllWaitingValidatorsKeys failed", "epoch", epoch)
+	}
+	computePeerType(newCache, nodesMapWaiting, core.WaitingList)
 
 	nodesMapElected, err := ptp.nodesCoordinator.GetAllElectedValidatorsKeys(epoch, false)
 	if err != nil {
--- a/core/process/peer/peerTypeProvider_test.go
+++ b/core/process/peer/peerTypeProvider_test.go
@@ -121,8 +121,9 @@ func TestNewPeerTypeProvider_createCache(t *testing.T) {
 		mutCache:         sync.RWMutex{},
 	}
 
-	cache := ptp.createNewCache(0)
+	cache, ok := ptp.createNewCache(0)
 
+	require.True(t, ok)
 	assert.NotNil(t, cache)
 
 	assert.NotNil(t, cache[pkElected])
@@ -275,8 +276,9 @@ func TestPeerTypeProvider_CreateNewCacheScenarios(t *testing.T) {
 	}
 	ptp, _ := NewPeerTypeProvider(arg)
 
-	cache := ptp.createNewCache(0)
-	assert.Len(t, cache, 3)
+	cache, ok := ptp.createNewCache(0)
+	require.True(t, ok)
+	require.Len(t, cache, 3)
 	assert.Equal(t, core.EligibleList, cache["elected1"].pType) // elected1 is also eligible as it have been updated in the eligible list
 	assert.Equal(t, core.ElectedList, cache["elected2"].pType)
 	assert.Equal(t, core.EligibleList, cache["eligible1"].pType)
@@ -319,14 +321,11 @@ func TestPeerTypeProvider_GetAllPeerTypeInfos(t *testing.T) {
 	assert.Empty(t, emptyPeerTypeInfos, "Should return empty slice for empty cache")
 }
 
-func TestPeerTypeProvider_CreateNewCache_IncludesWaitingList(t *testing.T) {
+func TestPeerTypeProvider_UpdateCache_KeepsPreviousOnFailure(t *testing.T) {
 	t.Parallel()
 
 	arg := createDefaultArgPeerTypeProvider()
 	arg.NodesCoordinator = &mock.NodesCoordinatorMock{
-		GetAllWaitingValidatorsKeysCalled: func() ([][]byte, error) {
-			return [][]byte{[]byte("waiting1")}, nil
-		},
 		GetAllElectedValidatorsKeysCalled: func() ([][]byte, error) {
 			return [][]byte{[]byte("elected1")}, nil
 		},
@@ -341,84 +340,32 @@ func TestPeerTypeProvider_CreateNewCache_IncludesWaitingList(t *testing.T) {
 		mutCache:         sync.RWMutex{},
 	}
 
-	cache := ptp.createNewCache(0)
+	ptp.UpdateCache(0)
+	require.Len(t, ptp.cache, 2)
 
-	assert.Len(t, cache, 3)
-	assert.Equal(t, core.WaitingList, cache["waiting1"].pType)
-	assert.Equal(t, core.ElectedList, cache["elected1"].pType)
-	assert.Equal(t, core.EligibleList, cache["eligible1"].pType)
-}
-
-func TestPeerTypeProvider_CreateNewCache_WaitingOverwrittenByElected(t *testing.T) {
-	t.Parallel()
-
-	pk := "overlapping_pk"
-
-	arg := createDefaultArgPeerTypeProvider()
-	arg.NodesCoordinator = &mock.NodesCoordinatorMock{
-		GetAllWaitingValidatorsKeysCalled: func() ([][]byte, error) {
-			return [][]byte{[]byte(pk)}, nil
-		},
+	ptp.nodesCoordinator = &mock.NodesCoordinatorMock{
 		GetAllElectedValidatorsKeysCalled: func() ([][]byte, error) {
-			return [][]byte{[]byte(pk)}, nil
+			return nil, fmt.Errorf("epoch config does not exist")
 		},
 	}
 
-	ptp := PeerTypeProvider{
-		nodesCoordinator: arg.NodesCoordinator,
-		cache:            nil,
-		mutCache:         sync.RWMutex{},
-	}
+	ptp.UpdateCache(99)
 
-	cache := ptp.createNewCache(0)
-
-	assert.Len(t, cache, 1)
-	assert.Equal(t, core.ElectedList, cache[pk].pType)
+	require.Len(t, ptp.cache, 2)
+	assert.Equal(t, core.ElectedList, ptp.cache["elected1"].pType)
+	assert.Equal(t, core.EligibleList, ptp.cache["eligible1"].pType)
 }
 
-func TestPeerTypeProvider_CreateNewCache_AllThreeOverlap(t *testing.T) {
-	t.Parallel()
-
-	pk := "overlapping_pk"
-
-	arg := createDefaultArgPeerTypeProvider()
-	arg.NodesCoordinator = &mock.NodesCoordinatorMock{
-		GetAllWaitingValidatorsKeysCalled: func() ([][]byte, error) {
-			return [][]byte{[]byte(pk)}, nil
-		},
-		GetAllElectedValidatorsKeysCalled: func() ([][]byte, error) {
-			return [][]byte{[]byte(pk)}, nil
-		},
-		GetAllEligibleValidatorsKeysCalled: func() ([][]byte, error) {
-			return [][]byte{[]byte(pk)}, nil
-		},
-	}
-
-	ptp := PeerTypeProvider{
-		nodesCoordinator: arg.NodesCoordinator,
-		cache:            nil,
-		mutCache:         sync.RWMutex{},
-	}
-
-	cache := ptp.createNewCache(0)
-
-	assert.Len(t, cache, 1)
-	assert.Equal(t, core.EligibleList, cache[pk].pType)
-}
-
-func TestPeerTypeProvider_CreateNewCache_WaitingErrorDoesNotAffectOtherLists(t *testing.T) {
+func TestPeerTypeProvider_CreateNewCache_FailsOnAnyGetterError(t *testing.T) {
 	t.Parallel()
 
 	arg := createDefaultArgPeerTypeProvider()
 	arg.NodesCoordinator = &mock.NodesCoordinatorMock{
-		GetAllWaitingValidatorsKeysCalled: func() ([][]byte, error) {
-			return nil, fmt.Errorf("waiting list unavailable")
-		},
 		GetAllElectedValidatorsKeysCalled: func() ([][]byte, error) {
 			return [][]byte{[]byte("elected1")}, nil
 		},
 		GetAllEligibleValidatorsKeysCalled: func() ([][]byte, error) {
-			return [][]byte{[]byte("eligible1")}, nil
+			return nil, fmt.Errorf("eligible list unavailable")
 		},
 	}
 
@@ -428,36 +375,8 @@ func TestPeerTypeProvider_CreateNewCache_WaitingErrorDoesNotAffectOtherLists(t *
 		mutCache:         sync.RWMutex{},
 	}
 
-	cache := ptp.createNewCache(0)
+	cache, ok := ptp.createNewCache(0)
 
-	assert.Len(t, cache, 2)
-	assert.Equal(t, core.ElectedList, cache["elected1"].pType)
-	assert.Equal(t, core.EligibleList, cache["eligible1"].pType)
-}
-
-func TestPeerTypeProvider_CreateNewCache_WaitingOverwrittenByEligible(t *testing.T) {
-	t.Parallel()
-
-	pk := "overlapping_pk"
-
-	arg := createDefaultArgPeerTypeProvider()
-	arg.NodesCoordinator = &mock.NodesCoordinatorMock{
-		GetAllWaitingValidatorsKeysCalled: func() ([][]byte, error) {
-			return [][]byte{[]byte(pk)}, nil
-		},
-		GetAllEligibleValidatorsKeysCalled: func() ([][]byte, error) {
-			return [][]byte{[]byte(pk)}, nil
-		},
-	}
-
-	ptp := PeerTypeProvider{
-		nodesCoordinator: arg.NodesCoordinator,
-		cache:            nil,
-		mutCache:         sync.RWMutex{},
-	}
-
-	cache := ptp.createNewCache(0)
-
-	assert.Len(t, cache, 1)
-	assert.Equal(t, core.EligibleList, cache[pk].pType)
+	assert.False(t, ok)
+	assert.Nil(t, cache)
 }
--- a/core/process/peer/peerTypeProvider_test.go
+++ b/core/process/peer/peerTypeProvider_test.go
@@ -84,7 +84,7 @@ func TestPeerTypeProvider_UpdateCache(t *testing.T) {
 		mutCache:         sync.RWMutex{},
 	}
 
-	ptp.updateCache(0)
+	ptp.UpdateCache(0)
 
 	assert.NotNil(t, ptp.cache)
 	assert.Equal(t, len(elected), len(ptp.cache))
@@ -316,4 +316,118 @@ func TestPeerTypeProvider_GetAllPeerTypeInfos(t *testing.T) {
 	ptp.cache = make(map[string]*peerListAndShard)
 	emptyPeerTypeInfos := ptp.GetAllPeerTypeInfos()
 	assert.Empty(t, emptyPeerTypeInfos, "Should return empty slice for empty cache")
+}
+
+func TestPeerTypeProvider_CreateNewCache_IncludesWaitingList(t *testing.T) {
+	t.Parallel()
+
+	arg := createDefaultArgPeerTypeProvider()
+	arg.NodesCoordinator = &mock.NodesCoordinatorMock{
+		GetAllWaitingValidatorsKeysCalled: func() ([][]byte, error) {
+			return [][]byte{[]byte("waiting1")}, nil
+		},
+		GetAllElectedValidatorsKeysCalled: func() ([][]byte, error) {
+			return [][]byte{[]byte("elected1")}, nil
+		},
+		GetAllEligibleValidatorsKeysCalled: func() ([][]byte, error) {
+			return [][]byte{[]byte("eligible1")}, nil
+		},
+	}
+
+	ptp := PeerTypeProvider{
+		nodesCoordinator: arg.NodesCoordinator,
+		cache:            nil,
+		mutCache:         sync.RWMutex{},
+	}
+
+	cache := ptp.createNewCache(0)
+
+	assert.Len(t, cache, 3)
+	assert.Equal(t, core.WaitingList, cache["waiting1"].pType)
+	assert.Equal(t, core.ElectedList, cache["elected1"].pType)
+	assert.Equal(t, core.EligibleList, cache["eligible1"].pType)
+}
+
+func TestPeerTypeProvider_CreateNewCache_WaitingOverwrittenByElected(t *testing.T) {
+	t.Parallel()
+
+	pk := "overlapping_pk"
+
+	arg := createDefaultArgPeerTypeProvider()
+	arg.NodesCoordinator = &mock.NodesCoordinatorMock{
+		GetAllWaitingValidatorsKeysCalled: func() ([][]byte, error) {
+			return [][]byte{[]byte(pk)}, nil
+		},
+		GetAllElectedValidatorsKeysCalled: func() ([][]byte, error) {
+			return [][]byte{[]byte(pk)}, nil
+		},
+	}
+
+	ptp := PeerTypeProvider{
+		nodesCoordinator: arg.NodesCoordinator,
+		cache:            nil,
+		mutCache:         sync.RWMutex{},
+	}
+
+	cache := ptp.createNewCache(0)
+
+	assert.Len(t, cache, 1)
+	assert.Equal(t, core.ElectedList, cache[pk].pType)
+}
+
+func TestPeerTypeProvider_CreateNewCache_AllThreeOverlap(t *testing.T) {
+	t.Parallel()
+
+	pk := "overlapping_pk"
+
+	arg := createDefaultArgPeerTypeProvider()
+	arg.NodesCoordinator = &mock.NodesCoordinatorMock{
+		GetAllWaitingValidatorsKeysCalled: func() ([][]byte, error) {
+			return [][]byte{[]byte(pk)}, nil
+		},
+		GetAllElectedValidatorsKeysCalled: func() ([][]byte, error) {
+			return [][]byte{[]byte(pk)}, nil
+		},
+		GetAllEligibleValidatorsKeysCalled: func() ([][]byte, error) {
+			return [][]byte{[]byte(pk)}, nil
+		},
+	}
+
+	ptp := PeerTypeProvider{
+		nodesCoordinator: arg.NodesCoordinator,
+		cache:            nil,
+		mutCache:         sync.RWMutex{},
+	}
+
+	cache := ptp.createNewCache(0)
+
+	assert.Len(t, cache, 1)
+	assert.Equal(t, core.EligibleList, cache[pk].pType)
+}
+
+func TestPeerTypeProvider_CreateNewCache_WaitingOverwrittenByEligible(t *testing.T) {
+	t.Parallel()
+
+	pk := "overlapping_pk"
+
+	arg := createDefaultArgPeerTypeProvider()
+	arg.NodesCoordinator = &mock.NodesCoordinatorMock{
+		GetAllWaitingValidatorsKeysCalled: func() ([][]byte, error) {
+			return [][]byte{[]byte(pk)}, nil
+		},
+		GetAllEligibleValidatorsKeysCalled: func() ([][]byte, error) {
+			return [][]byte{[]byte(pk)}, nil
+		},
+	}
+
+	ptp := PeerTypeProvider{
+		nodesCoordinator: arg.NodesCoordinator,
+		cache:            nil,
+		mutCache:         sync.RWMutex{},
+	}
+
+	cache := ptp.createNewCache(0)
+
+	assert.Len(t, cache, 1)
+	assert.Equal(t, core.EligibleList, cache[pk].pType)
 }
--- a/core/process/peer/peerTypeProvider_test.go
+++ b/core/process/peer/peerTypeProvider_test.go
@@ -1,6 +1,7 @@
 package peer
 
 import (
+	"fmt"
 	"sync"
 	"sync/atomic"
 	"testing"
@@ -403,6 +404,35 @@ func TestPeerTypeProvider_CreateNewCache_AllThreeOverlap(t *testing.T) {
 
 	assert.Len(t, cache, 1)
 	assert.Equal(t, core.EligibleList, cache[pk].pType)
+}
+
+func TestPeerTypeProvider_CreateNewCache_WaitingErrorDoesNotAffectOtherLists(t *testing.T) {
+	t.Parallel()
+
+	arg := createDefaultArgPeerTypeProvider()
+	arg.NodesCoordinator = &mock.NodesCoordinatorMock{
+		GetAllWaitingValidatorsKeysCalled: func() ([][]byte, error) {
+			return nil, fmt.Errorf("waiting list unavailable")
+		},
+		GetAllElectedValidatorsKeysCalled: func() ([][]byte, error) {
+			return [][]byte{[]byte("elected1")}, nil
+		},
+		GetAllEligibleValidatorsKeysCalled: func() ([][]byte, error) {
+			return [][]byte{[]byte("eligible1")}, nil
+		},
+	}
+
+	ptp := PeerTypeProvider{
+		nodesCoordinator: arg.NodesCoordinator,
+		cache:            nil,
+		mutCache:         sync.RWMutex{},
+	}
+
+	cache := ptp.createNewCache(0)
+
+	assert.Len(t, cache, 2)
+	assert.Equal(t, core.ElectedList, cache["elected1"].pType)
+	assert.Equal(t, core.EligibleList, cache["eligible1"].pType)
 }
 
 func TestPeerTypeProvider_CreateNewCache_WaitingOverwrittenByEligible(t *testing.T) {
--- a/node/heartbeat/componentHandler/heartbeatHandler_test.go
+++ b/node/heartbeat/componentHandler/heartbeatHandler_test.go
@@ -152,6 +152,9 @@ func TestHeartbeatHandler_RefreshPeerTypeCache(t *testing.T) {
 	arg := createMockArgument()
 	hbh, err := NewHeartbeatHandler(arg)
 	require.Nil(t, err)
+	t.Cleanup(func() {
+		require.NoError(t, hbh.Close())
+	})
 
 	peerType, _, _ := hbh.peerTypeProvider.ComputeForPubKey(newPK)
 	assert.Equal(t, core.ObserverList, peerType)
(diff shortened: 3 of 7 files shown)


core/process/peer/peerTypeProvider_test.go
```diff
@@ -317,3 +317,117 @@ func TestPeerTypeProvider_GetAllPeerTypeInfos(t *testing.T) {
...
+		mutCache:         sync.RWMutex{},
+	}
+
+	cache := ptp.createNewCache(0)
+
+	assert.Len(t, cache, 3)
+	assert.Equal(t, core.WaitingList, cache["waiting1"].pType)
+	assert.Equal(t, core.ElectedList, cache["elected1"].pType)
+	assert.Equal(t, core.EligibleList, cache["eligible1"].pType)
+}
```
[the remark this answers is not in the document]
MathijsBok: Fixed in 53bbece. Added TestPeerTypeProvider_CreateNewCache_WaitingErrorDoesNotAffectOtherLists: returns an error from GetAllWaitingValidatorsKeys while elected and eligible succeed, asserts both are still present in the cache.

node/heartbeat/componentHandler/heartbeatHandler_test.go
```diff
@@ -142,3 +143,42 @@ func TestNewHeartbeatHandler_ShouldWork(t *testing.T) {
...
+
+	hbh.RefreshPeerTypeCache(1)
+
+	peerType, _, _ = hbh.peerTypeProvider.ComputeForPubKey(newPK)
+	assert.Equal(t, core.ElectedList, peerType)
+
+	time.Sleep(time.Second)
+	_ = hbh.Close()
+	time.Sleep(time.Second)
+}
```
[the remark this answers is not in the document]
MathijsBok: Fixed in 53bbece. Replaced both time.Sleep calls with t.Cleanup(func() { require.NoError(t, hbh.Close()) }). Close() already waits on senderDone internally, so the sleeps were unnecessary.

node/node.go
```diff
@@ -332,6 +332,10 @@ func (n *Node) StartConsensus() error {
 		epoch = crtBlockHeader.GetEpoch()
 	}
 
+	if !check.IfNil(n.heartbeatHandler) {
+		n.heartbeatHandler.RefreshPeerTypeCache(epoch)
+	}
```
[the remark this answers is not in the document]
MathijsBok: Acknowledged. Adding a node-level StartConsensus integration test that verifies ordering (LoadStorage before RefreshPeerTypeCache) would require mocking 30+ Node dependencies (blockchain, consensus, p2p, storage, etc.). The ordering guarantee is covered by unit tests at lower levels: peerTypeProvider_test.go verifies UpdateCache rebuilds correctly, and heartbeatHandler_test.go verifies RefreshPeerTypeCache delegates to UpdateCache. A full node-level test is out of scope for this fix but noted for future consideration.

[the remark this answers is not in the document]
MathijsBok: You can open a follow-up issue

node/node.go
```diff
@@ -332,6 +332,10 @@ func (n *Node) StartConsensus() error {
 		epoch = crtBlockHeader.GetEpoch()
 	}
 
+	if !check.IfNil(n.heartbeatHandler) {
+		n.heartbeatHandler.RefreshPeerTypeCache(epoch)
```
fbsobreira, line 336: **In scope — this call can wipe a good cache.** Half one of the defect; the other half is at `peerTypeProvider.go:105`.

`epoch` here comes from `n.blkc.GetGenesisHeader().GetEpoch()` (line 329), overridden by the current block header only if one exists. When `LoadStorage()` fails — which is **explicitly non-critical** and Debug-logged only (`core/process/sync/metablock.go:97`) — `nodesCoordinator.LoadState` never runs and `GetCurrentBlockHeader()` is nil, so `epoch` stays **0**.

`RefreshPeerTypeCache(0)` → `UpdateCache(0)` → `createNewCache(0)`: the coordinator has no `nodesConfig[0]`, so all three getters return `ErrEpochNodesConfigDoesNotExist`, each swallowed at Debug, and an **empty** map is installed over the populated one.

I reproduced this against the real code with a mock coordinator failing all three getters:

```
PROBE before refresh: cacheLen=1  peerType=elected
PROBE after  refresh: cacheLen=0  peerType=observer
```

Live effect: the node broadcasts `peerType=observer` for itself, `/node/heartbeatstatus` returns no elected/eligible peers, and `klv_live_validator_nodes` reads 0 — until the next epoch-start block. **Pre-PR the construction-time cache survived this path**, so on this branch the fix is strictly worse than the bug.

Two things close it, and they compose:
1. Don't refresh with a fallback epoch — skip the call when `crtBlockHeader` is nil, or use the coordinator's own restored epoch rather than the block header's.
2. Add the keep-previous guard at `UpdateCache` (see the comment there) so no caller can wipe the cache by passing an epoch the coordinator doesn't know.

core/process/peer/peerTypeProvider.go
```diff
@@ -100,7 +100,8 @@ func (ptp *PeerTypeProvider) epochStartEventHandler() sharding.EpochStartActionH
 	return subscribeHandler
 }
 
-func (ptp *PeerTypeProvider) updateCache(epoch uint32) {
+// UpdateCache rebuilds the validator-type cache for the given epoch.
+func (ptp *PeerTypeProvider) UpdateCache(epoch uint32) {
 	newCache := ptp.createNewCache(epoch)
```
fbsobreira: **In scope — half two: `UpdateCache` installs whatever it gets, including nothing.**

`createNewCache` returns a plain map with no success signal; every getter failure is only `log.Debug`. So when all three fail together — which is exactly what happens for an epoch the coordinator has no config for — this installs an empty map over a populated cache, and `ComputeForPubKey` then returns `ObserverList` for every key.

Worth flagging directly: **this is the same defect PR #92 had to fix in this file**, and it was closed there by having `createNewCache` return an `ok` flag and `updateCache` keep the previous cache when `ok` is false. That guard is not in this diff, and this PR adds the first call site that can pass an unknown epoch — so it makes a latent weakness reachable.

Minimal fix, matching what worked on #92:

```go
newCache, ok := ptp.createNewCache(epoch)
if !ok {
    log.Warn("peerTypeProvider - no validator list available, keeping previous cache", "epoch", epoch)
    return
}
```

with `createNewCache` returning `false` when the lookups fail. Note #92 also found that `&&` across the three getters is wrong — treat **any** failure as failure, not all three, or you install a partial validator set and every missing validator reads as `observer`.

core/process/peer/peerTypeProvider.go
```diff
@@ -113,6 +114,12 @@ func (ptp *PeerTypeProvider) createNewCache(
 ) map[string]*peerListAndShard {
 	newCache := make(map[string]*peerListAndShard)
 
+	nodesMapWaiting, err := ptp.nodesCoordinator.GetAllWaitingValidatorsKeys(epoch, false)
+	if err != nil {
+		log.Debug("peerTypeProvider - GetAllWaitingValidatorsKeys failed", "epoch", epoch)
+	}
+	computePeerType(newCache, nodesMapWaiting, core.WaitingList)
```
fbsobreira: **In scope (or split out — see the review body).** Seeding `core.WaitingList` has two consequences that no caller was updated for.

**1. The two sides disagree about what a waiting node is.**
`Sender.updateMetrics` treats anything that is not `ObserverList` as a validator (`node/heartbeat/process/sender.go:179-187`), so a waiting node now publishes `klv_node_type=validator` / `klv_peer_type=waiting`. But `heartbeatMessageInfo.GetIsValidator` and `Monitor.shouldSkipValidator` still count only `ElectedList`/`EligibleList`, so the same node is excluded from `klv_live_validator_nodes` and hidden from the heartbeat status API once inactive past `HideInactiveValidatorIntervalInSec`. An operator sees `node_type=validator` on the node's own `/node/status` while it is absent from the network-wide validator view.

**2. Per-tick create/delete churn.**
`computeInactiveHeartbeatMessages` now materialises an entry for every waiting node from `GetAllPeerTypeInfos()`. `Monitor.Cleanup` then evaluates `shouldSkipValidator`, which shields only elected/eligible — so for a waiting node that has never broadcast (`isActive=false`, `timestamp=genesisTime`) it deletes the entry. The next tick recreates it. With `HeartbeatRefreshIntervalInSec=1` that is an endless allocate/insert/delete loop over the whole waiting set **while holding `mutHeartbeatMessages.Lock`**, which also blocks `addHeartbeatMessageToMap` for inbound heartbeats. And because `GetHeartbeats()` runs `Cleanup` first, waiting nodes appear in `/node/heartbeatstatus` only intermittently, with a multi-month `TotalDownTime` — nondeterministic across identical requests, which defeats the stated purpose of surfacing them.

Before this PR the invariant held: everything materialised from `GetAllPeerTypeInfos` was elected/eligible and therefore immune to `Cleanup`.

The change is defensible — waiting nodes *are* validators — but it needs the monitor side moved with it (`shouldSkipValidator` and `GetIsValidator` extended to `WaitingList`). My suggestion in the body is to split this into its own PR with those changes, so #110 stays a single-behaviour fix.

node/heartbeat/componentHandler/heartbeatHandler_test.go
```diff
@@ -142,3 +143,41 @@ func TestNewHeartbeatHandler_ShouldWork(t *testing.T) {
...
+
+	peerType, _, _ := hbh.peerTypeProvider.ComputeForPubKey(newPK)
+	assert.Equal(t, core.ObserverList, peerType)
+
+	ncMock := arg.NodesCoordinator.(*cMock.NodesCoordinatorMock)
+	ncMock.GetAllElectedValidatorsKeysCalled = func() ([][]byte, error) {
+		return [][]byte{newPK}, nil
+	}
+
+	hbh.RefreshPeerTypeCache(1)
```
fbsobreira: **In scope — the core of the fix is untested.**

`common/mock.NodesCoordinatorMock.GetAllElected/Eligible/WaitingValidatorsKeys(_ uint32, _ bool)` discards the epoch, and the stub hooks (`GetAllElectedValidatorsKeysCalled func() ([][]byte, error)`) take no arguments. So `hbh.RefreshPeerTypeCache(1)` here, and every `createNewCache(0)` assertion in `peerTypeProvider_test.go`, pass identically **no matter which epoch is forwarded**.

That means a regression that hardcodes epoch 0 — or forwards `arg.StartEpoch` again, reintroducing precisely the stale-cache bug this PR exists to fix — ships with a fully green suite.

The whole point of the change is *which epoch* reaches the coordinator, so that is the thing to assert. Give the mock hook the epoch (`GetAllElectedValidatorsKeysCalled func(epoch uint32, _ bool) ([][]byte, error)`), capture it, and assert it equals the value passed to `RefreshPeerTypeCache`.

core/process/peer/peerTypeProvider_test.go
```diff
@@ -317,3 +318,146 @@ func TestPeerTypeProvider_GetAllPeerTypeInfos(t *testing.T) {
...
+	ptp := PeerTypeProvider{
+		nodesCoordinator: arg.NodesCoordinator,
+		cache:            nil,
+		mutCache:         sync.RWMutex{},
+	}
+
+	cache := ptp.createNewCache(0)
+
+	assert.Len(t, cache, 3)
+	assert.Equal(t, core.WaitingList, cache["waiting1"].pType)
```
fbsobreira: **In scope — cheap, and it protects every other test in the package.**

`assert.Len(t, cache, 3)` on line 346 is `assert`, not `require`, so it records a failure and **continues**. Line 347 then does `cache["waiting1"].pType` — on a map miss that is a nil `*peerListAndShard`, so it panics with a nil-pointer dereference.

The panic aborts the whole `core/process/peer` test binary, so instead of a readable "expected 3 elements, got 2" CI reports an opaque runtime panic and the results of every other test in the package are lost.

Same pattern at lines 348, 349, 376, 406, 434, 435 and 462. Either switch the length check to `require.Len`, or look the entry up with the two-value form and assert `ok` before dereferencing.

core/process/peer/peerTypeProvider.go
```diff
@@ -113,6 +114,12 @@ func (ptp *PeerTypeProvider) createNewCache(
 ) map[string]*peerListAndShard {
 	newCache := make(map[string]*peerListAndShard)
 
+	nodesMapWaiting, err := ptp.nodesCoordinator.GetAllWaitingValidatorsKeys(epoch, false)
+	if err != nil {
+		log.Debug("peerTypeProvider - GetAllWaitingValidatorsKeys failed", "epoch", epoch)
```
fbsobreira: **In scope — trivial, and this is the line that would explain the failure above.**

This drops the error value and logs below the default level:

```go
log.Debug("peerTypeProvider - GetAllWaitingValidatorsKeys failed", "epoch", epoch)
```

When the coordinator has no config for the epoch — the exact case the new `StartConsensus` refresh can hit — the operator sees every validator reported as `observer` with **no log line at all** at default verbosity, and no way to distinguish "epoch config missing" from "waiting list genuinely empty" without restarting at Debug.

Add `"error", err` at minimum. Given the consequence is a silently degraded cache, `log.Warn` is more appropriate than `Debug` here. (The elected/eligible lines below have the same shape but are pre-existing — worth the same treatment, but only this one is new in the diff.)

core/process/peer/peerTypeProvider.go
```diff
@@ -47,7 +47,7 @@ func NewPeerTypeProvider(arg ArgPeerTypeProvider) (*PeerTypeProvider, error) {
 		mutCache:         sync.RWMutex{},
 	}
```
fbsobreira, line 49: **Out of scope — recorded for a follow-up, not a merge condition.**

The cache is now built twice at startup, and the first build is the known-wrong one that serves every read until `StartConsensus`.

`NewPeerTypeProvider` builds it with the pre-bootstrap epoch (three coordinator queries walking the full elected+eligible+waiting sets under `mutNodesConfig`/`mutNodesMaps`), then `StartConsensus` discards that and repeats all three. On a mid-epoch restart the first build is guaranteed garbage — yet `go hbh.startSendingHeartbeats` and the peer-type metric begin reading it the moment `StartHeartbeat` returns (`cmd/node/node.go:193`), so validators broadcast `observer` for the whole interval between node creation and `StartConsensus`.

The root cause is the seed at `heartbeatHandler.go:104` (`StartEpoch: arg.EpochStartTrigger.Epoch()`) being read before `LoadStorage()` restores state. Constructing the provider after bootstrap — or seeding `ArgPeerTypeProvider.StartEpoch` with the bootstrapped epoch already available at `cmd/node/node.go:337` — does the work once and never publishes a wrong cache.

Too big for this diff; noting it so the push-hook doesn't become the permanent shape.

node/node.go
```diff
@@ -332,6 +332,10 @@ func (n *Node) StartConsensus() error {
 		epoch = crtBlockHeader.GetEpoch()
 	}
 
+	if !check.IfNil(n.heartbeatHandler) {
```
fbsobreira: **Out of scope — recorded, not a merge condition.** The sibling has the identical bug and gets no refresh.

`validatorsProvider` (same package as `PeerTypeProvider`) is seeded from the same pre-bootstrap epoch — `currentEpoch: args.StartEpoch` at `validatorsProvider.go:84`, fed by `args.startEpochNum` at `factory/process.go:336` — and is not refreshed here. After a mid-epoch restart `/validator/statistics` keeps overlaying elected/eligible from the wrong epoch (`validatorsProvider.go:220-228`) until the next epoch boundary.

Flagging it because the shape of this fix decides the pattern: with a push hook, every future consumer of a coordinator-derived cache needs its own refresh method threaded through `node.HeartbeatHandler`. Fixing the epoch seed instead (see the comment on `peerTypeProvider.go:49`) fixes both consumers at once and needs no new interface method.

core/process/peer/peerTypeProvider_test.go
```diff
@@ -317,3 +318,146 @@ func TestPeerTypeProvider_GetAllPeerTypeInfos(t *testing.T) {
...
+	assert.Equal(t, core.ElectedList, cache["elected1"].pType)
+	assert.Equal(t, core.EligibleList, cache["eligible1"].pType)
+}
+
+func TestPeerTypeProvider_CreateNewCache_WaitingOverwrittenByElected(t *testing.T) {
```
fbsobreira: Optional — these three pin behaviour production cannot produce.

`sharding.computeNodesConfigFromList` (`sharding/nodesCoordinator.go:907-921`) assigns each validator to exactly one of `electedList`/`eligibleList`/`waitingList` via a `switch` on `validatorInfo.List`, and `SetNodes` preserves that partition. **A key can never appear in two of the three sets.**

So `WaitingOverwrittenByElected` (:352), `AllThreeOverlap` (:379) and `WaitingOverwrittenByEligible` (:438) are ~85 lines pinning the incidental statement order of `createNewCache` as if it were a contract. A maintainer who legitimately reorders the passes — say to make waiting win, which is what the names suggest the semantics should be — gets three red tests describing states no node can reach, while no test catches an actual regression.

Worth collapsing into one that documents the partition invariant, or dropping.

core/process/peer/peerTypeProvider_test.go
```diff
@@ -317,3 +318,146 @@ func TestPeerTypeProvider_GetAllPeerTypeInfos(t *testing.T) {
 	emptyPeerTypeInfos := ptp.GetAllPeerTypeInfos()
 	assert.Empty(t, emptyPeerTypeInfos, "Should return empty slice for empty cache")
 }
+
+func TestPeerTypeProvider_CreateNewCache_IncludesWaitingList(t *testing.T) {
```
fbsobreira: Optional — the five new tests are the same 25-line skeleton five times.

Lines 322-464 differ only in which mock closures are set and which assertions run: build arg, override `NodesCoordinator`, construct a `PeerTypeProvider` literal, call `createNewCache(0)`, assert length and per-key `pType`. A table of `{name string; waiting, elected, eligible [][]byte; want map[string]core.PeerType}` covers all five in roughly 40 lines instead of 145.

The copy-paste already produced the three redundant precedence cases noted above, and any change to the `PeerTypeProvider` literal (a new field, a required constructor) now has to be applied five times.

fbsobreira (changes requested):
## Review at `53bbecef`

The stale-cache bug this fixes is real and worth fixing. But the remedy — a push refresh from `StartConsensus` plus an unguarded cache swap — can make the symptom **worse** than the bug on one path, and I reproduced that against the real code rather than reasoning about it.

Same split as #108: **does this diff introduce or worsen it?**

### In scope — fix before merge

| Where | What |
|---|---|
| `node/node.go:336` + `peerTypeProvider.go:105` | The new refresh can **wipe a populated cache to empty**, making a real validator report `observer` for up to a full epoch — the exact outage this PR fixes. Two halves of one defect. |
| `peerTypeProvider.go:121` | Waiting-list seeding makes a node self-report `validator` while the monitor still counts only elected/eligible, and drives a per-tick create/delete churn under `mutHeartbeatMessages.Lock` |
| `heartbeatHandler_test.go:167` | No test verifies the epoch actually reaches the coordinator — a regression hardcoding epoch 0 ships green |
| `peerTypeProvider_test.go:347` | Map-miss dereference panics instead of failing, taking the whole package's results with it |
| `peerTypeProvider.go:119` | New log line drops the error value at Debug, making the silent degradation undiagnosable |

### Scope suggestion — split the waiting-list change out

The waiting-list addition is an independent behaviour change bundled with a cache-refresh fix. It has its own consequences (contradictory metrics, monitor churn) that need monitor-side changes to resolve properly. Landing it here means this PR can't be evaluated on the stale-cache fix alone.

**Suggest dropping it from #110** and doing it as its own PR alongside the `shouldSkipValidator` / `GetIsValidator` changes it needs. That keeps this PR to one behaviour, which is what made #108 easy to review and merge. If you'd rather keep it, then the two findings on `:121` become merge conditions.

### Explicitly out of scope — for a separate PR

Real, but pre-existing and untouched here. Recorded, **not** merge conditions:

- **The construction-time cache is still built from the pre-bootstrap epoch** (`heartbeatHandler.go:104` seeding `StartEpoch: arg.EpochStartTrigger.Epoch()` while `StartHeartbeat` runs before `LoadStorage()`). This PR adds a second build rather than fixing the seed, so the known-wrong first build still serves every read until `StartConsensus`.
- **`validatorsProvider` has the identical bug and gets no refresh** — after a mid-epoch restart `/validator/statistics` keeps overlaying the wrong epoch until the next boundary.

Both point at the same architectural fix (construct after bootstrap, or seed with the bootstrapped epoch) which would remove the need for a bespoke push hook per consumer. Worth its own issue; too big for this diff.

### Note from PR #92

`UpdateCache` installing unconditionally is the **same defect #92 had to fix in this exact file**, closed there with a keep-previous-on-failure guard. That guard is not in this diff, and this PR adds the first call site that can reach the failing path. Details inline.

### Commits

`4047c2a7` and `53bbecef` both carry `Co-Authored-By: Claude Opus 4.6`. Per repo `CLAUDE.md` ("Never add co-author unless explicitly asked") these need stripping **before** merge — it's an amend now, a history rewrite after.

Requesting changes on the in-scope items.

node/node.go
MathijsBok: Fixed in 874c125, both halves together with the guard on the UpdateCache side:

1. The refresh call is now guarded with `!check.IfNil(crtBlockHeader)` (IfNil rather than `!= nil`, so a typed nil cannot diverge from the epoch fallback check four lines up). When LoadStorage cannot restore a header, the refresh is skipped entirely and the construction-time cache stays as it was, which matches pre-PR behaviour on that path.
2. UpdateCache now keeps the previous cache when createNewCache fails (see the reply on the other half).

Your reproduced scenario (all three getters failing on epoch 0) now leaves the populated cache untouched and logs a Warn instead of installing an empty map.

core/process/peer/peerTypeProvider.go
MathijsBok: Fixed in 874c125, matching the shape from PR #92: createNewCache now returns `(map[string]*peerListAndShard, bool)` and fail-fasts on the FIRST getter error (any failure is failure, not all three, so a partial validator set can never be installed). UpdateCache keeps the previous cache and logs

```go
log.Warn("peerTypeProvider - keeping previous cache, validator list unavailable", "epoch", epoch)
```

when the build fails. Two new tests pin this: `TestPeerTypeProvider_UpdateCache_KeepsPreviousOnFailure` (populated cache survives a refresh with a failing coordinator) and `TestPeerTypeProvider_CreateNewCache_FailsOnAnyGetterError` (a single getter error fails the whole build).

core/process/peer/peerTypeProvider.go
MathijsBok: Split out as suggested. 874c125 removes the waiting-list seeding from createNewCache, so this PR is back to a single behaviour change (the stale-cache fix) and the pre-PR invariant holds again: everything materialised from GetAllPeerTypeInfos is elected/eligible and immune to Cleanup. The waiting-list change will come as its own PR together with the monitor-side updates it needs (shouldSkipValidator and GetIsValidator extended to WaitingList). The PR description documents the split and the reasons.

node/heartbeat/componentHandler/heartbeatHandler_test.go
MathijsBok: Fixed in 874c125. NodesCoordinatorMock gained epoch-aware hook variants (`GetAllElectedValidatorsKeysWithEpochCalled func(uint32, bool)` etc., taking precedence over the legacy argless hooks when set), and the test now captures the epoch and asserts it equals the value passed to RefreshPeerTypeCache:

```go
ncMock.GetAllElectedValidatorsKeysWithEpochCalled = func(epoch uint32, _ bool) ([][]byte, error) {
    capturedEpoch = epoch
    return [][]byte{newPK}, nil
}
hbh.RefreshPeerTypeCache(targetEpoch)
assert.Equal(t, targetEpoch, capturedEpoch)
```

A regression that hardcodes epoch 0 or forwards StartEpoch again now fails this test.

core/process/peer/peerTypeProvider_test.go
MathijsBok: Fixed in 874c125. The five waiting-list tests (including all the flagged lines) were removed with the waiting-list revert. The two tests that replaced them, plus the pre-existing ones I touched (createCache, CreateNewCacheScenarios), use `require.True(t, ok)` / `require.Len` before any map dereference, so a length mismatch now reports a readable failure instead of a nil-pointer panic aborting the package.

core/process/peer/peerTypeProvider.go
MathijsBok: Fixed in 874c125. The waiting-list line itself is gone with the revert, and since the elected/eligible getter failures are now the trigger for the keep-previous path (no longer a silent partial build), both were upgraded to Warn with the error value included:

```go
log.Warn("peerTypeProvider - GetAllElectedValidatorsKeys failed", "epoch", epoch, "error", err)
```

plus the keep-previous Warn in UpdateCache, so a degraded refresh is visible at default verbosity with the reason attached.

core/process/peer/peerTypeProvider.go
MathijsBok: Acknowledged, agreed this is the real architectural fix (seed with the bootstrapped epoch, or construct after bootstrap, fixing every consumer at once instead of a push hook per consumer). Recorded in the PR description under the follow-up note together with the validatorsProvider sibling. Will open a separate issue for it so the push hook does not become the permanent shape.

node/node.go
MathijsBok: Acknowledged. validatorsProvider is indeed the same bug with no refresh, and it strengthens the case for the epoch-seed fix over per-consumer push hooks. Recorded in the PR description; will be covered by the same follow-up issue as the construction-time seed.

core/process/peer/peerTypeProvider_test.go
MathijsBok: Resolved by the waiting-list revert in 874c125: all three precedence tests (WaitingOverwrittenByElected, AllThreeOverlap, WaitingOverwrittenByEligible) are gone. Agreed on the underlying point that the partition invariant from computeNodesConfigFromList makes overlap states unreachable in production; when the waiting-list PR lands, precedence pinning will not return in this form.

core/process/peer/peerTypeProvider_test.go
MathijsBok: Moot after 874c125: the five-test skeleton is gone with the waiting-list revert. What remains is two targeted tests with distinct shapes (keep-previous asserts on the provider after two UpdateCache calls, fail-fast asserts on the (map, ok) contract), so a table would not buy anything there. Noted for the follow-up waiting-list PR, where a scenario table is the right structure.

MathijsBok:
@fbsobreira thanks for the depth of this review, in particular for reproducing the cache-wipe path against the real code instead of reasoning about it. That reproduction changed the fix: it showed the branch was strictly worse than the bug on the LoadStorage-failure path, which none of the earlier review rounds had caught. Everything below is in `874c125` (force-pushed; history rewritten, see the commits section at the end).

## In-scope findings, one by one

### 1. Refresh could wipe a populated cache (node.go:336 + peerTypeProvider.go:105)

Your reproduction: LoadStorage fails silently (it is explicitly non-critical), `GetCurrentBlockHeader()` returns nil, epoch stays at the genesis fallback, all three coordinator getters return `ErrEpochNodesConfigDoesNotExist`, and the old code installed an empty map over the populated construction-time cache. Result: `cacheLen=1/elected` before the refresh, `cacheLen=0/observer` after.

Closed from both sides, composing exactly as you suggested:

- **Call-site guard** (`node/node.go:335`): the refresh now runs only when `!check.IfNil(crtBlockHeader)`. When LoadStorage cannot restore a header there is nothing better than the construction-time cache, so we keep it, which is pre-PR behaviour on that path. I used `check.IfNil` rather than `!= nil` deliberately: line 331 uses `IfNil` for the epoch fallback, and if any `ChainHandler` implementation ever returned a typed nil, `!= nil` on line 335 would diverge from it in the worst direction (epoch falls back to genesis while the refresh still runs). Same predicate on both lines removes that failure mode.
- **Provider guard** (`core/process/peer/peerTypeProvider.go`): `createNewCache` now returns `(map[string]*peerListAndShard, bool)` and fail-fasts on the FIRST getter error. Any failure is failure, per your note from PR #92 about `&&` across getters: a partial build can never be installed, so no validator can silently degrade to observer because one list was unavailable. `UpdateCache` keeps the previous cache and logs `Warn("peerTypeProvider - keeping previous cache, validator list unavailable", "epoch", epoch)` when the build fails. This restores the keep-previous shape #92 introduced, and it protects every call site (constructor, epoch-start handler, and the new StartConsensus path), not just the new one.

Tests pinning this: `TestPeerTypeProvider_UpdateCache_KeepsPreviousOnFailure` re-runs your scenario (populated cache, then a refresh against a coordinator that fails) and asserts the cache still holds the original two entries with their original types. `TestPeerTypeProvider_CreateNewCache_FailsOnAnyGetterError` pins the fail-fast contract with only the second getter failing, so the any-not-all semantics are what is asserted.

### 2. Waiting-list seeding (peerTypeProvider.go:121)

Dropped from this PR entirely, taking your split suggestion. Your analysis was correct on both counts and I did not find a cheap way around either: the sender treats everything non-observer as validator while `shouldSkipValidator`/`GetIsValidator` count only elected/eligible (contradictory metrics), and `Cleanup` deletes never-broadcast waiting entries every tick while `computeInactiveHeartbeatMessages` recreates them (churn under `mutHeartbeatMessages.Lock`, intermittent `/node/heartbeatstatus` output). Doing it properly means moving the monitor with it, which is its own change with its own review surface.

This restores the pre-PR invariant you named: everything materialised from `GetAllPeerTypeInfos` is elected/eligible and therefore immune to `Cleanup`. The PR is now a single behaviour change. Tracking issue for the reintroduction with the monitor-side scope: #114. The five waiting-list tests went with it, which also resolves your two optional findings (the three unreachable-overlap precedence tests and the five-times-copied skeleton) by removal rather than by refactor.

### 3. Epoch forwarding untested (heartbeatHandler_test.go:167)

You were right that every assertion passed identically no matter which epoch was forwarded, because the mock discarded it. `NodesCoordinatorMock` now has epoch-aware hook variants alongside the legacy argless ones (`GetAllElectedValidatorsKeysWithEpochCalled func(uint32, bool) ([][]byte, error)` etc.; the epoch-aware hook wins when both are set, and no existing test sets both). The test now does:

```go
ncMock.GetAllElectedValidatorsKeysWithEpochCalled = func(epoch uint32, _ bool) ([][]byte, error) {
    capturedEpoch = epoch
    return [][]byte{newPK}, nil
}
hbh.RefreshPeerTypeCache(targetEpoch)   // targetEpoch = 7, deliberately not 0 or 1
assert.Equal(t, targetEpoch, capturedEpoch)
```

A regression that hardcodes epoch 0 or reintroduces `arg.StartEpoch` now fails this test instead of shipping green.

### 4. assert.Len + dereference panics (peerTypeProvider_test.go:347)

All flagged lines are gone with the waiting-list tests. The two replacement tests and the pre-existing tests this PR touches (`createCache`, `CreateNewCacheScenarios`) use `require.True(t, ok)` / `require.Len` before any map access, so a mismatch reports as a readable failure instead of a nil-pointer panic taking the package's results with it.

### 5. Error dropped at Debug (peerTypeProvider.go:119)

The new waiting-list line is gone with the revert. Since the elected/eligible getter failures are now the trigger for the keep-previous path rather than a silently tolerated partial build, both were upgraded to `Warn` with the error value attached, plus the keep-previous `Warn` in `UpdateCache`. A degraded refresh is now visible at default verbosity with the reason, and "epoch config missing" is distinguishable from "list genuinely empty" without restarting at Debug.

## Out-of-scope findings

Both recorded and now tracked in #113 (construction-time pre-bootstrap epoch seed, covering both `PeerTypeProvider` and the `validatorsProvider` sibling you flagged). I agree the epoch-seed fix is the real shape: it fixes every consumer at once, needs no interface methods, and makes the `RefreshPeerTypeCache` push hook from this PR removable. The issue says exactly that, so the hook does not become permanent by inertia.

## Commits

Branch history rewritten: the `Co-Authored-By` lines are stripped from both earlier commits and the new commit has none, per repo convention. The tree before/after the rewrite is identical (verified with an empty `git diff` between the old and new heads); only messages changed.

## Verification on 874c125

- `go build ./...`, `go vet` clean on the touched packages.
- `go test` green on `core/process/peer`, `node/heartbeat/componentHandler`, and `node` (the packages this PR touches).
- An independent adversarial full-diff review pass after the fixes confirmed: the guard combination closes the wipe path (the coordinator returns `ErrEpochNodesConfigDoesNotExist` for unknown epochs, so fail-fast triggers correctly, while empty-but-valid lists still legitimately replace the cache), no new races (constructor call runs before handler registration and before the sender goroutine starts; the StartConsensus refresh runs before `StartSyncingBlocks`, so no epoch-start event can interleave), and the new tests fail under the regressions they claim to pin. It also caught a gofmt regression in the mock and the `IfNil` inconsistency, both fixed in the same commit.
- Known limitation, stated rather than silent: the StartConsensus wiring itself (call site, ordering, nil-header guard) has no node-level test; exercising `StartConsensus` needs 30+ mocked dependencies. The behaviour on both sides of the wiring is pinned by unit tests. Recorded in the PR description as a conscious trade-off.
- Pre-existing, not from this PR: `go test -race ./core/process/peer/...` fails on develop too (data races in `validatorsProvider`'s refresh goroutine), and the monitor edge-case tests have timing flakiness. Verified both against a clean develop baseline.
- CI: lint and tests pass on `874c125`; CodeRabbit round complete (its one comment about restoring the waiting list is answered as the deliberate split, with #114 linked). The sonarqube and notify jobs fail for infra reasons on fork PRs (Twingate cannot connect / empty webhook secrets), same as every previous push on this PR and unrelated to the diff.

core/process/peer/peerTypeProvider.go
```diff
@@ -110,22 +117,24 @@ func (ptp *PeerTypeProvider) updateCache(epoch uint32) {
...
 	nodesMapEligible, err := ptp.nodesCoordinator.GetAllEligibleValidatorsKeys(epoch, false)
 	if err != nil {
-		log.Debug("peerTypeProvider - GetAllEligibleValidatorsKeys failed", "epoch", epoch)
+		log.Warn("peerTypeProvider - GetAllEligibleValidatorsKeys failed", "epoch", epoch, "error", err)
+		return nil, false
 	}
 	computePeerType(newCache, nodesMapEligible, core.EligibleList)
 
-	return newCache
+	return newCache, true
```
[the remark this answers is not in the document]
MathijsBok: Intentional, not a regression. The waiting-list seeding was deliberately removed in 874c125 following the human review on this PR (see the review at 53bbece and the discussion on the earlier createNewCache comment): seeding WaitingList without updating Monitor.shouldSkipValidator and heartbeatMessageInfo.GetIsValidator produces contradictory metrics (node reports validator while excluded from the network-wide validator view) and per-tick create/delete churn in the monitor under lock. It will be reintroduced in a dedicated PR together with those monitor-side changes. The PR description has been updated accordingly, so "PR objectives" no longer include waiting-list support.

core/process/peer/peerTypeProvider.go
MathijsBok: Opened #113 to track this (covers both this seed and the validatorsProvider sibling).

node/node.go
MathijsBok: Opened #113 to track this together with the PeerTypeProvider construction-time seed.

core/process/peer/peerTypeProvider.go
MathijsBok: Opened #114 to track the waiting-list change with the monitor-side scope (shouldSkipValidator, GetIsValidator, metrics decision, churn-free tests).

core/process/peer/peerTypeProvider.go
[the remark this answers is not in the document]
MathijsBok: Tracking issue for the waiting-list reintroduction: #114.

fbsobreira (approved):

Issue #114, opened 2026-08-09
Report waiting-list validators in heartbeat peer-type metrics (needs monitor-side changes)


MathijsBok:
## Context

Split out of PR #110 following review. An earlier revision of that PR seeded the `PeerTypeProvider` cache with the waiting list, so waiting validators would stop reporting as `observer` in `klv_node_type`/`klv_peer_type`. Review showed the change is incomplete without monitor-side updates, so it was dropped there and tracked here.

## Problem

`PeerTypeProvider.createNewCache` only queries `GetAllElectedValidatorsKeys` and `GetAllEligibleValidatorsKeys`. A validator in the waiting list is absent from the cache, so `ComputeForPubKey` returns `core.ObserverList` and the node reports `observer` until it is promoted.

## Why just adding the waiting list is not enough

Seeding `core.WaitingList` in the cache without updating the consumers creates two problems (details in the PR #110 review thread on `peerTypeProvider.go:121`):

1. **Contradictory metrics.** `Sender.updateMetrics` treats anything that is not `ObserverList` as a validator (`node/heartbeat/process/sender.go:179-187`), so a waiting node would publish `klv_node_type=validator`. But `heartbeatMessageInfo.GetIsValidator` and `Monitor.shouldSkipValidator` only count `ElectedList`/`EligibleList`, so the same node stays excluded from `klv_live_validator_nodes` and is hidden from the heartbeat status API once inactive past `HideInactiveValidatorIntervalInSec`. An operator sees `validator` on the node's own `/node/status` while it is absent from the network-wide validator view.

2. **Per-tick churn in the monitor.** `computeInactiveHeartbeatMessages` materialises an entry for every key from `GetAllPeerTypeInfos()`. `Monitor.Cleanup` shields only elected/eligible via `shouldSkipValidator`, so an entry for a waiting node that never broadcast gets deleted every tick and recreated the next one. With `HeartbeatRefreshIntervalInSec=1` that is an endless allocate/insert/delete loop over the whole waiting set while holding `mutHeartbeatMessages.Lock`, which also blocks inbound heartbeat processing. Waiting nodes would appear in `/node/heartbeatstatus` only intermittently, defeating the purpose.

## Scope of the fix

Move all pieces together in one PR:

- Seed the waiting list in `createNewCache` before elected and eligible (precedence waiting < elected < eligible).
[...]

Issue #113, opened 2026-08-09
Seed PeerTypeProvider and validatorsProvider with the bootstrapped epoch instead of the pre-bootstrap epoch


MathijsBok:
## Context

Follow-up from the review on PR #110 (fix for #88). That PR adds a targeted cache refresh for `PeerTypeProvider` in `Node.StartConsensus`, but the review identified the underlying architectural problem, which is out of scope for that diff and recorded here so the push hook does not become the permanent shape.

## Problem

Two providers in `core/process/peer` are seeded with the pre-bootstrap epoch, read before `LoadStorage()` restores state:

1. **`PeerTypeProvider`**: `NewPeerTypeProvider` builds its cache with `StartEpoch: arg.EpochStartTrigger.Epoch()` (`node/heartbeat/componentHandler/heartbeatHandler.go:104`), during `StartHeartbeat`, which runs before `LoadStorage()`. On a mid-epoch restart this first build is guaranteed wrong (genesis-era keys), yet the heartbeat sender goroutine and the peer-type metric start reading it the moment `StartHeartbeat` returns (`cmd/node/node.go:193`). PR #110 adds a second build in `StartConsensus`, so the wrong first build now serves every read only for the window between node creation and `StartConsensus`, but it is still built, still wrong, and the work is done twice (three coordinator queries walking the full validator sets under `mutNodesConfig`/`mutNodesMaps`, twice).

2. **`validatorsProvider`**: has the identical bug and gets no refresh at all. It is seeded from the same pre-bootstrap epoch (`currentEpoch: args.StartEpoch` at `validatorsProvider.go:84`, fed by `args.startEpochNum` at `factory/process.go:336`). After a mid-epoch restart, `/validator/statistics` keeps overlaying elected/eligible from the wrong epoch (`validatorsProvider.go:220-228`) until the next epoch boundary.

## Why the current shape does not scale

With a push hook (PR #110's `RefreshPeerTypeCache`), every future consumer of a coordinator-derived cache needs its own refresh method threaded through `node.HeartbeatHandler` or a similar interface. Each new consumer re-creates the bug until someone notices.

## Proposed direction

Fix the epoch seed instead: either construct these providers after storage bootstrap, or seed them with the bootstrapped epoch that is already available at `cmd/node/node.go:337`. That does the work once, never publishes a wrong cache, fixes both consumers at once, and needs no new interface methods. Once in place, the `RefreshPeerTypeCache` push hook from PR #110 can be removed.

## References

[...]

Issue #106, opened 2026-08-05
Flaky tests make the CI test check unreliable (heartbeat monitor uptime, consensus revert integration)


MathijsBok:
## Summary

Two tests fail intermittently without any code change, which makes the `test` CI check unreliable. Because external contributors cannot rerun a failed workflow (`gh run rerun` returns `Must have admin rights to Repository`), a red `test` check on a fork PR currently has to be re-rolled by pushing a new commit, or interpreted by hand.

Both were observed while working on #102, #103, #104 and #105, on branches that cannot influence the failing packages. Measurements below were taken on a clean checkout of `develop` at `d85f4b4`, with no local modifications.

## 1. `TestMonitor_ObserverGapValidatorPartlyOnline5` — reproduces locally, ~30-40% failure rate

Package: `node/heartbeat/process`

```
go test -count=1 -run "TestMonitor_ObserverGapValidatorPartlyOnline5" ./node/heartbeat/process/
```

Failure rate on clean `develop` across separate invocations: **4/10** in one batch, **3/12** in another.

Note the flake only appears across *separate process invocations*. `go test -count=15` in a single process passes consistently, which points at start-up timing rather than shared state between iterations.

Failing assertions (`monitorEdgeCases_test.go:580-581`, via `verifyHeartBeat`):

```
monitorEdgeCases_test.go:580: Not equal: expected: 90    (TotalUpTime)
monitorEdgeCases_test.go:581: Not equal: expected: 10    (TotalDownTime)
```

The expected values vary per run (90, 140, 190 observed for uptime), so the accumulated up/down time drifts by a whole step. The test drives time through a mock timer (`timer.SetSeconds(...)`), so a real wall-clock read somewhere in the uptime accounting of `RefreshHeartbeatMessageInfo` would explain the drift: if the computation mixes mocked and real time, the boundary between "still up" and "went down" lands differently depending on where the process happens to start relative to a real-time second boundary.

CI occurrence: https://github.com/klever-io/klever-go/actions/runs/31029372425/job/92386116852 (on #103, which only touches `core/consensus/slot/**` and cannot reach this package).

## 2. `TestConsensus_RevertBlockAndTransactions` — CI only, does not reproduce locally

Package: `integrationTest/consensus`

Passed 6/6 locally on clean `develop` and 3/3 on the PR branches, so this one appears to be load- or timing-sensitive in the CI environment specifically.

CI occurrences, on three different PRs whose changes do not touch this path:

[...]

PR #115, opened 2026-08-09
fix: report waiting-list validators in heartbeat peer-type metrics

MathijsBok:
## Summary

Implements #114: waiting-list validators no longer report as `observer` in the heartbeat peer-type metrics. This is the change that was deliberately split out of PR #110 because seeding the cache alone, without the monitor-side updates, produces contradictory metrics and per-tick churn (see the #110 review thread). This PR moves all pieces together.

Behavior for a waiting-list validator, before and after:

| Surface | Before | After |
|---|---|---|
| `klv_node_type` (own node) | `observer` | `validator` |
| `klv_peer_type` (own node) | `observer` | `waiting` |
| `klv_live_validator_nodes` (monitor) | not counted | counted while actively heartbeating |
| `/node/heartbeatstatus` | hidden after `HideInactiveValidatorIntervalInSec` | listed with `PeerType=waiting`, shielded from Cleanup |

## Design decision: waiting counts toward `klv_live_validator_nodes`

Issue #114 required an explicit decision. Waiting nodes count toward the existing metric (no separate metric), because the sender publishes `klv_node_type=validator` for waiting keys and the issue requires sender and monitor to agree, and the `MetricPeerType` doc comment (core/metrics.go) has always named "waiting list" as an expected value.

**Operator note:** the metric's meaning changes. It now includes registered validators that do not participate in consensus yet, so dashboards or alerts comparing it against consensus-group size shift meaning. During a rolling upgrade, old and new nodes report different values for the same network state (the classification is computed locally per node; the heartbeat wire format is unchanged, so mixed-version networks are safe).

**Release note (for the next tagged release):** `klv_live_validator_nodes` now includes actively heartbeating waiting-list validators, and a waiting validator's own node reports `klv_node_type=validator` / `klv_peer_type=waiting` (previously `observer`). Alert thresholds comparing `klv_live_validator_nodes` to the consensus or elected+eligible set size must be adjusted, or moved to the new gauge below. Startup already initialised `klv_node_type=validator` unconditionally (`metrics.InitMetrics`), so before this change a waiting node's reported type flipped from `validator` to `observer` on its first heartbeat; that inconsistency disappears.

**New metric (added after review):** `klv_live_consensus_validator_nodes` counts only actively heartbeating elected and eligible validators, restoring the old "counts nodes that can actually sign" invariant next to the widened gauge. Alerts that guarded consensus liveness should target this series.

## Review round (fbsobreira, changes requested at 6121d30)

All four in-scope items are fixed and both requested decisions are made explicit:

1. **termui slot state** ([widgetsRender.go](../blob/HEAD/statusHandler/view/termuic/termuiRenders/widgetsRender.go)): a waiting node rendered a blank `Consensus slot state` (node type is validator, but no subround ever writes the slot-state metric for a non-elected node). The renderer now shows `N/A (waiting)` for waiting peers, analogous to `N/A (observer)`.
2. **Churn test masking** (monitor_test.go): the shielding test closed the monitor via `t.Cleanup`, so the background refresher could recreate an evicted entry between `Cleanup()` and the assertion. It now closes eagerly like its siblings; verified by mutation (dropping `WaitingList` from the predicate makes the test fail, where before the refresher masked it).
3. **Map-deref panic risk** (peerTypeProvider_test.go): `require.Contains` guards added before the `.pType` dereferences, so a wrong-key regression fails with an assertion instead of aborting the package binary.
4. **Predicate doc comment** (heartbeatMessageInfo.go): the caller enumeration (which was already incomplete) is replaced by the rule: every consumer classifying a peer as validator must use the predicate.
5. **Decision, map growth** (accepted, no cap): measured on mainnet the waiting list is currently 1 node (elected 21, eligible 103, inactive 52, jailed 23), far under the "at hundreds, worth a cap" threshold from the review. Revisit if registration economics change; a cap needs a design round anyway because evict-and-rematerialize would reintroduce the churn cycle for never-broadcast entries.
6. **Decision, metric semantics** (separate gauge added): `klv_live_consensus_validator_nodes` (elected+eligible only) re-establishes the old invariant; `klv_live_validator_nodes` keeps the widened meaning. Both published from the same refresh pass; pinned by the extended metric test (waiting+eligible+observer active gives live=2, consensus=1, connected=3).

## Changes

1. **Cache seeding** ([core/process/peer/peerTypeProvider.go:118-146](../blob/HEAD/core/process/peer/peerTypeProvider.go)): `createNewCache` now queries `GetAllWaitingValidatorsKeys` first, then elected, then eligible; later writes win, so the precedence waiting < elected < eligible holds by insertion order. The getter follows the PR #110 fail-fast contract: any error aborts the build and `UpdateCache` keeps the previous cache. All three coordinator getters fail on the identical single condition (`ErrEpochNodesConfigDoesNotExist` on the same epoch lookup), so this adds no new failure mode where the previous code would have succeeded.
2. **Shared predicate** ([node/heartbeat/process/heartbeatMessageInfo.go:209-217](../blob/HEAD/node/heartbeat/process/heartbeatMessageInfo.go)): `isValidatorPeerType` (waiting, elected, eligible) is the single definition used by `GetIsValidator`, `Monitor.shouldSkipValidator`, and `Sender.updateMetrics`, so the three consumers cannot diverge again; the divergence between these call sites is exactly what issue #114 documents.
3. **Metric counting** (`GetIsValidator`, [heartbeatMessageInfo.go:219-224](../blob/HEAD/node/heartbeat/process/heartbeatMessageInfo.go)): active waiting nodes count in `klv_live_validator_nodes`.
4. **Cleanup shielding** (`shouldSkipValidator`, [node/heartbeat/process/monitor.go:484-502](../blob/HEAD/node/heartbeat/process/monitor.go)): waiting entries are never deleted by `Cleanup`, which ends the delete/recreate churn cycle over the waiting set (entry materialized every refresh tick, deleted on every Cleanup). The shielding is self-correcting: when a key leaves the waiting list, the next cache rebuild drops it, `getValsForUpdate` demotes the inactive entry to observer, and Cleanup removes it after the hide interval (pinned by a test).
5. **Race fix on the rewritten line** ([monitor.go:484-502](../blob/HEAD/node/heartbeat/process/monitor.go)): `shouldSkipValidator` read `v.peerType` and `v.timestamp` without `v.updateMutex` while `HeartbeatReceived` writes them under only that mutex (concurrent path: p2p heartbeat processing vs `/node/heartbeatstatus` triggering `Cleanup`). Pre-existing race, reproduced with an actual race-detector report during the pre-push audit; since this PR rewrites that exact expression, the fields are now snapshotted under one `RLock`. Reads are direct field reads, not `GetIsActive`/`GetIsValidator`, because those take the exclusive lock and `RWMutex` is not reentrant.
6. **Sender hardening** ([node/heartbeat/process/sender.go:176-186](../blob/HEAD/node/heartbeat/process/sender.go)): `updateMetrics` derives `klv_node_type` from `isValidatorPeerType` instead of the inverted "anything not observer is a validator" heuristic. Behavior today is identical (the provider's value domain is exactly waiting/elected/eligible plus observer fallback, and the sender test passes unchanged against develop); this closes the last consumer that could silently diverge if another list is ever seeded into the cache.
7. **SonarQube S1186** ([peerTypeProvider.go:96-98](../blob/HEAD/core/process/peer/peerTypeProvider.go)): nested comment added to the empty epoch-start prepare handler. Strictly outside the #114 scope, included deliberately: it is the only open SonarQube issue on a touched file (open since 2021) and the pre-push gate requires clearing those.
8. **Test hook** (node/heartbeat/mock/peerTypeProviderStub.go): `GetAllPeerTypeInfosCalled` hook added; nil-handler default unchanged, so no other test changes meaning.

## Tests

Every behavior test fails against the unpatched production code (verified by reverting the production files and re-running; independently replicated by a reviewer applying only the test diffs to a develop worktree):

- `TestPeerTypeProvider_ComputeForPubKey_PartitionScenarios`: partition table per the issue's test-design note (waiting/elected/eligible/absent), matching the `computeNodesConfigFromList` partition; no overlap-precedence tests, as specified.
- `TestPeerTypeProvider_CreateNewCache_FailsOnAnyGetterError`: table over all three getters; a failing waiting getter aborts the whole build.
- `TestPeerTypeProvider_UpdateCache_KeepsPreviousOnFailure`: a populated cache including a waiting entry survives a failing refresh.
- `TestHeartbeatMessageInfo_GetIsValidator_PeerTypeWaitingShouldReturnTrue` / `_PeerTypeObserverShouldReturnFalse`.
- `TestMonitor_RemoveInactiveValidatorsIfIntervalExceeded` (extended): waiting survives the hide interval alongside elected/eligible (6 entries, 3 survive).
- `TestMonitor_WaitingNodeSurvivesRefreshAndCleanupRounds`: churn regression test; the entry survives repeated refresh plus Cleanup rounds far past the hide interval.
- `TestMonitor_ActiveWaitingNodeCountsAsLiveValidator`: active waiting node counts (1) while observer only counts as connected (2); made deterministic by stopping the background refresher before installing the status handler.
- `TestMonitor_UnstakedWaitingNodeIsDemotedAndCleaned`: the shielding exit path; a key that leaves the waiting list is demoted to observer and removed by Cleanup (guards against permanent ghost entries).
- `TestSender_SendHeartbeatUpdatesNodeTypeMetrics`: waiting reports `validator`/`waiting`, observer reports `observer`/`observer`; pins the sender-monitor agreement (passes on develop by design, guarding the cross-side contract).

## Verification

- `go build ./...`, `gofmt`/`goimports` clean, `golangci-lint` clean on the changed trees (one pre-existing errorlint finding in the untouched `node/heartbeat/storage/heartbeatStorer.go`).
- `go test -race -count=1` green on `core/process/peer` and all `node/heartbeat` packages.
- SonarQube analysis of every changed source file: zero new issues; line coverage verified on every changed production line including both branches of the new error path (mocks are coverage-excluded by project config).
- Full `-short` suite (CI mirror): 247 packages ok; the only failure observed is the pre-existing flaky `TestMonitor_ObserverGapValidator*` family (issue #106), which fails at comparable rates on a clean develop worktree (3/8 runs) and whose mechanism this PR does not touch; root-cause analysis posted on #106.
- Pre-push audit: security, code-review, Go-idiom, and an independent full-diff reviewer; all findings fixed in this PR or tracked (see #106), none Critical.

## Follow-ups (tracked, not in this PR)

- #106: pre-existing flaky monitor uptime tests; may cause an unrelated red CI check on this PR.
- #113: seed providers with the bootstrapped epoch (touches the same file; no conflict, this PR only changes what the cache contains, not when it becomes correct).
- #116: jailed/leaving validators still report `observer` (the leaving list is not queried anywhere); same problem class, deliberately out of scope here.
- #117: consolidate the per-entry locking in `computeAllHeartbeatMessages` into one snapshot (review follow-up; pre-existing pattern, enlarged relevance now that the map includes waiting).
- #118: deduplicate the getter/log/computePeerType blocks in `createNewCache` and the validatorsProvider sibling (review nudge).

## Summary

- Correct heartbeat peer-type classification for waiting-list validators.
- Include active waiting validators in `klv_live_validator_nodes`.
- Add `klv_live_consensus_validator_nodes` for elected and eligible validators only.
- Preserve waiting validators during heartbeat cleanup and refresh cycles.
- Demote and remove validators after they leave the waiting list and exceed the inactivity interval.
- Protect monitor state reads with locking to prevent data races.
- Make cache refresh fail fast and preserve the existing cache when any validator-list lookup fails.
- Update terminal UI output to show `N/A (waiting list)` for waiting validators.
- Add shared validator and consensus-capable predicates, test hooks, and coverage for cache errors, peer classifications, metrics, cleanup, and sender behavior.

The changes affect networking, monitoring, metrics, and terminal status reporting. They do not change consensus execution, transaction processing, state management, or KVM behavior. The cache error handling and monitor locking reduce risks from partial state updates and concurrent access. Waiting-list cache growth remains unbounded by design.

### DIFF
--- a/core/process/peer/peerTypeProvider.go
+++ b/core/process/peer/peerTypeProvider.go
@@ -93,7 +93,9 @@ func (ptp *PeerTypeProvider) epochStartEventHandler() sharding.EpochStartActionH
 				"epoch", hdr.GetEpoch())
 			ptp.UpdateCache(hdr.GetEpoch())
 		},
-		func(_ data.HeaderHandler) {},
+		func(_ data.HeaderHandler) {
+			// nothing to prepare before an epoch start; the cache is rebuilt in the action handler above
+		},
 		core.IndexerOrder,
 	)
 
@@ -119,6 +121,15 @@ func (ptp *PeerTypeProvider) createNewCache(
 	epoch uint32,
 ) (map[string]*peerListAndShard, bool) {
 	newCache := make(map[string]*peerListAndShard)
+
+	// waiting is seeded first so elected and eligible entries take precedence
+	// if a key ever appears in more than one list
+	nodesMapWaiting, err := ptp.nodesCoordinator.GetAllWaitingValidatorsKeys(epoch, false)
+	if err != nil {
+		log.Warn("peerTypeProvider - GetAllWaitingValidatorsKeys failed", "epoch", epoch, "error", err)
+		return nil, false
+	}
+	computePeerType(newCache, nodesMapWaiting, core.WaitingList)
 
 	nodesMapElected, err := ptp.nodesCoordinator.GetAllElectedValidatorsKeys(epoch, false)
 	if err != nil {
--- a/core/process/peer/peerTypeProvider_test.go
+++ b/core/process/peer/peerTypeProvider_test.go
@@ -355,6 +355,9 @@ func TestPeerTypeProvider_UpdateCache_KeepsPreviousOnFailure(t *testing.T) {
 	ptp.UpdateCache(99)
 
 	require.Len(t, ptp.cache, 3)
+	require.Contains(t, ptp.cache, "waiting1")
+	require.Contains(t, ptp.cache, "elected1")
+	require.Contains(t, ptp.cache, "eligible1")
 	assert.Equal(t, core.WaitingList, ptp.cache["waiting1"].pType)
 	assert.Equal(t, core.ElectedList, ptp.cache["elected1"].pType)
 	assert.Equal(t, core.EligibleList, ptp.cache["eligible1"].pType)
--- a/core/process/peer/peerTypeProvider_test.go
+++ b/core/process/peer/peerTypeProvider_test.go
@@ -326,6 +326,9 @@ func TestPeerTypeProvider_UpdateCache_KeepsPreviousOnFailure(t *testing.T) {
 
 	arg := createDefaultArgPeerTypeProvider()
 	arg.NodesCoordinator = &mock.NodesCoordinatorMock{
+		GetAllWaitingValidatorsKeysCalled: func() ([][]byte, error) {
+			return [][]byte{[]byte("waiting1")}, nil
+		},
 		GetAllElectedValidatorsKeysCalled: func() ([][]byte, error) {
 			return [][]byte{[]byte("elected1")}, nil
 		},
@@ -341,17 +344,18 @@ func TestPeerTypeProvider_UpdateCache_KeepsPreviousOnFailure(t *testing.T) {
 	}
 
 	ptp.UpdateCache(0)
-	require.Len(t, ptp.cache, 2)
+	require.Len(t, ptp.cache, 3)
 
 	ptp.nodesCoordinator = &mock.NodesCoordinatorMock{
-		GetAllElectedValidatorsKeysCalled: func() ([][]byte, error) {
+		GetAllWaitingValidatorsKeysCalled: func() ([][]byte, error) {
 			return nil, fmt.Errorf("epoch config does not exist")
 		},
 	}
 
 	ptp.UpdateCache(99)
 
-	require.Len(t, ptp.cache, 2)
+	require.Len(t, ptp.cache, 3)
+	assert.Equal(t, core.WaitingList, ptp.cache["waiting1"].pType)
 	assert.Equal(t, core.ElectedList, ptp.cache["elected1"].pType)
 	assert.Equal(t, core.EligibleList, ptp.cache["eligible1"].pType)
 }
@@ -359,24 +363,97 @@ func TestPeerTypeProvider_UpdateCache_KeepsPreviousOnFailure(t *testing.T) {
 func TestPeerTypeProvider_CreateNewCache_FailsOnAnyGetterError(t *testing.T) {
 	t.Parallel()
 
+	okGetter := func() ([][]byte, error) {
+		return [][]byte{[]byte("pk1")}, nil
+	}
+	failingGetter := func() ([][]byte, error) {
+		return nil, fmt.Errorf("list unavailable")
+	}
+
+	testCases := []struct {
+		name        string
+		coordinator *mock.NodesCoordinatorMock
+	}{
+		{
+			name: "waiting getter fails",
+			coordinator: &mock.NodesCoordinatorMock{
+				GetAllWaitingValidatorsKeysCalled:  failingGetter,
+				GetAllElectedValidatorsKeysCalled:  okGetter,
+				GetAllEligibleValidatorsKeysCalled: okGetter,
+			},
+		},
+		{
+			name: "elected getter fails",
+			coordinator: &mock.NodesCoordinatorMock{
+				GetAllWaitingValidatorsKeysCalled:  okGetter,
+				GetAllElectedValidatorsKeysCalled:  failingGetter,
+				GetAllEligibleValidatorsKeysCalled: okGetter,
+			},
+		},
+		{
+			name: "eligible getter fails",
+			coordinator: &mock.NodesCoordinatorMock{
+				GetAllWaitingValidatorsKeysCalled:  okGetter,
+				GetAllElectedValidatorsKeysCalled:  okGetter,
+				GetAllEligibleValidatorsKeysCalled: failingGetter,
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ptp := PeerTypeProvider{
+				nodesCoordinator: tc.coordinator,
+				cache:            nil,
+				mutCache:         sync.RWMutex{},
+			}
+
+			cache, ok := ptp.createNewCache(0)
+
+			assert.False(t, ok)
+			assert.Nil(t, cache)
+		})
+	}
+}
+
+func TestPeerTypeProvider_ComputeForPubKey_PartitionScenarios(t *testing.T) {
+	t.Parallel()
+
 	arg := createDefaultArgPeerTypeProvider()
 	arg.NodesCoordinator = &mock.NodesCoordinatorMock{
+		GetAllWaitingValidatorsKeysCalled: func() ([][]byte, error) {
+			return [][]byte{[]byte("waiting1")}, nil
+		},
 		GetAllElectedValidatorsKeysCalled: func() ([][]byte, error) {
 			return [][]byte{[]byte("elected1")}, nil
 		},
 		GetAllEligibleValidatorsKeysCalled: func() ([][]byte, error) {
-			return nil, fmt.Errorf("eligible list unavailable")
+			return [][]byte{[]byte("eligible1")}, nil
 		},
 	}
 
-	ptp := PeerTypeProvider{
-		nodesCoordinator: arg.NodesCoordinator,
-		cache:            nil,
-		mutCache:         sync.RWMutex{},
+	ptp, err := NewPeerTypeProvider(arg)
+	require.Nil(t, err)
+
+	// a key can only be in one of the three lists in production
+	// (computeNodesConfigFromList partitions on validatorInfo.List),
+	// so the scenarios cover the partition, not overlap states
+	testCases := []struct {
+		name         string
+		pubKey       string
+		expectedType core.PeerType
+	}{
+		{name: "key in waiting list resolves to waiting", pubKey: "waiting1", expectedType: core.WaitingList},
+		{name: "key in elected list resolves to elected", pubKey: "elected1", expectedType: core.ElectedList},
+		{name: "key in eligible list resolves to eligible", pubKey: "eligible1", expectedType: core.EligibleList},
+		{name: "key in no list resolves to observer", pubKey: "unknown1", expectedType: core.ObserverList},
 	}
 
-	cache, ok := ptp.createNewCache(0)
-
-	assert.False(t, ok)
-	assert.Nil(t, cache)
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			peerType, _, err := ptp.ComputeForPubKey([]byte(tc.pubKey))
+			assert.Nil(t, err)
+			assert.Equal(t, tc.expectedType, peerType)
+		})
+	}
 }
--- a/node/heartbeat/process/heartbeatMessageInfo.go
+++ b/node/heartbeat/process/heartbeatMessageInfo.go
@@ -206,19 +206,34 @@ func (hbmi *heartbeatMessageInfo) GetIsActive() bool {
 	return isActive
 }
 
-// isValidatorPeerType returns true for the peer types that count as validators
-// (waiting, elected, eligible). GetIsValidator and Monitor.shouldSkipValidator
-// must share this definition so the klv_live_validator_nodes metric and the
-// Cleanup shielding cannot diverge.
-func isValidatorPeerType(peerType string) bool {
+// isConsensusCapablePeerType is the single definition of which peer types can
+// participate in the consensus rotation: elected and eligible. It feeds the
+// klv_live_consensus_validator_nodes metric.
+func isConsensusCapablePeerType(peerType string) bool {
 	return peerType == string(core.ElectedList) ||
-		peerType == string(core.EligibleList) ||
+		peerType == string(core.EligibleList)
+}
+
+// isValidatorPeerType is the single definition of which peer types count as
+// validators: waiting plus the consensus-capable types. Every consumer that
+// classifies a peer as a validator (metrics counting, Cleanup shielding,
+// node-type reporting) must use this predicate instead of comparing peer
+// types itself, so the classifications cannot diverge.
+func isValidatorPeerType(peerType string) bool {
+	return isConsensusCapablePeerType(peerType) ||
 		peerType == string(core.WaitingList)
 }
 
 // GetIsValidator will return true is the peer is a validator
 func (hbmi *heartbeatMessageInfo) GetIsValidator() bool {
-	hbmi.updateMutex.Lock()
-	defer hbmi.updateMutex.Unlock()
+	hbmi.updateMutex.RLock()
+	defer hbmi.updateMutex.RUnlock()
 	return isValidatorPeerType(hbmi.peerType)
+}
+
+// GetIsConsensusCapable will return true if the peer can participate in consensus
+func (hbmi *heartbeatMessageInfo) GetIsConsensusCapable() bool {
+	hbmi.updateMutex.RLock()
+	defer hbmi.updateMutex.RUnlock()
+	return isConsensusCapablePeerType(hbmi.peerType)
 }
--- a/node/heartbeat/process/heartbeatMessageInfo.go
+++ b/node/heartbeat/process/heartbeatMessageInfo.go
@@ -206,9 +206,19 @@ func (hbmi *heartbeatMessageInfo) GetIsActive() bool {
 	return isActive
 }
 
+// isValidatorPeerType returns true for the peer types that count as validators
+// (waiting, elected, eligible). GetIsValidator and Monitor.shouldSkipValidator
+// must share this definition so the klv_live_validator_nodes metric and the
+// Cleanup shielding cannot diverge.
+func isValidatorPeerType(peerType string) bool {
+	return peerType == string(core.ElectedList) ||
+		peerType == string(core.EligibleList) ||
+		peerType == string(core.WaitingList)
+}
+
 // GetIsValidator will return true is the peer is a validator
 func (hbmi *heartbeatMessageInfo) GetIsValidator() bool {
 	hbmi.updateMutex.Lock()
 	defer hbmi.updateMutex.Unlock()
-	return hbmi.peerType == string(core.ElectedList) || hbmi.peerType == string(core.EligibleList)
+	return isValidatorPeerType(hbmi.peerType)
 }
--- a/node/heartbeat/process/monitor.go
+++ b/node/heartbeat/process/monitor.go
@@ -482,11 +482,17 @@ func (m *Monitor) GetHeartbeats() []data.PubKeyHeartbeat {
 }
 
 func (m *Monitor) shouldSkipValidator(v *heartbeatMessageInfo) bool {
-	isInactiveObserver := !v.GetIsActive() &&
-		(v.peerType != string(core.ElectedList) &&
-			v.peerType != string(core.EligibleList))
-	if isInactiveObserver {
-		lastInactiveInterval := m.timer.Now().Sub(v.timestamp)
+	// snapshot the fields under updateMutex: HeartbeatReceived writes them while
+	// holding only that mutex, so raw reads here race with inbound heartbeats
+	v.updateMutex.RLock()
+	isActive := v.isActive
+	peerType := v.peerType
+	timestamp := v.timestamp
+	v.updateMutex.RUnlock()
+
+	isInactiveNonValidator := !isActive && !isValidatorPeerType(peerType)
+	if isInactiveNonValidator {
+		lastInactiveInterval := m.timer.Now().Sub(timestamp)
 		if lastInactiveInterval.Seconds() > float64(m.hideInactiveValidatorIntervalInSec) {
 			return true
 		}
--- a/node/heartbeat/process/monitor.go
+++ b/node/heartbeat/process/monitor.go
@@ -373,6 +373,7 @@ func (m *Monitor) computePeerType(pubkey []byte) string {
 func (m *Monitor) computeAllHeartbeatMessages() {
 	m.mutHeartbeatMessages.Lock()
 	counterActiveValidators := 0
+	counterActiveConsensusValidators := 0
 	counterConnectedNodes := 0
 	hbChangedStateToInactiveMap := make(map[string]*heartbeatMessageInfo)
 	for key, v := range m.heartbeatMessages {
@@ -386,6 +387,9 @@ func (m *Monitor) computeAllHeartbeatMessages() {
 			if v.GetIsValidator() {
 				counterActiveValidators++
 			}
+			if v.GetIsConsensusCapable() {
+				counterActiveConsensusValidators++
+			}
 		}
 		changedStateToInactive := previousActive && !isActive
 		if changedStateToInactive {
@@ -397,8 +401,9 @@ func (m *Monitor) computeAllHeartbeatMessages() {
 	go m.SaveMultipleHeartbeatMessageInfos(hbChangedStateToInactiveMap)
 
 	m.mutAppStatusHandler.Lock()
-	m.appStatusHandler.SetUInt64Value(core.MetricLiveValidatorNodes, uint64(counterActiveValidators)) // #nosec G115
-	m.appStatusHandler.SetUInt64Value(core.MetricConnectedNodes, uint64(counterConnectedNodes))       // #nosec G115
+	m.appStatusHandler.SetUInt64Value(core.MetricLiveValidatorNodes, uint64(counterActiveValidators))                   // #nosec G115
+	m.appStatusHandler.SetUInt64Value(core.MetricLiveConsensusValidatorNodes, uint64(counterActiveConsensusValidators)) // #nosec G115
+	m.appStatusHandler.SetUInt64Value(core.MetricConnectedNodes, uint64(counterConnectedNodes))                         // #nosec G115
 	m.mutAppStatusHandler.Unlock()
 }
 
--- a/node/heartbeat/process/monitor_test.go
+++ b/node/heartbeat/process/monitor_test.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/klever-io/klever-go/core"
+	"github.com/klever-io/klever-go/data/state"
 	"github.com/klever-io/klever-go/network/p2p"
 	"github.com/klever-io/klever-go/node/heartbeat"
 	"github.com/klever-io/klever-go/node/heartbeat/data"
@@ -16,6 +17,7 @@ import (
 	"github.com/klever-io/klever-go/node/heartbeat/storage"
 	"github.com/klever-io/klever-go/tools/check"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 var fromConnectedPeerId = core.PeerID("from connected peer Id")
@@ -409,6 +411,7 @@ func TestMonitor_RemoveInactiveValidatorsIfIntervalExceeded(t *testing.T) {
 	pubKey2 := "pk2-eligible"
 	pubKey3 := "pk3-observer"
 	pubKey4 := "pk4-inactive"
+	pubKey5 := "pk5-waiting"
 
 	storer, _ := storage.NewHeartbeatDbStorer(mock.NewStorerMock(), &mock.MarshalizerMock{})
 
@@ -436,6 +439,8 @@ func TestMonitor_RemoveInactiveValidatorsIfIntervalExceeded(t *testing.T) {
 					return core.ObserverList, 0, nil
 				case pubKey4:
 					return core.InactiveList, 0, nil
+				case pubKey5:
+					return core.WaitingList, 0, nil
 				}
 				return core.ObserverList, 0, nil
 			},
@@ -447,28 +452,32 @@ func TestMonitor_RemoveInactiveValidatorsIfIntervalExceeded(t *testing.T) {
 		HideInactiveValidatorIntervalInSec: 600,
 	}
 	mon, _ := process.NewMonitor(arg)
+	defer func() {
+		_ = mon.Close()
+	}()
 	mon.SendHeartbeatMessage(&data.Heartbeat{Pubkey: []byte(pkValidator)})
 	mon.SendHeartbeatMessage(&data.Heartbeat{Pubkey: []byte(pubKey1)})
 	mon.SendHeartbeatMessage(&data.Heartbeat{Pubkey: []byte(pubKey2)})
 	mon.SendHeartbeatMessage(&data.Heartbeat{Pubkey: []byte(pubKey3)})
 	mon.SendHeartbeatMessage(&data.Heartbeat{Pubkey: []byte(pubKey4)})
+	mon.SendHeartbeatMessage(&data.Heartbeat{Pubkey: []byte(pubKey5)})
 
 	// Check that all are added
 	mon.RefreshHeartbeatMessageInfo()
 	hbStatus := mon.GetHeartbeats()
-	assert.Equal(t, 5, len(hbStatus))
+	assert.Equal(t, 6, len(hbStatus))
 
 	timer.IncrementSeconds(int(arg.HideInactiveValidatorIntervalInSec) - 20)
 	mon.RefreshHeartbeatMessageInfo()
 	hbStatus = mon.GetHeartbeats()
-	assert.Equal(t, 5, len(hbStatus))
+	assert.Equal(t, 6, len(hbStatus))
 
 	// increase to over HideInactiveValidatorIntervalInSec ~ 10 min
 	timer.IncrementSeconds(int(arg.HideInactiveValidatorIntervalInSec) + 10)
 	mon.RefreshHeartbeatMessageInfo()
 	hbStatus = mon.GetHeartbeats()
-	// check if pk1 and pk2 are still on
-	assert.Equal(t, 2, len(hbStatus))
+	// check if pk1, pk2 and pk5 are still on
+	assert.Equal(t, 3, len(hbStatus))
 }
 
 func TestMonitor_ProcessReceivedMessageImpersonatedMessageShouldErr(t *testing.T) {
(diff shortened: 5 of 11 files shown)


core/process/peer/peerTypeProvider_test.go
```diff
@@ -341,42 +344,116 @@ func TestPeerTypeProvider_UpdateCache_KeepsPreviousOnFailure(t *testing.T) {
...
 
-	assert.False(t, ok)
-	assert.Nil(t, cache)
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			peerType, _, err := ptp.ComputeForPubKey([]byte(tc.pubKey))
+			assert.Nil(t, err)
+			assert.Equal(t, tc.expectedType, peerType)
+		})
+	}
```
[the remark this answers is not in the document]
MathijsBok: Skipped deliberately, on two grounds:

1. Issue #114's test-design note rules this out explicitly: a key can never be in two of the three lists in production (`computeNodesConfigFromList` partitions on `validatorInfo.List`), so "precedence tests describing overlap states are not needed; a scenario table covering the partition is". This table is that required coverage.
2. The overwrite order is nevertheless already regression-protected at the mechanics level: the pre-existing `TestPeerTypeProvider_CreateNewCacheScenarios` seeds `elected1` into both the elected and eligible getters and asserts `core.EligibleList` wins, which pins the later-write-wins behavior that the waiting < elected < eligible seeding order relies on.

So the contract this suggestion protects is covered, and the issue defining this PR's scope explicitly asked not to add overlap-state scenarios.

node/heartbeat/process/monitor_test.go
```diff
@@ -447,28 +452,32 @@ func TestMonitor_RemoveInactiveValidatorsIfIntervalExceeded(t *testing.T) {
 		HideInactiveValidatorIntervalInSec: 600,
 	}
 	mon, _ := process.NewMonitor(arg)
+	defer func() {
+		_ = mon.Close()
+	}()
```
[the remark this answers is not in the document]
MathijsBok: Fixed in 87016fa, all six sites:

- `monitor_test.go` `TestMonitor_RemoveInactiveValidatorsIfIntervalExceeded` and `TestMonitor_WaitingNodeSurvivesRefreshAndCleanupRounds`: the deferred unchecked `Close` calls are replaced with `t.Cleanup(func() { require.NoError(t, mon.Close()) })`.
- `monitor_test.go` `TestMonitor_ActiveWaitingNodeCountsAsLiveValidator`: `require.NoError` on both the early `mon.Close()` and `mon.SetAppStatusHandler(...)`.
- `monitor_test.go` `TestMonitor_UnstakedWaitingNodeIsDemotedAndCleaned`: `require.NoError(t, mon.Close())`.
- `sender_test.go` `TestSender_SendHeartbeatUpdatesNodeTypeMetrics`: the `NewSender` error is captured and asserted with `require.NoError` before `SendHeartbeat`.

Verified after the change: `go build ./node/heartbeat/...` and `go test -race -count=1 ./node/heartbeat/process/` both green.

node/heartbeat/process/monitor.go
```diff
@@ -482,11 +482,17 @@ func (m *Monitor) GetHeartbeats() []data.PubKeyHeartbeat {
...
 
 func (m *Monitor) shouldSkipValidator(v *heartbeatMessageInfo) bool {
-	isInactiveObserver := !v.GetIsActive() &&
-		(v.peerType != string(core.ElectedList) &&
-			v.peerType != string(core.EligibleList))
-	if isInactiveObserver {
-		lastInactiveInterval := m.timer.Now().Sub(v.timestamp)
+	// snapshot the fields under updateMutex: HeartbeatReceived writes them while
+	// holding only that mutex, so raw reads here race with inbound heartbeats
+	v.updateMutex.RLock()
```
MathijsBok, line 492: ✅ **Verified clean in self-review** (S1, confidence 10/10): before requesting human review we ran this diff through an automated review against develop; this note documents what was explicitly checked here. The pre-existing `shouldSkipValidator` read `v.peerType` and `v.timestamp` with no lock while `HeartbeatReceived`, driven by inbound p2p heartbeats, writes those fields under `updateMutex`: a data race (CWE-362) reachable from network input via `GetHeartbeats` -> `Cleanup`. The rewrite snapshots `isActive`, `peerType` and `timestamp` under a single `RLock`, which removes the race and the TOCTOU between the previously separate reads. Checked: lock ordering stays uniform across the file (`mutHeartbeatMessages` outer, per-entry `updateMutex` inner), so no deadlock path is introduced; `go test -race` green on the package; cyclomatic complexity went from 5 to 4; both outcomes are exercised in the coverage profile (skip branch 7 hits, keep branch 88 hits). The rename to `isInactiveNonValidator` matches the actual condition.

MathijsBok (commented):
## Automated Self-Review

Before requesting human review, this PR's diff was put through an automated three-agent review against develop (code quality, security, code metrics; findings deduplicated, confidence 8+ only). This is the author's own quality gate, published so the reviewer can see what was already checked. Outcome: **0 critical, 0 warnings, 2 suggestions (both addressed on the branch), 4 verification notes** documenting what was explicitly checked and came out clean.

Legend: 🔴 Critical | 🟡 Warning | 💡 Suggestion | ✅ Verified clean in self-review

Both suggestions are addressed: S4 via the release-note block added to the PR description (plus a confirmation request to the reviewer in the comment), R6 via commit 6121d30.

Verification evidence from the review run: `go test -race` green on both changed package trees, govulncheck reports 0 reachable vulnerabilities in the changed packages, every changed production function is far under the complexity thresholds with 100% changed-line coverage, and the only observed test failure is the pre-existing ObserverGap flake tracked in #106 (identical failure rates measured at base and head).

node/heartbeat/process/heartbeatMessageInfo.go
```diff
@@ -206,9 +206,19 @@ func (hbmi *heartbeatMessageInfo) GetIsActive() bool {
 	return isActive
 }
 
+// isValidatorPeerType returns true for the peer types that count as validators
+// (waiting, elected, eligible). GetIsValidator and Monitor.shouldSkipValidator
+// must share this definition so the klv_live_validator_nodes metric and the
+// Cleanup shielding cannot diverge.
+func isValidatorPeerType(peerType string) bool {
```
MathijsBok: ✅ **Verified clean in self-review** (S2, confidence 10/10): checked the shared `isValidatorPeerType` predicate for consistency and boundedness. All three consumers now use one definition (`GetIsValidator` for klv_live_validator_nodes, `Monitor.shouldSkipValidator` for Cleanup shielding, `Sender.updateMetrics` for klv_node_type), closing the divergence class from issue #114; a repo-wide search found no remaining divergent peer-type comparisons in the heartbeat package, and the only production `PeerTypeProviderHandler` implementation is the updated provider. The Cleanup shield is bounded by chain state rather than attacker-creatable entries: `TestMonitor_UnstakedWaitingNodeIsDemotedAndCleaned` pins that a key leaving the waiting list is demoted on refresh and then evicted, and liveness spoofing is blocked upstream (peer signature binds the claimed Pubkey to the message Pid; mismatches are blacklisted). Metrics from the review run: net cognitive complexity of the changed production code decreased while behavior was added; the predicate is 100% covered with dedicated waiting=true and observer=false cases.

core/process/peer/peerTypeProvider.go
```diff
@@ -120,6 +122,15 @@ func (ptp *PeerTypeProvider) createNewCache(
 ) (map[string]*peerListAndShard, bool) {
 	newCache := make(map[string]*peerListAndShard)
 
+	// waiting is seeded first so elected and eligible entries take precedence
+	// if a key ever appears in more than one list
+	nodesMapWaiting, err := ptp.nodesCoordinator.GetAllWaitingValidatorsKeys(epoch, false)
```
MathijsBok, line 127: ✅ **Verified clean in self-review** (S3, confidence 10/10): checked the waiting-list seeding against the fail-secure contract from PR #110. Any getter error returns `(nil, false)` and `UpdateCache` keeps the previous cache, so a partially built cache, which would demote every validator to observer and un-shield their monitor entries, can never be installed (CWE-636 avoided). All three coordinator getters share the identical single failure condition (missing epoch config), so the added getter introduces no failure mode where the previous code would have succeeded. The seeding order gives waiting < elected < eligible precedence by insertion; production partitions the lists in `computeNodesConfigFromList`, so overlap cannot occur there. Pinned per getter by `TestPeerTypeProvider_CreateNewCache_FailsOnAnyGetterError`, and all three error branches show hits in the coverage profile.

node/heartbeat/process/heartbeatMessageInfo.go
```diff
@@ -206,9 +206,19 @@ func (hbmi *heartbeatMessageInfo) GetIsActive() bool {
...
+	return peerType == string(core.ElectedList) ||
+		peerType == string(core.EligibleList) ||
+		peerType == string(core.WaitingList)
```
MathijsBok, line 224: 💡 **Suggestion** (S4, confidence 9/10): Operator-facing semantics change worth an explicit release-note callout, not a code defect. Including `WaitingList` here means klv_live_validator_nodes now counts active waiting nodes (via `GetIsValidator`) and a waiting node's own klv_node_type flips to validator (via `Sender.updateMetrics`). Any dashboard or alert keyed to an absolute live-validator count silently shifts baseline after upgrading, which can mask or falsely trigger liveness alerts (OWASP A09:2021 in the alert-accuracy sense). Two facts surfaced during review: startup already initialised klv_node_type=validator unconditionally (`metrics.InitMetrics`), so this change actually removes the previous validator-to-observer flip a waiting node exhibited after its first heartbeat; and if operators need to distinguish consensus-eligible from waiting capacity, a separate waiting-nodes gauge is a possible follow-up.

**Addressed on this branch:** the PR description now carries a dedicated release-note block with the alert-threshold warning. @fbsobreira please confirm the combined-metric choice (waiting counts in klv_live_validator_nodes, no separate gauge) matches the direction you had in mind on the #110 review thread; a separate gauge can still be split out as a follow-up if preferred.

node/heartbeat/process/monitor_test.go
```diff
@@ -558,6 +567,150 @@ func TestMonitor_AddAndGetDoubleSignerPeersShouldWork(t *testing.T) {
 	assert.Equal(t, uint64(1), mon.GetNumInstancesOfPublicKey(string("pk3")))
 }
 
+func TestMonitor_WaitingNodeSurvivesRefreshAndCleanupRounds(t *testing.T) {
```
MathijsBok, line 570: ✅ **Verified clean in self-review** (R5, confidence 9/10): checked the new monitor tests for discriminating power, meaning: do they actually fail against the pre-fix code? All three do, verified by reverting the production files and re-running: `TestMonitor_WaitingNodeSurvivesRefreshAndCleanupRounds` reproduces the churn cycle (timer far past HideInactiveValidatorIntervalInSec, so the count assertion separates the old delete/recreate behavior from the new shielding), `TestMonitor_ActiveWaitingNodeCountsAsLiveValidator` pins the klv_live_validator_nodes and klv_connected_nodes values, and `TestMonitor_UnstakedWaitingNodeIsDemotedAndCleaned` covers the inverse path, proving the shielding cannot cause unbounded retention after a key leaves the waiting list. Calling `Close` before installing the status handler makes the metric assertions deterministic against the background refresher; everything is green under `go test -race`.

node/heartbeat/process/sender_test.go
```diff
@@ -293,6 +295,75 @@ func TestSender_SendHeartbeatShouldWork(t *testing.T) {
...
+		{
+			name:             "observer node reports observer node type",
+			peerType:         core.ObserverList,
+			expectedNodeType: string(core.NodeTypeObserver),
+			expectedPeerType: string(core.ObserverList),
+		},
+		{
+			// pins the allowlist default: a peer type outside the validator
+			// set (unreachable via the provider today) maps to observer
+			name:             "inactive node reports observer node type",
```
MathijsBok, line 322: 💡 **Suggestion** (R6, confidence 8/10): The waiting and observer cases also pass under the old inverted heuristic (anything not observer was validator), so nothing pinned the rewritten allowlist condition itself. A case with a peer type outside the validator set expecting NodeTypeObserver makes this test able to catch a regression of the sender condition.

**Addressed in 6121d30:** this InactiveList case was added and verified to fail against the pre-PR sender logic (reverting sender.go fails exactly this subtest), so the allowlist default is now pinned.

node/heartbeat/process/monitor.go
```diff
@@ -482,11 +482,17 @@ func (m *Monitor) GetHeartbeats() []data.PubKeyHeartbeat {
...
+	isActive := v.isActive
+	peerType := v.peerType
+	timestamp := v.timestamp
+	v.updateMutex.RUnlock()
+
+	isInactiveNonValidator := !isActive && !isValidatorPeerType(peerType)
```
fbsobreira, line 498: **Decision needed — this is correct per #114, but the consequence is new.**

Shielding waiting from `Cleanup` is exactly what the issue asked for; the *previous* behaviour (per-tick create/delete churn) was the bug. So the direction is right.

The consequence: waiting keys are now **both** auto-created by `computeInactiveHeartbeatMessages` (`:432-441`, fed by the newly waiting-seeded cache) **and** permanently exempt from eviction. For K registered validators sitting in the waiting list who never run a node:

- `GET /node/heartbeatstatus` returns K extra entries forever, with `isActive=false`, `timestamp=genesisTime` and ever-growing `totalDownTime`, that no configuration can hide
- `m.heartbeatMessages` permanently holds K extra entries, walked under the global `mutHeartbeatMessages` write lock on every inbound heartbeat via `computeAllHeartbeatMessages`

The shift that matters: elected+eligible was inherently **bounded by the consensus set**. Waiting is bounded only by registration. So both the API payload and the per-heartbeat lock-held walk now scale with a set you don't control.

**The question is just how big your waiting list is on mainnet.** At tens, ignore this. At hundreds or unbounded, worth a cap — e.g. keep the shielding but let `Cleanup` still evict waiting entries that have *never* broadcast and are older than some multiple of `HideInactiveValidatorIntervalInSec`, which preserves the fix (no churn for real waiting nodes) without the unbounded tail.

node/heartbeat/process/heartbeatMessageInfo.go
fbsobreira, line 224: **Decision needed — silent metric semantics change.**

`GetIsValidator` is consumed only by `Monitor.computeAllHeartbeatMessages`, which increments `counterActiveValidators` and publishes it as `core.MetricLiveValidatorNodes` (`monitor.go:386-400`). Before this diff that gauge counted only elected+eligible; now it also counts every active waiting-list peer.

On a node upgraded mid-epoch the gauge jumps by the size of the waiting list with no operator-side change.

The concrete risk: an alert of the form `klv_live_validator_nodes < consensusGroupSize*2/3` was a natural threshold under the old semantics, because the count was bounded by elected+eligible. Under the new semantics, if 15 of 21 elected validators go offline while a dozen waiting nodes keep heartbeating, the gauge **still reads above the alarm level** and the outage is never paged.

The old invariant — "this gauge counts nodes that can actually sign blocks" — is not re-established anywhere; no elected/eligible-only counter was added.

Two workable options: keep the widened gauge and update alerts as part of the rollout, or add a separate elected+eligible counter so both questions remain answerable. Either is fine — but it should be deliberate, and it belongs in the release notes since it changes an existing series' meaning.

node/heartbeat/process/sender.go
```diff
@@ -176,10 +176,8 @@ func (s *Sender) getCurrentPrivateAndPublicKeys() (crypto.PrivateKey, crypto.Pub
 func (s *Sender) updateMetrics(hb *heartbeatData.Heartbeat) {
 	result := s.computePeerList(hb.Pubkey)
 
-	nodeType := ""
-	if result == string(core.ObserverList) {
-		nodeType = string(core.NodeTypeObserver)
-	} else {
+	nodeType := string(core.NodeTypeObserver)
+	if isValidatorPeerType(result) {
 		nodeType = string(core.NodeTypeValidator)
```
fbsobreira, line 181: **In scope — small UI regression that falls out of the allowlist flip.**

To be clear, the allowlist itself is an improvement over the old `if result == observer` denylist — a future peer type now defaults to observer instead of silently becoming a validator. Good change.

The side effect: `updateMetrics` now sets `klv_node_type=validator` for a waiting node, and `statusHandler/view/termuic/termuiRenders/widgetsRender.go:298` branches on exactly that metric:

```go
if instanceType == string(core.NodeTypeObserver) → "Consensus slot state: N/A (observer)"
else                                             → GetConsensusSlotState()  // klv_consensus_slot_state
```

A waiting node is not in the elected set, so no BLS subround ever writes `klv_consensus_slot_state` past the `""` seeded at `cmd/node/metrics/metrics.go:56` and reset each slot at `subslotStartSlot.go:108`. So an operator running a waiting node in termui sees `Consensus slot state: ` — permanently blank — where before it correctly showed `N/A (observer)`.

Blank reads as "broken", which is worse than the honest `N/A`. The renderer branch needs to treat waiting like observer for this row (it is equally not-in-consensus), or the slot-state metric needs a non-empty default for non-elected nodes.

Same flip also relabels the node as `klv_build_info{node_type="validator"}` — correct under the new semantics, just worth noting alongside the gauge change.

node/heartbeat/process/monitor_test.go
```diff
@@ -558,6 +567,150 @@ func TestMonitor_AddAndGetDoubleSignerPeersShouldWork(t *testing.T) {
...
+		},
+		GetAllPeerTypeInfosCalled: func() []*state.PeerTypeInfo {
+			return []*state.PeerTypeInfo{
+				{PublicKey: pkWaiting, PeerType: string(core.WaitingList)},
+			}
+		},
+	}
+	mon, err := process.NewMonitor(arg)
+	require.Nil(t, err)
+	t.Cleanup(func() {
```
fbsobreira: **In scope — this test can pass while the thing it guards is broken.**

`mon.Close()` is deferred to `t.Cleanup`, so the Monitor's background refresh goroutine (ticker at `HeartbeatRefreshIntervalInSec=1`) is **still running** while the loop at `:608-613` asserts entry counts after each `mon.Cleanup()`.

That goroutine is the one that re-creates waiting entries via `computeInactiveHeartbeatMessages`. So if a future change drops `core.WaitingList` from `isValidatorPeerType`, `mon.Cleanup()` deletes the entry — and the refresher recreates it from `GetAllPeerTypeInfosCalled` before `assert.Equal(t, 1, mon.GetNumHearbeatMessages())` runs. The test reports PASS while the shielding it verifies is gone.

The tell that this is unintentional: the two sibling tests added in the same commit (`TestMonitor_ActiveWaitingNodeCountsAsLiveValidator`, `TestMonitor_UnstakedWaitingNodeIsDemotedAndCleaned`) both call `require.NoError(t, mon.Close())` **eagerly**, precisely so the refresher can't interfere. This one is the odd one out.

Close eagerly here too, before the Cleanup/assert loop.

core/process/peer/peerTypeProvider_test.go
```diff
@@ -341,42 +344,116 @@ func TestPeerTypeProvider_UpdateCache_KeepsPreviousOnFailure(t *testing.T) {
...
+		GetAllWaitingValidatorsKeysCalled: func() ([][]byte, error) {
 			return nil, fmt.Errorf("epoch config does not exist")
 		},
 	}
 
 	ptp.UpdateCache(99)
 
-	require.Len(t, ptp.cache, 2)
+	require.Len(t, ptp.cache, 3)
+	assert.Equal(t, core.WaitingList, ptp.cache["waiting1"].pType)
```
fbsobreira, line 361: **In scope — cheap, and it protects the whole package.**

`ptp.cache` is `map[string]*peerListAndShard`. A map miss yields a nil pointer, and `.pType` on it panics. `require.Len(t, ptp.cache, 3)` on the preceding line pins the **count**, not the keys.

Concrete regression this would mishandle: `createNewCache` seeding waiting under a different key — e.g. the flag flipped to `GetAllWaitingValidatorsKeys(epoch, true)`, which returns `OwnerAddress` instead of `PubKey`. The cache still has 3 entries, so `require.Len` passes, but there is no `"waiting1"` and line 358 panics.

A panic aborts the entire `core/process/peer` test binary, so CI shows a crash stack and every other test in the package is reported as failed/no-result — instead of a one-line "expected waiting, got X" pointing straight at the bug.

`entry, ok := ptp.cache["waiting1"]; require.True(t, ok)` before dereferencing, or `require.Contains`.

node/heartbeat/process/heartbeatMessageInfo.go
```diff
@@ -206,9 +206,19 @@ func (hbmi *heartbeatMessageInfo) GetIsActive() bool {
 	return isActive
 }
 
+// isValidatorPeerType returns true for the peer types that count as validators
```
fbsobreira: **In scope — trivial, but it is already wrong in the commit that adds it.**

The comment names two callers:

> `GetIsValidator` and `Monitor.shouldSkipValidator` must share this definition

There are three. The same commit adds `Sender.updateMetrics` (`sender.go:180`), which publishes the `klv_node_type` metric.

That matters more than a normal stale comment: a maintainer widening or narrowing this predicate will trust the list, audit the two named sites, and ship an unreviewed change to a published metric.

Caller enumerations go stale by construction. State the rule — which peer types count as validators, and why the definition must be shared — and drop the list.

node/heartbeat/process/monitor.go
```diff
@@ -482,11 +482,17 @@ func (m *Monitor) GetHeartbeats() []data.PubKeyHeartbeat {
 }
```
fbsobreira, line 487: **Out of scope — recorded, not a merge condition.** (Anchored here; the loop is at `computeAllHeartbeatMessages`, `:379`, outside this diff.)

Worth noting because this PR fixes exactly this pattern one function away and enlarges the map the loop walks.

`shouldSkipValidator` now takes a single `updateMutex.RLock` snapshot — correct. But `computeAllHeartbeatMessages` still performs four separate **exclusive** `Lock` acquisitions per entry: `GetIsActive`, `ComputeActive`, `GetIsActive`, `GetIsValidator` — and `GetIsActive` (`heartbeatMessageInfo.go:202`) and `GetIsValidator` (`:220`) use `Lock`, not `RLock`, for pure reads. All of it while holding the global `mutHeartbeatMessages` write lock.

With waiting seeded, each refresh tick is now `4*(elected+eligible+waiting)` exclusive round-trips with the global map lock held, while `addHeartbeatMessageToMap` blocks on that same lock for inbound heartbeats.

Also not a consistent snapshot: `HeartbeatReceived` can land between the `ComputeActive` and `GetIsValidator` calls, so `klv_live_validator_nodes` can be computed from an `isActive` from one state and a `peerType` from another.

The fix is the same shape you already applied in `shouldSkipValidator` — one `RLock` snapshot per entry — plus switching those two getters to `RLock`. Pre-existing, so I would not grow this diff for it, but it is the natural follow-up.

core/process/peer/peerTypeProvider.go
fbsobreira, line 127: Optional — third copy of the same shape.

`createNewCache` now has three near-identical getter / log / `computePeerType` blocks differing by three tokens. A table of `{name, peerType, getter}` with one loop collapses them.

The risk is specific: adding a fourth list is a copy-paste where the pasted `computePeerType(newCache, nodesMapX, core.WaitingList)` constant or the log string is easy to leave un-updated. Nothing in the compiler catches it, and the table test is itself copy-pasted — so every key of the new list would silently report the wrong `pType` through `ComputeForPubKey` and `GetAllPeerTypeInfos`.

The duplication predates this diff (and `validatorsProvider.go:219-232` has the same shape), so this is a nudge rather than a request.

fbsobreira (changes requested):
## Review at `6121d30f`

This is the cleanest of the four in this series. The split from #110 paid off: the change does what #114 asked, and two things go beyond it.

**Credit where it's due.** The shared `isValidatorPeerType` predicate makes the divergence I originally flagged *structurally impossible* rather than merely fixed — three consumers, one definition. And `sender.go` flipping from a denylist (`anything not observer is validator`) to an allowlist means a future peer type defaults to observer instead of silently counting as a validator. That's a better fix than what was asked for.

You also caught a genuine pre-existing data race in `shouldSkipValidator` (raw reads of `isActive`/`peerType`/`timestamp` while `HeartbeatReceived` writes them under `updateMutex`) and snapshotted it under `RLock`. That's the first thing surfaced now that `-race` is usable after #112.

### Scope check: this cannot affect consensus

Verified, since it's the natural worry when peer-type semantics change:

- The diff touches **no** file under `core/consensus/`, `sharding/`, `headerCheck/` or `process/block/`.
- `isValidatorPeerType` is package-private; all three call sites are inside `node/heartbeat/process/`.
- **`PeerShardMapper` does not consume `PeerTypeProvider`.** It calls `nodesCoordinator.GetValidatorWithPublicKey` directly, and that is untouched here. So the antiflood chain (`PeerShardMapper` → `IsOriginatorElectedForTopic` → interceptor) reads the coordinator, not this cache.
- `ComputeForPubKey` has exactly two production consumers (`monitor.go:364`, `sender.go:189`) and `GetAllPeerTypeInfos` one (`monitor.go:432`) — all heartbeat.

Everything below is metrics, API payload, memory and lock time. Nothing can change who signs, who is admitted, or how a consensus group is computed.

### Needs a decision from you — not a code defect

Two of these are correct implementations of #114 whose *consequences* need an explicit call:

| Where | Decision |
|---|---|
| `monitor.go:493` | Waiting entries are now permanently unprunable, so `heartbeatMessages` and `/node/heartbeatstatus` grow with the **registered** validator set rather than the consensus set. Fine at tens; worth bounding at hundreds. |
| `heartbeatMessageInfo.go:216` | `klv_live_validator_nodes` silently changes meaning. An existing alert threshold tuned to the old semantics would stop firing during a real validator outage. |

Neither is wrong. Both are irreversible-in-practice once dashboards and alerts adapt, so they are better decided now than discovered later.

### In scope — worth fixing before merge

- `sender.go:181` — waiting nodes now render a **blank** `Consensus slot state` in termui instead of `N/A (observer)`
- `monitor_test.go:597` — deferred `Close` lets the refresher goroutine mask the regression the test exists to catch (the two sibling tests close eagerly — asymmetric)
- `peerTypeProvider_test.go:358` — map-index deref panics instead of failing, aborting the package binary
- `heartbeatMessageInfo.go:209` — the new doc comment lists its callers and is already incomplete in the same commit

### Out of scope — recorded, not merge conditions

- `monitor.go:379` — `computeAllHeartbeatMessages` takes four *exclusive* locks per entry for pure reads. Pre-existing, but this PR enlarges the map it walks. Ironically this PR fixes exactly this pattern one function away.
- `peerTypeProvider.go:127` — third copy of the getter/log/`computePeerType` block. The duplication predates this diff; the sibling `validatorsProvider.go:219-232` has it too.

Commits are trailer-free. Requesting changes on the in-scope items; the two decisions are yours.

node/heartbeat/process/monitor.go
MathijsBok: Decision made: accepted without a cap, grounded in data. Measured on mainnet today via the public validator list: waiting=1, elected=21, eligible=103 (plus inactive=52 and jailed=23, which never enter this cache). The heartbeat map grows from 124 to 125 entries, the very bottom of your 'at tens, ignore' band, and the registration economics (min self-stake per waiting entry) keep the set expensive to inflate.

Recorded criterion for revisiting: if the waiting list ever approaches hundreds, the cap needs a design round rather than a patch, because the sketch (evict never-broadcast waiting entries older than N x hide interval) reintroduces the churn cycle for exactly those entries: `computeInactiveHeartbeatMessages` re-materialises every evicted key from `GetAllPeerTypeInfos` on the next tick, so a real cap also needs eviction memory or materialisation filtering. Documented in the PR description (Review round, item 5).

node/heartbeat/process/heartbeatMessageInfo.go
MathijsBok: Decision made: both options, in 4101d00. `klv_live_validator_nodes` keeps the widened meaning (waiting counts), and a new gauge `klv_live_consensus_validator_nodes` counts only active elected+eligible, restoring the 'counts nodes that can actually sign' invariant as its own queryable series. Implementation: `isConsensusCapablePeerType` (elected, eligible) next to `isValidatorPeerType`, which now composes with it so the two definitions cannot drift; both counters are computed in the same `computeAllHeartbeatMessages` pass and published together. Your outage scenario now pages: 15 of 21 elected going offline drops `klv_live_consensus_validator_nodes` regardless of how many waiting nodes keep heartbeating.

Pinned by the extended `TestMonitor_ActiveWaitingNodeCountsAsLiveValidator`: active waiting+eligible+observer gives live=2, consensus=1, connected=3. The release-note block in the PR description points alert owners at the new series.

node/heartbeat/process/sender.go
MathijsBok: Fixed in 4101d00. `widgetsRender.go` now reads `GetPeerType` next to `GetNodeType` in the slot-state branch: observer keeps `N/A (observer)`, a waiting peer gets `N/A (waiting)`, and everything else falls through to `GetConsensusSlotState` as before. I chose the render-side fix over a slot-state metric default because waiting is a peer-type fact the renderer already has, and it labels the row honestly instead of borrowing the observer label. The termuiRenders package tests pass; the `klv_build_info{node_type="validator"}` relabel you noted is documented in the PR description's release note.

node/heartbeat/process/monitor_test.go
MathijsBok: Fixed in 4101d00. The monitor is closed eagerly right after construction, before the Cleanup/assert rounds, same as both siblings; the deferred variant is gone. Verified that the masking is actually eliminated by mutation: with `core.WaitingList` temporarily dropped from `isValidatorPeerType`, the test now fails at the post-Cleanup count assertion, where the deferred-close version could let the refresher re-create the entry first. Good catch on the asymmetry tell.

core/process/peer/peerTypeProvider_test.go
MathijsBok: Fixed in 4101d00: `require.Contains` for all three keys before the `.pType` dereferences, so the regression you sketched (the `ownerKey` flag flipped, cache keyed by owner address) fails as a one-line missing-key assertion instead of a nil-pointer panic aborting the package binary. The same dereference pattern in the older createCache tests predates this PR and was left untouched to keep the diff scoped.

node/heartbeat/process/heartbeatMessageInfo.go
MathijsBok: Fixed in 4101d00. The comment now states the rule without enumerating callers: every consumer that classifies a peer as a validator (metrics counting, Cleanup shielding, node-type reporting) must use the predicate instead of comparing peer types itself. The new `isConsensusCapablePeerType` added in the same commit gets the same treatment (it documents which metric it feeds, not who calls it). You were right that the list was stale in the very commit that added it; the third caller was `Sender.updateMetrics`.

node/heartbeat/process/monitor.go
MathijsBok: Recorded as #117, with your analysis carried over and updated for the post-review state of the loop: after 4101d00 it is three exclusive acquisitions (`GetIsActive`, `ComputeActive`, `GetIsActive`) plus two read acquisitions (`GetIsValidator`, `GetIsConsensusCapable`) per entry under the global write lock, and the inconsistent-snapshot window you described (`HeartbeatReceived` landing between `ComputeActive` and the classification reads) is quoted in the issue. The fix shape there is the one you suggested: a single per-entry snapshot like `shouldSkipValidator` now uses, plus moving `GetIsActive` to `RLock`. `GetIsValidator` and the new `GetIsConsensusCapable` were already switched to `RLock` in this PR since those lines were being touched anyway.

core/process/peer/peerTypeProvider.go
MathijsBok: Recorded as #118, including the `validatorsProvider.go:219-232` sibling you pointed at, the copy-paste failure mode (an un-updated `computePeerType` constant silently mislabeling every key of a new list), and the constraint that the table refactor must preserve the fail-fast contract exactly. Left out of this PR per your nudge framing: the diff has been through four review rounds and the refactor is mechanical enough to stand alone.

node/heartbeat/process/heartbeatMessageInfo.go
fbsobreira: Confirming, though the question is superseded by your own later work — you asked at 19:57, my review landed at 23:10, and `4101d00` then implemented the separate gauge. So:

**No, combined-only was not the direction — but what is on the branch now is.** `klv_live_validator_nodes` keeping the widened meaning *plus* `klv_live_consensus_validator_nodes` restoring the can-actually-sign invariant is exactly right, and better than either option alone. Composing the predicates (`isValidatorPeerType = isConsensusCapablePeerType || waiting`) is what makes it safe: the two series cannot drift, which was the real risk. No follow-up split needed.

**Your `InitMetrics` finding is a good catch and it corrects my framing.** I verified it: `cmd/node/startup.go:634` passes `core.NodeTypeValidator` unconditionally, so every node starts reporting `klv_node_type=validator` regardless of role.

That means the pre-PR behaviour for a waiting node was not a stable `observer` baseline — it was `validator` at startup, then a **flip to `observer`** once `Sender.updateMetrics` ran on the first heartbeat. This change removes that flip. So on the `klv_node_type` axis specifically, this PR makes a waiting node *more* consistent, not less.

That does not change the `klv_live_validator_nodes` baseline-shift warning — that one is real, it is about the gauge rather than the label, and the release-note block covers it. But it is worth having both facts recorded together, because "waiting nodes used to report observer" is the intuitive assumption and it turns out to be wrong for the first few seconds of every run.

On the possible waiting-nodes gauge: agreed it is a reasonable follow-up, not needed now. With elected+eligible and elected-only both queryable, waiting is derivable as the difference — worth adding only if operators want to alert on waiting capacity directly.

Nothing outstanding from my side on this thread.

fbsobreira (approved):

Issue #124, opened 2026-08-10
numToStay promotion is all-or-nothing: no partial backfill when the leaving list is smaller than the consensus deficit


MathijsBok:
## Context

Observed by two independent review passes during the pre-push audit of the #116 branch, which added tests around this logic without changing it.

## Problem

In `computeNodesConfigFromList` (sharding/nodesCoordinator.go:938-944) the consensus backfill promotes leaving (jailed) validators into the eligible list only when `len(leavingList) >= numToStay`. If the leaving list has fewer members than the shortfall, none are promoted at all, and elected+eligible can stay below `consensusGroupSize` even though partial backfill was possible.

```go
numToStay := ihgs.consensusGroupSize - (len(electedList) + len(eligibleList))
if numToStay > 0 {
    if len(leavingList) >= numToStay {
        eligibleList = append(eligibleList, leavingList[:numToStay]...)
        leavingList = leavingList[numToStay:]
    }
}
```

This may be intentional (promote only when the deficit can be fully covered), but the asymmetry is undocumented and the partial-shortfall branch has no test. TestNodesCoordinator_computeNodesConfigFromList_JailedPromotedWhenConsensusShort pins only the >= path.

## Task

Decide whether partial backfill is wanted. Either implement it (promote min(numToStay, len(leavingList))) or document the all-or-nothing rule with a short comment and pin the below-threshold branch with a test.

## References

- sharding/nodesCoordinator.go:938-944
- #116 (why promoted validators leave the leaving list matters for peer-type reporting)

MathijsBok:
Adjacent defect in the same six lines split off as #133: the promotion slices `leavingList` at sharding/nodesCoordinator.go:946 while the sort only happens at :951, so *which* validators get promoted depends on the input order, independently of the all-or-nothing threshold discussed here.

The two are separable but any fix touches the other's lines, so they are probably best done in one PR.

Issue #123, opened 2026-08-10
selectValidators bounds guard off-by-one: selected index equal to elected list length passes the guard and panics


MathijsBok:
## Context

Found during the pre-push audit of the #116 branch. Pre-existing behavior, untouched by that PR.

## Problem

The bounds guard uses a strict greater-than (sharding/nodesCoordinator.go:599):

```go
if slices.Max(selectedIndexes) > electedListLen || len(electedList) == 0 {
```

An index exactly equal to `len(electedList)` passes the guard and then panics at the subsequent `electedList[selectedIndexes[i]]` access (index out of range). The guard should be `>=`.

## Task

Change the comparison to `>=` and add a regression test with a selected index equal to the list length, asserting the error path instead of a panic.

## References

- sharding/nodesCoordinator.go:599

Issue #122, opened 2026-08-10
nodesCoordinator.validatorsInfo map is read and written without synchronization (latent data race)


MathijsBok:
## Context

Flagged twice by independent review passes during the pre-push audit of the #116 branch. Pre-existing behavior, untouched by that PR.

## Problem

`ihgs.validatorsInfo` is written by `SetEpochValidatorsInfo` (sharding/nodesCoordinator.go:714, including its old-epoch delete loop) and read by `getEpochValidatorsInfo` (:740), called from `EpochStartPrepare` on the epoch-start notifier goroutine, with no mutex guarding the map. `SetEpochValidatorsInfo` can be called from processing goroutines while the notifier fans out, so concurrent map read/write is reachable and would panic the node (CWE-362).

The race detector has not tripped on it in unit tests because tests drive both sides sequentially.

## Task

Guard the map with a mutex (a dedicated one, or document why an existing coordinator lock is the right scope), and add a concurrent regression test (SetEpochValidatorsInfo racing EpochStartPrepare under -race).

## References

- sharding/nodesCoordinator.go:714 (writer), :740 (reader), EpochStartPrepare call site

Issue #121, opened 2026-08-10
computeNodesConfigFromList silently drops validators with List observer or leaving (no default case)


MathijsBok:
> [!NOTE]
> **Rewritten on 2026-08-13** after a full re-trace against `develop` @ `e48c2c2` including the open PRs. The original body named three classes, of which one (`inactive`) was already handled and one (`revoked`) is structurally unreachable, and it did not name the one class that actually matters (`observer`). The original title and body also contradicted each other on `inactive`. The superseded framing and the full evidence for every correction are preserved in the [re-assessment comment](https://github.com/klever-io/klever-go/issues/121#issuecomment-5286174948). Everything below is verified line by line at that commit.
>
> Scope was reduced accordingly. Two more serious findings from the same re-trace were split off as #132 and #133.

## Context

Split off from PR follow-up work on #116 (jailed validators in the peer-type cache) and confirmed by two independent review passes on that branch. Re-scoped on 2026-08-13, see the note above.

## Problem

The switch in `computeNodesConfigFromList` (sharding/nodesCoordinator.go:928-940) has cases for eligible, elected, waiting, inactive (logged and dropped) and jailed (goes to the leaving list). It has no `default`.

Two enum values therefore hit no case at all and vanish from every coordinator list without so much as a log line:

- **`observer`** (`state.List_observer`)
- **`leaving`** (`state.List_leaving`)

Both are latent today: no non-test code path writes either value. The complete set of `SetList`/`SetListAndIndex` call sites writes only `inactive`, `waiting`, `eligible`, `elected` and `jailed`, and the one dynamic route (`state.List(state.List_value[peerType])` at core/kapp/validators/validators.go:1039) is fed only with `elected`, `eligible` and `waiting`.

For `leaving` the codebase is nevertheless already prepared in two places, neither of them exercised: the reader in core/validatorInfo/validatorInfoUtils.go:20, and its single caller at core/kapp/validators/peersUpdate.go:36, where rating processing deliberately includes validators that were active while on the leaving list.

## Why `observer` is the class that matters

**PR #86 (validator node version attestation with observer demotion) makes `observer` a live producer, and it depends on this exact drop as its exclusion mechanism.** Its description states that demoted validators are *"excluded from election (`computeNodesConfigFromList` ignores observers)"*, and its diff writes `state.List_observer` via `resolveVersionEnforcedList`, with tests asserting demotion from both `eligible` and `elected`.

[...]

PR #125, opened 2026-08-10
fix: report jailed validators in peer-type cache and heartbeat metrics

### LINKED ISSUE #116: Jailed and leaving validators report peer type observer (leaving list is never queried)
## Context

Split off as a documented follow-up from PR #115 (waiting-list validators in heartbeat peer-type metrics, issue #114). Same problem class as #114, deliberately left out of that PR's scope.

## Problem

`PeerTypeProvider.createNewCache` seeds the peer-type cache from `GetAllWaitingValidatorsKeys`, `GetAllElectedValidatorsKeys` and `GetAllEligibleValidatorsKeys` (since PR #115; before that only the last two). The nodes coordinator's `computeNodesConfigFromList` (sharding/nodesCoordinator.go) also builds a `leavingList`: validators with `List == jailed` are placed there (inactive ones are dropped entirely). That list is never queried by the peer-type cache, so a jailed validator's key is absent from the cache and `ComputeForPubKey` falls back to `observer`.

Consequences for a jailed validator's node:
- `klv_node_type` reports `observer` and `klv_peer_type` reports `observer` instead of something truthful (`jailed`/`leaving`).
- The heartbeat monitor treats the entry as an inactive observer, so it disappears from `/node/heartbeatstatus` after `HideInactiveValidatorIntervalInSec`.

Nuance: leaving-list validators that are promoted back into the eligible list to fill the consensus size (`numToStay` logic in `computeNodesConfigFromList`) do appear as `eligible`, so this only affects the non-promoted remainder.

## Why this needs an explicit design decision first

PR #115 established that all three metric consumers must move together (the lesson from PR #110): cache seeding, `isValidatorPeerType` (feeds `klv_live_validator_nodes` and the sender's `klv_node_type`), and `Monitor.shouldSkipValidator` (Cleanup shielding). For jailed validators the right answers are less obvious than for waiting:
- Should a jailed validator count toward `klv_live_validator_nodes`? (Probably not: it is being punished, not queued for work.)
- Should its node self-report `klv_node_type=validator`? (Arguable either way.)
- Should its heartbeat entry be shielded from Cleanup so operators can see it? (Probably yes, visibility is the point.)

Deciding those three explicitly, then wiring the seeding (`core.JailedList` or `core.LeavingList` as the cached type), the predicate, and the shielding consistently, is the scope of this issue.

## References

- PR #115 follow-up note; issue #114 (waiting-list variant of the same class); PR #110 review thread (why partial changes to this metric surface are worse than none).

### LINKED ISSUE #118: Deduplicate the getter/log/computePeerType blocks in PeerTypeProvider.createNewCache (and the validatorsProvider sibling)
## Context

Recorded from the PR #115 review (fbsobreira, inline comment on peerTypeProvider.go:127): `createNewCache` now contains three near-identical getter / warn-log / `computePeerType` blocks (waiting, elected, eligible) that differ by three tokens. The duplication predates PR #115 (it went from two copies to three there), and `core/process/peer/validatorsProvider.go:219-232` carries the same shape.

## Risk being addressed

Adding a fourth list is a copy-paste where the pasted `computePeerType(newCache, nodesMapX, core.<Type>)` constant or the log string is easy to leave un-updated. The compiler catches none of it, and the scenario-table test is itself copy-pasted, so every key of a new list could silently report the wrong `pType` through `ComputeForPubKey` and `GetAllPeerTypeInfos`.

## Fix shape

A table of `{name, peerType, getter}` with one loop over it, preserving the existing semantics exactly: fail-fast on the first getter error (return `nil, false` so `UpdateCache` keeps the previous cache), `log.Warn` with the list name and error value, and seeding order waiting, elected, eligible (later writes win). Apply the same treatment to the validatorsProvider sibling in the same PR.

## References

- PR #115 review thread on `createNewCache` (fbsobreira): explicitly a nudge, not a merge condition
- Issue #114 / PR #110 for the fail-fast contract that must be preserved

MathijsBok:
## Summary

Closes #116. Closes #118.

A jailed validator's key never reached the heartbeat peer-type cache, so its node reported `klv_node_type=observer` and `klv_peer_type=observer`, and its heartbeat entry disappeared from `/node/heartbeatstatus` after `HideInactiveValidatorIntervalInSec`. Root cause: the nodes coordinator computed its leaving list (validators with `List == jailed` in `computeNodesConfigFromList`) but then discarded it. `EpochStartPrepare` never handed it to `SetNodes`, `epochNodesConfig.leavingList` was always empty, and the registry did not persist it, so the list was unreachable for the peer-type cache in every scenario (live, epoch start, restart).

## What changed

**Commit 1 (5cc8f9d, #118)**: `PeerTypeProvider.createNewCache` and `validatorsProvider.createNewCache` are now table-driven (`{name, peerType, getter}` plus one loop), so adding a list is one data row instead of an error-prone copy-paste. Semantics preserved exactly: the peerTypeProvider keeps fail-fast (any getter error keeps the previous cache, `log.Warn` with the error), the validatorsProvider keeps its soft-fail overlay. A new subtest pins the soft-fail contract (`failing getter keeps trie-based cache and remaining overlays`), and the empty epoch-start prepare handler got the same explanatory comment as its sibling (sonar go:S1186).

**Commit 2 (49067cc, #116)**: the coordinator now stores, persists and exposes the leaving list, and the peer-type surface consumes it:

- `SetNodes` gained a `leaving` parameter and stores it on the epoch config; `EpochStartPrepare` passes `newNodesConfig.leavingList` (sharding/nodesCoordinator.go).
- The registry serializes it as `leavingValidators` and `LoadState`/the bootstrap path in `cmd/node/node.go` restore it.
- New getter `GetAllLeavingValidatorsKeys(epoch, ownerKey)` on `sharding.NodesCoordinator`; all implementers updated, and the dead `GetAllLeavingValidatorsPublicKeys` mock lookalikes (different name and signature, zero callers) were removed.
- `PeerTypeProvider.createNewCache` seeds the leaving list first, labeled `core.JailedList`; working lists win on any overlap (the numToStay promotion keeps the lists a partition in production).
- New predicate tier `isRegisteredValidatorPeerType` (node/heartbeat/process/heartbeatMessageInfo.go) drives Cleanup shielding (`monitor.go` `shouldSkipValidator`) and `klv_node_type` self-reporting (`sender.go` `updateMetrics`). The gauges are unchanged.

**Commit 3 (015b05f)**: fixes from the first review round: the four `GetAllXValidatorsKeys` getters were consolidated behind one `getAllValidatorsKeys` helper (flagged independently by two reviewers; the fourth copy also risked the SonarQube duplicated-lines gate on new code), the ownerKey test leg now uses real validators with distinct owner addresses (the validator mock drops its owner argument, making the old assertion vacuous), a discarded constructor error in a new test is now asserted, the mock parameter `includeLeaving` was renamed to `ownerKey`, and a new test pins that a malformed leaving validator in the registry fails `LoadState` with `ErrNilPubKey`.

**Commit 4 (bd6804b)**: gate-review nits: the overlay `log.Debug` now includes the error value (errors are never silently discarded), and the precedence comment in the provider scenario test moved to the assertion that actually proves it.

## Design decisions

1. Jailed does **not** count toward `klv_live_validator_nodes`: the gauge remains the working set (elected, eligible, waiting).
2. A jailed validator's node self-reports `klv_node_type=validator`; the punished state is visible as `klv_peer_type=jailed`.
3. Jailed heartbeat entries are shielded from cleanup so operators keep seeing the node; entries are even created from the cache when the node never sent a heartbeat.
4. The reported label is `jailed`, not `leaving`: the coordinator fills the leaving list exclusively from `List == jailed`, and `/validator/statistics` already reports `jailed` from the trie.

The predicate tiers are now: consensus-capable (elected, eligible) inside working validators (plus waiting) inside registered validators (plus jailed), one predicate per tier, with the consumers documented at the definitions.

## Compatibility

- Registry JSON gains `leavingValidators`. Old snapshot on a new binary: field absent, restores an empty list, converges at the next epoch start (pinned by `TestNodesCoordinator_LoadStateWithoutLeavingFieldIsNilSafe`). New snapshot on an old binary (rollback): unknown field is ignored. A malformed leaving entry fails the restore with `ErrNilPubKey` instead of being silently dropped (pinned by `TestNodesCoordinator_LoadStateWithMalformedLeavingValidatorFails`).
- The heartbeat wire format is unchanged and peer types are computed node-locally, so mixed-version networks only differ in their own reporting.

## Operational notes

- `klv_peer_type` gains the value `jailed`. Dashboards or alerts with an enum assumption (elected/eligible/waiting/observer) should be updated; alerts that used `observer` as a proxy for "my validator is broken" change meaning for jailed nodes.
- `klv_live_validator_nodes` now explicitly excludes jailed (doc comment updated); a live jailed node still counts in `klv_connected_nodes`.

## Tests

- Coordinator: `SetNodes` stores the list and the getter returns it (pubkeys and owner addresses, unknown epoch errors), save/load round-trip, legacy registry without the field, malformed leaving entry fails restore, `EpochStartPrepare` stores the computed list end to end, `computeNodesConfigFromList` puts jailed in the leaving list and the numToStay promotion moves the promoted key out of it.
- Providers: scenario tables extended with leaving/jailed rows (fail-fast per getter, partition resolution, keep-previous-cache, precedence on overlap); soft-fail overlay pinned for the validatorsProvider.
- Heartbeat: predicate table test pins jailed as registered and the raw string `leaving` as deliberately not registered; jailed is excluded from both gauges; monitor lifecycle tests pin shielded survival across refresh/cleanup rounds, gauge exclusion for an active jailed node, and demotion plus cleanup once the key leaves the lists; sender table rows pin `jailed` reporting `validator` and `leaving` falling back to `observer`.

## Verification

- `go build ./...`, `go vet ./...`, `gofmt` clean; `golangci-lint` (module mode) reports zero issues on changed files.
- All affected packages green, `-race` clean on `node/heartbeat/...`; full `go test ./...`: 247 packages ok. The two failing packages are baseline: `data/retriever/txpool/memorytests` fails identically on clean develop, and `network/p2p/libp2p` only timed out under full parallel load while passing standalone on both develop and this branch.
- SonarQube API: zero open issues on changed lines (two pre-existing on touched files: go:S107 at cmd/node/node.go:34 and a go:S1135 TODO at nodesCoordinatorStub.go:9).
- Review: seven agent passes (three code-quality, two security, one Go-specific, one metrics); every actionable finding was fixed in commits 3 and 4; the final independent full-diff pass reported no findings.

## Follow-ups (tracked)

- #121: `computeNodesConfigFromList` silently drops validators with `List` leaving, inactive or revoked; they keep reporting observer.
- #122: `validatorsInfo` map is accessed without synchronization (latent data race).
- #123: `selectValidators` bounds guard off-by-one can panic on an index equal to the elected list length.
- #124: numToStay promotion is all-or-nothing; no partial backfill when the leaving list is smaller than the deficit.

Out of scope, tracked separately: the pre-bootstrap epoch seeding window (#113, applies to jailed exactly as it already did to waiting) and the heartbeat monitor concurrency work (#117, #120).

## Summary

- Persist and restore jailed validators in the coordinator’s leaving list.
- Expose leaving-validator keys through `NodesCoordinator`.
- Seed the peer-type cache with `JailedList` entries.
- Keep working-validator classifications higher priority on overlap.
- Keep jailed validators out of `klv_live_validator_nodes`.
- Report jailed validators as `klv_node_type=validator` and `klv_peer_type=jailed`.
- Retain jailed validators in heartbeat status and protect them from cleanup.
- Remove jailed validators from heartbeat tracking after they leave the jailed list.
- Deduplicate validator-list getter and classification logic.
- Preserve fail-fast cache updates in `PeerTypeProvider`.
- Preserve soft-fail overlay behavior in `validatorsProvider`.
- Add compatibility handling for persisted state without a leaving list.

## Blockchain impact

- **Consensus and state management:** The coordinator now carries jailed-validator data across epochs and persistence boundaries. This prevents loss of validator status during restart or epoch changes.
- **Networking and heartbeat:** Jailed validators remain identifiable and visible without being counted as live validators. This improves monitoring and prevents premature cleanup.
- **Transaction processing and KVM:** No direct changes.
- **Node stability:** Cache construction retains the previous cache when `PeerTypeProvider` getters fail. `validatorsProvider` continues to apply available overlays when a getter fails.
- **Data integrity:** Legacy persisted state remains loadable. Malformed leaving-validator data returns an error instead of being silently accepted.
- **Concurrency:** No explicit concurrency model changes are introduced.

### DIFF
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -373,7 +373,15 @@ func createNodesCoordinator(
 				return nil, err
 			}
 
-			err = nodesCoordinator.SetNodes(electedValidators, eligibleValidators, waitingValidators, currentEpoch-1)
+			// registries saved before the leaving list was persisted have no
+			// LeavingValidators field; the list is then simply empty
+			leaving := prevEpochsConfig.LeavingValidators
+			leavingValidators, err := sharding.SerializableValidatorsToValidators(leaving)
+			if err != nil {
+				return nil, err
+			}
+
+			err = nodesCoordinator.SetNodes(electedValidators, eligibleValidators, waitingValidators, leavingValidators, currentEpoch-1)
 			if err != nil {
 				return nil, err
 			}
--- a/common/mock/nodesCoordinatorMock.go
+++ b/common/mock/nodesCoordinatorMock.go
@@ -24,7 +24,8 @@ type NodesCoordinatorMock struct {
 	GetAllEligibleValidatorsKeysWithEpochCalled func(uint32, bool) ([][]byte, error)
 	GetAllWaitingValidatorsKeysCalled           func() ([][]byte, error)
 	GetAllWaitingValidatorsKeysWithEpochCalled  func(uint32, bool) ([][]byte, error)
-	GetAllLeavingValidatorsPublicKeysCalled     func() ([][]byte, error)
+	GetAllLeavingValidatorsKeysCalled           func() ([][]byte, error)
+	GetAllLeavingValidatorsKeysWithEpochCalled  func(uint32, bool) ([][]byte, error)
 	CheckValidatorSlotCalled                    func(epoch uint32, slotIndex int64, pubkey []byte) bool
 	ConsensusGroupSizeCalled                    func() int
 	LoadValidatorsCalled                        func(validators []*state.ValidatorInfo) error
@@ -90,6 +91,17 @@ func (ncm *NodesCoordinatorMock) GetAllWaitingValidatorsKeys(epoch uint32, inclu
 	}
 	if ncm.GetAllWaitingValidatorsKeysCalled != nil {
 		return ncm.GetAllWaitingValidatorsKeysCalled()
+	}
+	return nil, nil
+}
+
+// GetAllLeavingValidatorsKeys -
+func (ncm *NodesCoordinatorMock) GetAllLeavingValidatorsKeys(epoch uint32, includeLeaving bool) ([][]byte, error) {
+	if ncm.GetAllLeavingValidatorsKeysWithEpochCalled != nil {
+		return ncm.GetAllLeavingValidatorsKeysWithEpochCalled(epoch, includeLeaving)
+	}
+	if ncm.GetAllLeavingValidatorsKeysCalled != nil {
+		return ncm.GetAllLeavingValidatorsKeysCalled()
 	}
 	return nil, nil
 }
@@ -223,11 +235,6 @@ func (ncm *NodesCoordinatorMock) GetValidatorWithPublicKey(publicKey []byte) (sh
 	}
 
 	return nil, sharding.ErrValidatorNotFound
-}
-
-// GetAllLeavingValidatorsPublicKeys -
-func (ncm *NodesCoordinatorMock) GetAllLeavingValidatorsPublicKeys(_ uint32) ([][]byte, error) {
-	return nil, nil
 }
 
 // GetOwnPublicKey -
--- a/common/mock/nodesCoordinatorMock.go
+++ b/common/mock/nodesCoordinatorMock.go
@@ -96,9 +96,9 @@ func (ncm *NodesCoordinatorMock) GetAllWaitingValidatorsKeys(epoch uint32, inclu
 }
 
 // GetAllLeavingValidatorsKeys -
-func (ncm *NodesCoordinatorMock) GetAllLeavingValidatorsKeys(epoch uint32, includeLeaving bool) ([][]byte, error) {
+func (ncm *NodesCoordinatorMock) GetAllLeavingValidatorsKeys(epoch uint32, ownerKey bool) ([][]byte, error) {
 	if ncm.GetAllLeavingValidatorsKeysWithEpochCalled != nil {
-		return ncm.GetAllLeavingValidatorsKeysWithEpochCalled(epoch, includeLeaving)
+		return ncm.GetAllLeavingValidatorsKeysWithEpochCalled(epoch, ownerKey)
 	}
 	if ncm.GetAllLeavingValidatorsKeysCalled != nil {
 		return ncm.GetAllLeavingValidatorsKeysCalled()
--- a/core/bootstrap/disabled/disabledNodesCoordinator.go
+++ b/core/bootstrap/disabled/disabledNodesCoordinator.go
@@ -44,6 +44,11 @@ func (n *nodesCoordinator) GetAllWaitingValidatorsKeys(_ uint32, _ bool) ([][]by
 	return nil, nil
 }
 
+// GetAllLeavingValidatorsKeys -
+func (n *nodesCoordinator) GetAllLeavingValidatorsKeys(_ uint32, _ bool) ([][]byte, error) {
+	return nil, nil
+}
+
 // CheckValidatorSlot -
 func (n *nodesCoordinator) CheckValidatorSlot(epoch uint32, slotIndex int64, pubkey []byte) bool {
 	return false
--- a/core/metrics.go
+++ b/core/metrics.go
@@ -159,7 +159,8 @@ const MetricCountConsensus = "klv_count_consensus"
 // MetricNodeType is the metric for monitoring the type of the node
 const MetricNodeType = "klv_node_type"
 
-// MetricPeerType is the metric which tells the peer's type (in eligible list, in waiting list, or observer)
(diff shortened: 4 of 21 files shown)


MathijsBok:
Superseded by #126: identical commits (head bd6804b), reopened from the upstream branch now that write access is in place, so the secret-dependent CI jobs (sonarqube, notify, result-check) actually run. All review context (audit summary, follow-ups #121 to #124) carries over in the new PR body.

Issue #94, opened 2026-07-20
CI lint step never fails the build (golangci-lint run ... || true)


Test0rMaik:
`.github/workflows/go-setup-lint.yaml:71` runs:
```
golangci-lint run --output.checkstyle.path=golangci-report.xml || true
```
The `|| true` means lint failures never fail the CI job — results only flow into the checkstyle report (presumably for the SonarQube dashboard) as an informational artifact, not a merge gate. `.golangci.yml` has a real rule set configured (`unused`, `gosec`, `govet`, `staticcheck`, `ineffassign`, `errcheck`, `errorlint`), so this isn't "no linting," it's "linting that can't block anything."

Practical effect: a PR can introduce new lint violations (including `gosec` findings) and CI stays green — nobody would notice from the check status alone; it'd only show up if someone actively opens the SonarQube dashboard.

Not proposing a specific fix here since removing `|| true` could turn CI red across the board if there's existing lint debt (worth checking current lint-clean status first, which needs a golangci-lint v2 binary locally since `.golangci.yml` uses the v2 config schema — noting this because it tripped us up locally). Raising as a policy question for the team: is this intentional (e.g. staged rollout, known debt) or worth tightening?

fbsobreira:
Thanks for raising this. The `|| true` is more or less expected: lint isn't meant to block the build directly. The checkstyle report is uploaded as an artifact and passed to SonarQube (`-Dsonar.go.golangci-lint.reportPaths=golangci-report.xml` in `pr-qa-sec.yaml`), where the findings are imported as external issues. The SonarQube quality gate is what blocks merges on PRs: `sonarqube-quality-gate-action` runs with `continue-on-error` only for non-PR events. SonarQube judges a PR on its new code, so new lint violations can fail the gate, while existing lint debt doesn't turn CI red.

You did point to a real gap, though. `|| true` also hides golangci-lint **failing to run**: config errors, typecheck failures, timeouts, or a `latest` release that breaks the v2 config schema. In that case the report is empty or incomplete, SonarQube sees zero lint issues, and the quality gate passes, so enforcement quietly stops working.

We're tracking a follow-up internally (KLC-2703):

- accept only golangci-lint's "issues found" exit code (`|| [ $? -eq 1 ]`) and fail on any other error
- pin the golangci-lint version instead of `latest`
- confirm the quality gate settings and required status checks on our side

Also a good note about the v2 config schema tripping up local runs. Pinning the version should make that easier to reproduce too.

Issue #90, opened 2026-07-20
Fallback node stalls at an epoch boundary and reports itself synchronized (new-epoch headers rejected until the epoch-start block commits)


MathijsBok:
> [!NOTE]
> **Rewritten on 2026-08-12** after a full re-trace against `develop` @ `e48c2c2`. The original analysis in this body was wrong in four of its core claims and led to a fix (PR #103) that was closed without merging. The superseded reasoning is preserved in the comment history: the [original fix-status comment](https://github.com/klever-io/klever-go/issues/90#issuecomment-5023607960) and the [re-assessment](https://github.com/klever-io/klever-go/issues/90#issuecomment-5256076475). Everything below is verified line by line at that commit.

## What was observed in production

A continuously running validator fallback (`--redundancy-level=1`, backup mode, alongside an elected main node) stopped following the chain per slot at an epoch switch. No restart was involved.

After the boundary it advanced only in bursts of a varying number of blocks, always below 20, which is distinct from the fixed ~20-block bursts of the known sync-is-stuck path.

The elected set changed at that boundary (the main node's key moved from elected to eligible).

Note on evidence: the burst sizes and the timing are observations. The original claim that the node "flips to not-synchronized" was an inference, and it is the opposite of what the code does. See below.

### 1. The rejection window

`nodesConfig[newEpoch]` is created only when the epoch-start block N is **committed**: `CommitBlock` -> `commitEpochStart` (`core/process/block/block.go:645`) -> `trigger.SetProcessed` -> `NotifyAllPrepare` (`eventNotifier/epochStart/trigger.go:242`) -> `EpochStartPrepare` -> `SetNodes` (`sharding/nodesCoordinator.go:795`). Until then, `ComputeConsensusGroup` returns `ErrEpochNodesConfigDoesNotExist` (`sharding/nodesCoordinator.go:517-527`).

The epoch-start block N itself is unaffected: both `VerifySignature` (`core/process/headerCheck/headerSignatureVerify.go:104-108`) and `getLeader` (`:288-292`) decrement the epoch for `IsEpochStart` headers, so N validates under the previous epoch. Block N+1 is not an epoch-start block, so it validates under the new epoch. That is the gap.

During the window **all three** paths by which a node can learn about a new block are blocked:

[...]

PR #130, opened 2026-08-13
fix: request headers when a node falsely reports itself synchronized

MathijsBok:
## The problem

A node whose header intake is blocked keeps a frozen fork detector. `computeNodeState` derives `hasLastBlock` from `probableHighestNonce` (`core/process/sync/baseSync.go:304`), and that counter only moves for headers that were *accepted*. So a node that is rejecting every incoming header sits at a stale nonce, computes `hasLastBlock = true`, reports `NsSynchronized`, and writes `MetricIsSyncing = 0`. `syncBlock` then returns early (`:605-607`) and the node stops asking for anything.

Issue #90 reaches this through an epoch boundary: `nodesConfig[newEpoch]` is only built when the epoch-start block commits, so until then every new-epoch header fails validation on all three intake paths (gossip, self-requested headers, and the consensus topic). But nothing about the mechanism is epoch-specific.

## What this changes

`shouldTryToRequestHeaders` now derives the lag from the slot manager and the last committed block instead of from the fork detector, so it stays truthful in exactly that state.

**This replaces the previous `slotIndex % 20` trigger rather than adding to it, and the reason matters.** That trigger could never fire in this state: `baseForkDetector.isConsensusStuck` (`:598-617`) uses the same lag against the same threshold of 10 and fires on `process.SlotModulusTrigger` (5). Since 20 is a multiple of 5, every slot that satisfied the old modulus also produced a forced-rollback `ForkInfo{IsDetected: true, Nonce: MaxUint64, Hash: nil}`, which is exactly what `isForcedRollBackOneBlock` matches, and `shouldTryToRequestHeaders` short-circuits on that guard before reaching the modulus.

So the baseline behaviour was not a quiet idle. Every fifth slot the node ran `rollBackOneBlockForced`, reverting state and dropping out of consensus for that slot. Requesting on the four slots in between is what keeps the fork detector out of that loop: once a requested header lands, `isSyncing()` turns true and the next stuck check stands down. `SlotModulusTriggerWhenSyncIsStuck` had no other user and is removed, so nobody tunes a constant that does nothing.

`isNodeSynchronized` is deliberately untouched. It gates consensus participation through `initCurrentSlot` (`core/consensus/slot/bls/subslotStartSlot.go:102-106`), so making it slot-lag-aware would push every validator out of consensus during a genuine network-wide halt.

## Also in this change

**A second, higher-impact underflow.** `isConsensusStuck` had the identical unguarded `SafeI64ToU64(Index()) - lastCheckpoint().slot`. There a wrapped lag does not merely waste a header request: it clears the threshold and forces a rollback of a block that was just committed. `checkBlockBasicValidity` deliberately accepts headers one slot ahead of the local index, so a node whose clock trails its peers reaches this without any peer misbehaving.

**A warning at default log level.** Everything about this state was previously invisible: `MetricIsSyncing` reads 0, `GetNodeState` answers synchronized, and both relevant log lines are debug while the node default is info. A node that stalls and simply receives nothing produced no signal at all. The new line fires at most once per slot and both its operands are local (own slot index, own last committed block), so a peer cannot drive it.

**A dedicated debug line for the epoch-config rejection**, from both verification entry points in `core/process/headerCheck`. It stays at debug on purpose: an unauthenticated peer can reach that path with a header carrying an arbitrary epoch, because `isEpochCorrect` admits any `epoch >= trigger.Epoch()`. It names the epoch actually looked up as well as the header's own epoch, which differ for an epoch-start header, so it cannot send an operator to the wrong configuration. The header hash is deliberately absent: it is not computed at that point, and computing one would put a marshal plus a hash on an attacker-drivable path.

**`errors.Is` for the `ErrTimeIsOut` comparison** in `doJobOnSyncBlockFail`, which `errorlint` flags and which fails on wrapped errors.

## What this does not do

It does not remove the false `NsSynchronized` report. During the window the node still advertises synchronized and `MetricIsSyncing` still reads 0; only the request path and the new warning react. Detection latency is unchanged at 11 slots (~44 s at `slotInterval: 4000`); what changes is that past that threshold the node reacts on the next slot rather than on the next multiple of 20, and that it no longer sits in the forced-rollback loop while waiting.

It also does not change `isConsensusStuck`'s semantics beyond the underflow guard. During a genuine network-wide halt the forced-rollback loop still runs. Whether that is correct is an open question that needs the measurement the new diagnostics provide, not a guess.

## Tests

Every production change is pinned by a test that fails without it, verified by reverting each change in a scratch copy:

- `TestShouldTryToRequestHeaders_SyncedNodeReactsToSlotLagOnEverySlot` sweeps slot indices 1 to 40. Without the change it fails on 28 of them (11-19 and 21-39), the ones the old modulus missed.
- `TestShouldTryToRequestHeaders_GuardsShortCircuitBeforeSlotLag` pins that `BeforeGenesis` and both forced-rollback guards still run first, using a lag of 100 so a regression cannot hide.
- `TestSlotsSinceLastCommittedBlock` covers the current-header path, the genesis fallback, and the underflow guard.
- `TestRequestHeadersIfSyncIsStuck` pins the burst formula operators see in the logs: `min(MaxHeadersToRequestInAdvance, lag-1)` starting at the nonce after the last committed block, the cap at 20, and that the underflow guard prevents a burst for nonces that cannot exist.
- `TestMetaForkDetector_CheckForkNoForcedRollBackWhenCheckpointIsAheadOfSlotIndex` and its counterpart `...StillDetectsStuckConsensusWhenCheckpointIsBehind` cover both directions of the fork-detector guard, so it cannot silently disable the mechanism.
- `TestHeaderSigVerifier_EpochNodesConfigMissingIsSurfacedIntact` asserts the sentinel survives both entry points (the consensus worker's blacklist suppression keys off it) and, through a log observer, that the line is actually emitted, carries the fields needed to measure the window, names the previous epoch for an epoch-start header, and stays silent for an unrelated error.

## Verification

`gofmt` clean, `go build ./...` clean, `go vet` clean, `golangci-lint` reports 0 issues on both touched packages, `go test -race` green on both, and `go test ./core/... ./sharding/...` is 75 ok with no failures.

Changed-line coverage: `shouldTryToRequestHeaders`, `slotsSinceLastCommittedBlock` and `logIfEpochConfigMissing` at 100%, and the new fork-detector guard covered in both directions. One new line is uncovered, the `errors.Is` fix in `doJobOnSyncBlockFail`, a function at 0% coverage whose fixture is disproportionate to a one-line lint fix.

Two pre-existing conditions worth knowing, neither caused here: `golangci-lint` cannot run with the repo's own config because `.golangci.yml` sets `modules-download-mode: vendor` and there is no `vendor/` tree, so `--modules-download-mode=mod` was used; and CI runs the lint step as `golangci-lint run ... || true` (#94), so it cannot fail the build.

Refs #90

## Summary

- Fix header synchronization requests when stale fork-detector state reports false synchronization.
- Calculate synchronization lag from the slot manager and last committed block.
- Enforce one header-request burst and warning per slot.
- Exclude import-db mode from slot-lag requests.
- Prevent slot underflow during synchronization and consensus-stuck checks.
- Preserve existing synchronization and consensus-participation behavior.
- Detect wrapped `ErrTimeIsOut` errors with `errors.Is`.
- Add throttled warnings for stalled synchronized nodes and checkpoint-ahead conditions.
- Add debug logs for missing epoch configuration during header verification.
- Remove the obsolete `SlotModulusTriggerWhenSyncIsStuck` constant.
- Add regression tests for request guards, import mode, lag calculations, underflow handling, fork-detector behavior, request throttling, timeout errors, warning conditions, and epoch configuration errors.

## Impact

- **Consensus and networking:** Header recovery uses local slot lag and handles stale fork-detector state. Consensus-stuck detection avoids rollback decisions when the checkpoint is ahead of the local slot.
- **Node stability:** Underflow guards prevent incorrect recovery decisions. Throttled warnings improve diagnosis of stalled nodes.
- **Data integrity:** The change does not modify committed blockchain data, transaction processing, state management, or KVM behavior.
- **Concurrency and performance:** Atomic slot tracking limits request bursts and warning frequency. The change does not add multi-slot backoff, so recovery is not delayed.
- **Error handling:** Wrapped timeout errors remain detectable. Missing epoch configuration errors remain intact and include diagnostic logging.

### DIFF
--- a/core/process/headerCheck/headerSignatureVerify.go
+++ b/core/process/headerCheck/headerSignatureVerify.go
@@ -1,6 +1,7 @@
 package headerCheck
 
 import (
+	"errors"
 	"math/bits"
 
 	logger "github.com/klever-io/klever-go-logger"
@@ -113,6 +114,7 @@ func (hsv *HeaderSigVerifier) VerifySignature(header data.HeaderHandler) error {
 		epoch,
 	)
 	if err != nil {
+		logIfEpochConfigMissing(err, header, epoch, "signature")
 		return err
 	}
 
@@ -293,9 +295,42 @@ func (hsv *HeaderSigVerifier) getLeader(header data.HeaderHandler) (crypto.Publi
 
 	headerConsensusGroup, err := hsv.nodesCoordinator.ComputeConsensusGroup(prevRandSeed, header.GetSlot(), epoch)
 	if err != nil {
+		logIfEpochConfigMissing(err, header, epoch, "leader")
 		return nil, err
 	}
 
 	leaderPubKeyValidator := headerConsensusGroup[0]
 	return hsv.keyGen.PublicKeyFromByteArray(leaderPubKeyValidator.PubKey())
+}
+
+// logIfEpochConfigMissing emits a dedicated line when a header is rejected only
+// because the consensus configuration for its epoch has not been built yet. That
+// configuration is created when the epoch-start block is committed, so between an
+// epoch boundary and that commit every new-epoch header fails here, on gossip, on
+// self-requested headers and on the consensus topic alike (issue #90).
+//
+// Kept at debug level on purpose: an unauthenticated peer can trigger this path
+// with a header carrying an arbitrary future epoch, so a louder level would be a
+// cheap log-flood lever. Enable process/headerCheck:DEBUG to measure how many
+// slots the window actually spans. The header hash is deliberately absent for the
+// same reason: it is not computed yet at this point, and computing one here would
+// put a marshal plus a hash on a path an unauthenticated peer can drive.
+//
+// lookupEpoch is the epoch whose configuration was actually missing, which is not
+// the header's own epoch for an epoch-start header: both callers verify those
+// against the previous epoch. Both values are logged so the line cannot send an
+// operator looking for the wrong configuration.
+func logIfEpochConfigMissing(err error, header data.HeaderHandler, lookupEpoch uint32, stage string) {
+	if !errors.Is(err, sharding.ErrEpochNodesConfigDoesNotExist) {
+		return
+	}
+
+	log.Debug("header rejected, epoch consensus config not built yet",
+		"stage", stage,
+		"missing config for epoch", lookupEpoch,
+		"header epoch", header.GetEpoch(),
+		"nonce", header.GetNonce(),
+		"slot", header.GetSlot(),
+		"is epoch start", header.GetIsEpochStart(),
+	)
 }
--- a/core/process/headerCheck/headerSignatureVerify_test.go
+++ b/core/process/headerCheck/headerSignatureVerify_test.go
@@ -3,8 +3,10 @@ package headerCheck_test
 import (
 	"bytes"
 	"errors"
+	"fmt"
 	"testing"
 
+	logger "github.com/klever-io/klever-go-logger"
 	"github.com/klever-io/klever-go/common"
 	cMock "github.com/klever-io/klever-go/common/mock"
 	"github.com/klever-io/klever-go/core/process"
@@ -725,4 +727,140 @@ func TestHeaderSigVerifier_VerifySignatureOkWhenFallbackThresholdCouldBeApplied(
 	err := hdrSigVerifier.VerifySignature(header)
 	require.Nil(t, err)
 	require.True(t, wasCalled)
+}
+
+// epochConfigLogFormatter keeps everything except the epoch-config rejection line
+// out of the observer buffer, so unrelated log output from other tests in this
+// package cannot make the assertions below flaky. Same idiom as
+// core/process/transactionLog/printTxLogProcessor_test.go.
+type epochConfigLogFormatter struct {
+	logger.PlainFormatter
+}
+
+func (f *epochConfigLogFormatter) Output(line logger.LogLineHandler) []byte {
+	if line.GetMessage() != "header rejected, epoch consensus config not built yet" {
+		return nil
+	}
+
+	return f.PlainFormatter.Output(line)
+}
+
+// The epoch consensus configuration is only built when the epoch-start block is
+// committed, so between an epoch boundary and that commit every new-epoch header
+// is rejected here (issue #90).
+//
+// Two things must hold. The sentinel has to reach the caller intact, because the
+// worker's blacklist suppression keys off it (core/consensus/slot/worker.go) and
+// a header rejected only for this reason is early rather than invalid. And the
+// rejection has to be observable, because counting the slots it spans is how the
+// window's real duration gets measured.
+//
+// Not parallel: it mutates the global log level and observer list.
+func TestHeaderSigVerifier_EpochNodesConfigMissingIsSurfacedIntact(t *testing.T) {
+	const missingEpoch = uint32(7)
+
+	previousPattern := logger.GetLogLevelPattern()
+	require.Nil(t, logger.SetLogLevel("*:DEBUG"))
+
+	buff := &bytes.Buffer{}
+	require.Nil(t, logger.AddLogObserver(buff, &epochConfigLogFormatter{}))
+
+	t.Cleanup(func() {
+		require.Nil(t, logger.RemoveLogObserver(buff))
+		require.Nil(t, logger.SetLogLevel(previousPattern))
+	})
+
+	newArgsRejectingEpochConfig := func() *headerCheck.ArgsHeaderSigVerifier {
+		args := createHeaderSigVerifierArgs()
+		args.NodesCoordinator = &cMock.NodesCoordinatorMock{
+			ComputeValidatorsGroupCalled: func(_ []byte, _ uint64, epoch uint32) ([]sharding.Validator, error) {
+				// Same wrapping the real coordinator applies, so the test also
+				// pins that errors.Is survives it.
+				return nil, fmt.Errorf("%w epoch=%v", sharding.ErrEpochNodesConfigDoesNotExist, epoch)
+			},
+		}
+
+		return args
+	}
+
+	t.Run("leader lookup path", func(t *testing.T) {
+		hdrSigVerifier, err := headerCheck.NewHeaderSigVerifier(newArgsRejectingEpochConfig())
+		require.Nil(t, err)
+
+		header := &block.Block{Header: &block.BlockHeader{Epoch: missingEpoch, Nonce: 42, Slot: 43}}
+
+		err = hdrSigVerifier.VerifyRandSeed(header)
+
+		require.True(t, errors.Is(err, sharding.ErrEpochNodesConfigDoesNotExist), "got %v", err)
+		require.Contains(t, buff.String(), "leader")
+	})
+
+	t.Run("signature path", func(t *testing.T) {
+		hdrSigVerifier, err := headerCheck.NewHeaderSigVerifier(newArgsRejectingEpochConfig())
+		require.Nil(t, err)
+
+		header := &block.Block{
+			Header:        &block.BlockHeader{Epoch: missingEpoch, Nonce: 42, Slot: 43},
+			PubKeysBitmap: []byte("1"),
+		}
+
+		err = hdrSigVerifier.VerifySignature(header)
+
+		require.True(t, errors.Is(err, sharding.ErrEpochNodesConfigDoesNotExist), "got %v", err)
+		require.Contains(t, buff.String(), "signature")
+	})
+
+	// The emitted lines must carry the fields an operator needs to count how many
+	// slots the window spans, not just the message.
+	t.Run("emitted lines carry epoch, nonce and slot", func(t *testing.T) {
+		output := buff.String()
+
+		require.Contains(t, output, "missing config for epoch")
+		require.Contains(t, output, "header epoch")
+		require.Contains(t, output, "nonce")
+		require.Contains(t, output, "slot")
+		require.Contains(t, output, "42")
+		require.Contains(t, output, "43")
+	})
+
+	// For an epoch-start header both callers verify against the previous epoch,
+	// so the line must name that epoch as the missing one. Reporting the header's
+	// own epoch would send an operator looking for the wrong configuration.
+	t.Run("epoch-start header names the previous epoch as the missing one", func(t *testing.T) {
+		buff.Reset()
+
+		hdrSigVerifier, err := headerCheck.NewHeaderSigVerifier(newArgsRejectingEpochConfig())
+		require.Nil(t, err)
+
+		header := &block.Block{
+			Header: &block.BlockHeader{Epoch: missingEpoch, Nonce: 42, Slot: 43, IsEpochStart: true},
+		}
+
+		err = hdrSigVerifier.VerifyRandSeed(header)
+		require.True(t, errors.Is(err, sharding.ErrEpochNodesConfigDoesNotExist), "got %v", err)
+
+		require.Contains(t, buff.String(), "missing config for epoch = 6")
+		require.Contains(t, buff.String(), "header epoch = 7")
+	})
+
+	// A different error must leave the diagnostic silent, otherwise the line stops
+	// meaning "the epoch window is open".
+	t.Run("stays silent for an unrelated error", func(t *testing.T) {
+		buff.Reset()
+
+		args := createHeaderSigVerifierArgs()
+		args.NodesCoordinator = &cMock.NodesCoordinatorMock{
+			ComputeValidatorsGroupCalled: func(_ []byte, _ uint64, _ uint32) ([]sharding.Validator, error) {
+				return nil, errors.New("some other failure")
+			},
+		}
+
+		hdrSigVerifier, err := headerCheck.NewHeaderSigVerifier(args)
+		require.Nil(t, err)
+
+		err = hdrSigVerifier.VerifyRandSeed(&block.Block{Header: &block.BlockHeader{}})
+		require.NotNil(t, err)
+
+		require.Empty(t, buff.String())
+	})
 }
--- a/core/process/headerCheck/headerSignatureVerify_test.go
+++ b/core/process/headerCheck/headerSignatureVerify_test.go
@@ -810,17 +810,27 @@ func TestHeaderSigVerifier_EpochNodesConfigMissingIsSurfacedIntact(t *testing.T)
 		require.Contains(t, buff.String(), "signature")
 	})
 
-	// The emitted lines must carry the fields an operator needs to count how many
-	// slots the window spans, not just the message.
-	t.Run("emitted lines carry epoch, nonce and slot", func(t *testing.T) {
-		output := buff.String()
+	// A single emitted line must carry the fields an operator needs to count how
+	// many slots the window spans. Asserting on one line rather than on the
+	// accumulated buffer is what makes this meaningful: over the concatenation of
+	// several lines the field assertions would still pass with a field missing
+	// from one of them.
+	t.Run("a single emitted line carries epoch, nonce and slot", func(t *testing.T) {
+		buff.Reset()
 
-		require.Contains(t, output, "missing config for epoch")
-		require.Contains(t, output, "header epoch")
-		require.Contains(t, output, "nonce")
-		require.Contains(t, output, "slot")
-		require.Contains(t, output, "42")
-		require.Contains(t, output, "43")
+		hdrSigVerifier, err := headerCheck.NewHeaderSigVerifier(newArgsRejectingEpochConfig())
+		require.Nil(t, err)
+
+		header := &block.Block{Header: &block.BlockHeader{Epoch: missingEpoch, Nonce: 42, Slot: 43}}
+
+		err = hdrSigVerifier.VerifyRandSeed(header)
+		require.True(t, errors.Is(err, sharding.ErrEpochNodesConfigDoesNotExist), "got %v", err)
+
+		output := buff.String()
+		require.Contains(t, output, "missing config for epoch = 7")
+		require.Contains(t, output, "header epoch = 7")
+		require.Contains(t, output, "nonce = 42")
+		require.Contains(t, output, "slot = 43")
 	})
 
 	// For an epoch-start header both callers verify against the previous epoch,
--- a/core/process/sync/baseForkDetector.go
+++ b/core/process/sync/baseForkDetector.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"math"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/klever-io/klever-go/core/consensus"
@@ -52,6 +53,10 @@ type baseForkDetector struct {
 	genesisSlot        uint64
 	maxForkHeaderEpoch uint32
 	tmpStuck           time.Time
+	// lastCheckpointAheadWarnSlot throttles the checkpoint-ahead warning to once
+	// per slot; CheckFork can run on every sync-loop iteration while the node is
+	// not synchronized.
+	lastCheckpointAheadWarnSlot atomic.Int64
 }
 
 // SetRollBackNonce sets the nonce where the chain should roll back
@@ -606,17 +611,18 @@ func (bfd *baseForkDetector) isConsensusStuck() bool {
 
 	currentSlot := tools.SafeI64ToU64(bfd.slotManager.Index())
 	lastCheckpointSlot := bfd.lastCheckpoint().slot
-	if currentSlot < lastCheckpointSlot {
+	slotsDifference, err := tools.SafeSubUint64(currentSlot, lastCheckpointSlot)
+	if err != nil {
 		// The last checkpoint is ahead of our own slot index, so no slots have
 		// elapsed since it. Subtracting would wrap around on uint64 and report an
 		// enormous lag, which clears the threshold below and forces a rollback of
 		// a block that was just committed. checkBlockBasicValidity deliberately
 		// accepts headers one slot ahead of the local index, so a node whose clock
 		// trails its peers can reach this state without any peer misbehaving.
+		bfd.warnOnceCheckpointAheadOfSlotIndex(currentSlot, lastCheckpointSlot)
 		return false
 	}
 
-	slotsDifference := currentSlot - lastCheckpointSlot
 	if slotsDifference <= process.MaxSlotsWithoutCommittedBlock {
 		return false
 	}
@@ -626,6 +632,31 @@ func (bfd *baseForkDetector) isConsensusStuck() bool {
 	}
 
 	return true
+}
+
+// warnOnceCheckpointAheadOfSlotIndex surfaces the clock-trails-tip state. While
+// it lasts, the node is otherwise silent: incoming headers are dropped below the
+// default log level because their slot exceeds the local index, the
+// bootstrapper's slot lag reads zero so its stall warning cannot fire, and the
+// node keeps reporting NsSynchronized. After an NTP step-back or a VM resume
+// this can hold for a long time, so this line is the only operator-visible
+// signal. Throttled to once per slot because CheckFork runs on every sync-loop
+// iteration while the node is not synchronized.
+//
+// The format does happen under mutNodeState, since CheckFork's only production
+// caller holds it. That is accepted here, unlike for the stall warning that was
+// moved out of that lock: the throttle caps this at one format per slot, and
+// the state has no out-of-lock observer to move it to.
+func (bfd *baseForkDetector) warnOnceCheckpointAheadOfSlotIndex(currentSlot uint64, checkpointSlot uint64) {
+	slotIndex := bfd.slotManager.Index()
+	if bfd.lastCheckpointAheadWarnSlot.Swap(slotIndex) == slotIndex {
+		return
+	}
+
+	log.Warn("last checkpoint is ahead of the local slot index, node clock appears to trail the network",
+		"local slot index", currentSlot,
+		"checkpoint slot", checkpointSlot,
+		"checkpoint nonce", bfd.lastCheckpoint().nonce)
 }
 
 func (bfd *baseForkDetector) isSyncing() bool {
--- a/core/process/sync/baseForkDetector.go
+++ b/core/process/sync/baseForkDetector.go
@@ -604,7 +604,19 @@ func (bfd *baseForkDetector) isConsensusStuck() bool {
 		return false
 	}
 
-	slotsDifference := tools.SafeI64ToU64(bfd.slotManager.Index()) - bfd.lastCheckpoint().slot
+	currentSlot := tools.SafeI64ToU64(bfd.slotManager.Index())
+	lastCheckpointSlot := bfd.lastCheckpoint().slot
+	if currentSlot < lastCheckpointSlot {
+		// The last checkpoint is ahead of our own slot index, so no slots have
+		// elapsed since it. Subtracting would wrap around on uint64 and report an
+		// enormous lag, which clears the threshold below and forces a rollback of
+		// a block that was just committed. checkBlockBasicValidity deliberately
+		// accepts headers one slot ahead of the local index, so a node whose clock
+		// trails its peers can reach this state without any peer misbehaving.
+		return false
+	}
+
+	slotsDifference := currentSlot - lastCheckpointSlot
 	if slotsDifference <= process.MaxSlotsWithoutCommittedBlock {
 		return false
 	}
--- a/core/process/sync/baseSync.go
+++ b/core/process/sync/baseSync.go
@@ -3,6 +3,7 @@ package sync
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"math"
 	"sync"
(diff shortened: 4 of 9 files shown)


core/process/sync/baseSync.go
```diff
@@ -358,17 +359,78 @@ func (boot *baseBootstrap) shouldTryToRequestHeaders() bool {
...
+	// that forced-rollback loop: once a requested header lands, isSyncing() turns
+	// true and the next stuck check stands down.
+	//
+	// Import mode is excluded because the once-per-slot bound below does not hold
+	// there: GetNodeState always answers NsNotSynchronized while importing, so
+	// syncBlock never takes its early return, the defer clearing
+	// isNodeStateCalculated runs on every pass, and computeNodeState re-enters
+	// every sleepTime. Replaying historical blocks also produces an unbounded
+	// slot lag by construction, and there are no peers to request from.
+	if boot.isInImportMode {
```
MathijsBok, line 385: 🟡 **Changed after review** (S1)

My original justification for putting this branch at `Warn` and for calling it bounded was that it fires at most once per slot. That does not hold in import-db mode, and I missed it in the first commit.

`GetNodeState()` returns `NsNotSynchronized` unconditionally while `isInImportMode` (`baseSync.go:1004`), so `syncBlock` never takes its early return, the `defer` clearing `isNodeStateCalculated` runs on every pass, and `computeNodeState` re-enters every `sleepTime` (5 ms). Meanwhile `isNodeSynchronized` itself can be computed true, since a replay sits at `ProbableHighestNonce() <= currentBlockNonce`. Net effect on a long import: this warning and a `go requestHeadersIfSyncIsStuck()` roughly 200 times per second.

Fixed in `8d4806c` by returning early in import mode. Only the new branch is gated, so pre-existing behaviour in import mode is untouched. Beyond the log volume it is also simply the right answer: replaying historical blocks produces an unbounded slot lag by construction, and there are no peers to request from.

Covered by `TestShouldTryToRequestHeaders_ImportModeIgnoresSlotLag`, which also asserts the counter-case so it cannot pass for the wrong reason.

core/process/sync/baseSync.go
```diff
@@ -576,6 +638,12 @@ func (boot *baseBootstrap) syncBlock() error {
 	boot.computeNodeState()
 	nodeState := boot.GetNodeState()
 	if nodeState != core.NsNotSynchronized {
+		// Returning here leaves isNodeStateCalculated set, so computeNodeState
```
MathijsBok, line 660: 💡 **Rationale** (R2)

This comment exists because the once-per-slot bound that the warning and the stuck-request goroutine rely on lives here, in a different function from the one that depends on it.

The dependency is fragile in a specific way: hoisting the `defer` above this `return nil` looks like harmless cleanup, and would put both onto the 5 ms loop with nothing in `shouldTryToRequestHeaders` pointing back here. A comment on the depended-upon side is what actually survives that refactor.

Added in `8d4806c`.

core/process/headerCheck/headerSignatureVerify_test.go
```diff
@@ -726,3 +728,149 @@ func TestHeaderSigVerifier_VerifySignatureOkWhenFallbackThresholdCouldBeApplied(
...
+		require.True(t, errors.Is(err, sharding.ErrEpochNodesConfigDoesNotExist), "got %v", err)
+		require.Contains(t, buff.String(), "signature")
+	})
+
+	// A single emitted line must carry the fields an operator needs to count how
+	// many slots the window spans. Asserting on one line rather than on the
+	// accumulated buffer is what makes this meaningful: over the concatenation of
+	// several lines the field assertions would still pass with a field missing
+	// from one of them.
+	t.Run("a single emitted line carries epoch, nonce and slot", func(t *testing.T) {
```
MathijsBok, line 818: 💡 **Changed after review** (R1, M1)

This subtest used to assert on the accumulated buffer left behind by the two subtests above it, with no `Reset()` of its own. Two consequences, both confirmed by running it:

```
go test ./core/process/headerCheck/ -count=1 -run '.../emitted_lines_carry_epoch,_nonce_and_slot'
--- FAIL: ...
    "" does not contain "missing config for epoch"
```

It was not independently runnable, so narrowing down a failure with `-run` produced an empty-buffer error that looks like a product bug. And because it read the concatenation of two lines, `Contains(output, "42")` and `"43"` did not establish that a *single* line carried both fields; it would still have passed with one field missing from one stage.

Fixed in `8d4806c`: it now emits its own rejection and asserts on that one line, which also let the assertions tighten from bare `"42"` to `"nonce = 42"` and `"slot = 43"`.

core/process/sync/baseSync_test.go
```diff
@@ -0,0 +1,337 @@
...
+		assert.Empty(t, stub.requestedNonces)
+	})
+}
+
+// doJobOnSyncBlockFail decides whether a failed sync attempt rolls the chain
+// back. The decision hinges on comparing the failure against
+// process.ErrTimeIsOut: a timeout is expected and must not roll back, anything
+// else must. Comparing with != instead of errors.Is silently rolls back on a
+// wrapped timeout, so both forms are pinned here.
+func TestDoJobOnSyncBlockFail(t *testing.T) {
```
MathijsBok, line 279: 💡 **Added after review** (M2)

The `errors.Is` change at `baseSync.go:596` was the only uncovered new line in the first commit, and it sits in the highest-complexity function this branch touches (cognitive 10, cyclomatic 8, previously 0% coverage).

That line is not cosmetic: it feeds `isProcessWithError`, which feeds `shouldRollBack`, which is the difference between reverting a committed block and not. A wrapped `ErrTimeIsOut` under the old `!=` comparison would have rolled back where a plain one did not.

This table pins all four cases. Verified by mutation: restoring `err != process.ErrTimeIsOut` fails exactly the wrapped-timeout subtest and nothing else. Coverage of `doJobOnSyncBlockFail` goes from 0% to 88.2%.

The cases deliberately observe the decision through `forkDetector.RemoveHeader`, which is reached inside the rollback branch but before `rollBack` itself, so the test needs no rollback fixture.

core/process/sync/baseForkDetector.go
```diff
@@ -604,7 +604,19 @@ func (bfd *baseForkDetector) isConsensusStuck() bool {
 		return false
 	}
 
-	slotsDifference := tools.SafeI64ToU64(bfd.slotManager.Index()) - bfd.lastCheckpoint().slot
+	currentSlot := tools.SafeI64ToU64(bfd.slotManager.Index())
+	lastCheckpointSlot := bfd.lastCheckpoint().slot
+	if currentSlot < lastCheckpointSlot {
```
MathijsBok: 🟢 **Deliberate, please do not simplify away** (S2, R3, M4)

This guard is the highest-impact part of the branch and it is easy to mistake for defensive noise. It is not: without it, `SafeI64ToU64(Index()) - lastCheckpoint().slot` wraps whenever the checkpoint slot exceeds the local slot index. The wrapped value clears `MaxSlotsWithoutCommittedBlock`, so `isConsensusStuck()` returns true, `CheckFork()` returns the `NewForkInfo()` default shape (`IsDetected=true, Nonce=MaxUint64, Hash=nil`), which is exactly what `isForcedRollBackOneBlock()` matches, driving `rollBackOneBlockForced` into `rollBack(false)`. That reverts the chain head and the account state of a block the node had just committed.

Reachable without any peer misbehaving: `checkBlockBasicValidity` explicitly permits `header.GetSlot() == Index()+1` (`:125-127`), so a node whose clock trails its peers by under one slot can commit a header one slot ahead, and `doJobOnBHProcessed` writes that slot into the checkpoint. CWE-191 into CWE-754.

Returning `false` rather than clamping to 0 is intentional: zero elapsed slots and "not stuck" are the same answer here, and because `checkBlockBasicValidity` caps an accepted header at `Index()+1`, the condition can only hold for at most one slot, so the guard cannot mask a genuine stall.

core/process/sync/metaForkDetector_test.go
```diff
@@ -153,3 +154,69 @@ func TestMetaForkDetector_AddHeaderHigherNonceThanSlotShouldErr(t *testing.T) {
...
+
+	forkInfo := bfd.CheckFork()
+
+	assert.False(t, forkInfo.IsDetected)
+}
+
+// The counterpart of the guard above: with the checkpoint genuinely behind the
+// slot index, the consensus-stuck detection must still fire. Without this the
+// guard could disable the mechanism entirely and no test would notice.
+func TestMetaForkDetector_CheckForkStillDetectsStuckConsensusWhenCheckpointIsBehind(t *testing.T) {
```
MathijsBok, line 198: 🟢 **Deliberate pairing** (S4)

This test looks redundant next to the one above it and is not. A guard of the form `if a < b { return false }` placed in front of a safety mechanism can disable that mechanism outright, and a suite that only asserts "no forced rollback when the checkpoint is ahead" would pass just as happily against `func isConsensusStuck() bool { return false }`.

Asserting that a genuinely-behind checkpoint still yields `IsDetected` with the exact `Nonce == MaxUint64, Hash == nil` shape that `isForcedRollBackOneBlock` matches is what pins the recovery path the guard sits in front of. If either test is ever dropped, drop this one last.

core/process/headerCheck/headerSignatureVerify.go
```diff
@@ -293,9 +295,42 @@ func (hsv *HeaderSigVerifier) getLeader(header data.HeaderHandler) (crypto.Publi
...
+// lookupEpoch is the epoch whose configuration was actually missing, which is not
+// the header's own epoch for an epoch-start header: both callers verify those
+// against the previous epoch. Both values are logged so the line cannot send an
+// operator looking for the wrong configuration.
+func logIfEpochConfigMissing(err error, header data.HeaderHandler, lookupEpoch uint32, stage string) {
+	if !errors.Is(err, sharding.ErrEpochNodesConfigDoesNotExist) {
+		return
+	}
+
+	log.Debug("header rejected, epoch consensus config not built yet",
```
MathijsBok, line 328: 🟢 **Three deliberate choices** (S3)

Worth stating explicitly, because each looks like an omission:

**Debug, not warn.** `SingleDataInterceptor.ProcessReceivedMessage` runs `CheckValidity()` before the whitelist and originator check (`singleDataInterceptor.go:135-141`), so any connected peer can reach this with an arbitrary epoch and get the sentinel back from the `nodesConfig[epoch]` map miss. At the default level nothing is emitted, so it is not a flood lever.

**No header hash.** It is not computed at this point, and computing one here would put a marshal plus a hash on a path an unauthenticated peer can drive.

**The error is returned unchanged.** `Worker.shouldBlacklistPeer` suppresses blacklisting on exactly this sentinel (`worker.go:449`). Substituting or flattening it would blacklist every honest peer gossiping a new-epoch header during the boundary window, which is an availability and partition risk. The test pinning `errors.Is` survival through the coordinator's `%w` wrapping is there to keep that from breaking silently.

The logged fields are all values the peer supplied, so there is nothing local disclosed. Both the looked-up epoch and the header's own epoch are logged because they differ for an epoch-start header, which verifies against the previous epoch.

core/process/sync/baseSync_test.go
```diff
@@ -0,0 +1,337 @@
...
+// A node whose header intake is blocked keeps a frozen fork detector, so it
+// reports itself synchronized while the chain moves on (issue #90). It must then
+// react to the slot lag on every slot, not only on the every-20th-slot cadence
+// this replaced.
+func TestShouldTryToRequestHeaders_SyncedNodeReactsToSlotLagOnEverySlot(t *testing.T) {
+	t.Parallel()
+
+	const lastCommittedSlot = uint64(0)
+
+	for slotIndex := int64(1); slotIndex <= 40; slotIndex++ {
```
MathijsBok, line 97: 🟢 **Why a sweep and not a sample** (R4, M3)

Sweeping indices 1 to 40 rather than picking a couple is what makes this a regression test rather than a restatement of the implementation: against the previous every-20th-slot trigger it fails on 28 of the 40 indices (11 to 19 and 21 to 39), instead of merely passing against the new one.

Same reasoning elsewhere in this file: `TestRequestHeadersIfSyncIsStuck` asserts the exact nonce list `8..17` rather than "something was requested", which pins the `min(MaxHeadersToRequestInAdvance, lag-1)` formula and the start nonce that operators actually see in the burst logs.

MathijsBok (commented):
## Self-review notes

Recording what came out of my own review pass over this branch, so the reasoning behind a few non-obvious choices is on the record rather than only in my head. Everything actionable is already addressed in `8d4806c`; the rest is rationale for reviewers.

8 notes: 1 that changed behaviour, 3 that changed tests or comments, 4 recording why something is the way it is.

Legend: 🟡 changed behaviour | 💡 rationale or test change | 🟢 deliberate choice worth not undoing

fbsobreira (changes requested):
The core mechanism checks out and I want this fix in — verified the once-per-slot bound, the `errors.Is` behavior under the coordinator's `%w` wrapping, lock ordering, `-race`, and a clean trial merge against develop. I also traced the `isConsensusStuck` guard through history: it **restores** the pre-`31c39abf4` signed-arithmetic semantics that the 2024 G115 overflow sweep accidentally inverted (the original `int64` difference went negative when the checkpoint was ahead, correctly returning `false`; the unsigned conversion made it underflow and force a rollback of a just-committed block). So the guard is a regression fix, not new behavior — good catch.

Requesting two small changes before merge, plus one suggestion:

### Required

1. **Log inside both underflow guards.** In `isConsensusStuck` (`baseForkDetector.go` `currentSlot < lastCheckpointSlot` branch) and `slotsSinceLastCommittedBlock` (`baseSync.go` same condition), the guard returns silently. In the clock-trails-tip case (NTP step-back, VM resume), header intake is simultaneously blocked (`checkBlockBasicValidity` rejects at Debug, `processReceivedHeader` skips `AddHeader` at Trace), the lag reads 0 so the new Warn at `baseSync.go:402` can never fire, and the node sits frozen reporting `NsSynchronized` with zero diagnostic output — potentially for hours. Pre-PR the wrap at least produced visible rollback churn. A `log.Warn` in each branch restores observability for the exact state this PR makes quiet.

2. **Move the new `log.Warn` out of the `mutNodeState` critical section.** `shouldTryToRequestHeaders` (and its Warn at `baseSync.go:402`) runs inside `computeNodeState`'s `mutNodeState.Lock()`, which consensus contends on via `GetNodeState` (`worker.go:338`, `subslotStartSlot.go:103`) — once per stall slot, consensus callers block behind log formatting inside a write lock. Moving the Warn into `requestHeadersIfSyncIsStuck` (already a goroutine outside the lock, already recomputes the lag) preserves semantics and the once-per-slot bound, keeps the predicate side-effect-free, and drops the test-only need to stub the fork detector for the log call.

### Suggested (one field solves two issues)

A `lastStuckRequestSlot` field checked in `requestHeadersIfSyncIsStuck` would:
- make the once-per-slot throttle **self-enforcing** — today it holds only because `syncBlock`'s early return sits above the defer that clears `isNodeStateCalculated`, an invariant defended by comments in three places and pinned by no test; a future refactor hoisting the defer silently turns this into a request burst every 5ms loop iteration with all tests green;
- allow cheap **backoff**: re-firing only every N slots (e.g. the old 20) after the first burst caps request amplification during a network-wide halt — currently every synced node emits up to 20 header requests every slot for the whole outage, and the requested-items TimeCache spans exactly one slot interval so it never dedups across slots.

### Noted, out of scope for this PR (tickets)

- The false `NsSynchronized` report itself is unchanged (the PR states this) — that's the klc-1920/klc-2389 branch's territory; the two are complementary and should be reconciled after this lands.
- Rare stall variant: a header cached in the pool during a slot-skew window is never registered with the fork detector, and `requestHeaders` skips pooled nonces — the burst can't recover that case.
- The whole recovery relies on epoch-start headers verifying against epoch-1 (`headerSignatureVerify.go:105`, `:290`, both TODO-marked). If that TODO is ever executed, issue #90 regresses with sync tests still green — worth an integration test pinning "a requested epoch-start header is verifiable while the new epoch's config is missing".

### Optional cleanup

The saturating slot subtraction is hand-rolled in both new sites while `tools.SafeSubUint64` exists (`tools/computers.go:160`); `slotsSinceLastCommittedBlock` re-derives the last-committed-slot that `computeNodeState` already computes inline. Fine to fold in if you're touching the files anyway.

Also a reminder: the ruleset requires all review threads resolved — the 8 open self-annotation threads will block merge until closed.

MathijsBok:
Both required changes are in as of a081057, plus the throttle half of the suggestion. Point by point, including where I deviated and why.

Also: thanks for the `31c39abf4` trace. Verified it on my side: the G115 sweep replaced `bfd.slotManager.Index() - int64(bfd.lastCheckpoint().slot)` with the unsigned form, so the guard indeed restores the pre-sweep semantics rather than adding new behavior.

### Required 1: log inside both underflow guards

Done, with one deviation on level, stated openly so it is a decision and not an omission.

- **Fork detector guard** (`baseForkDetector.go`, `isConsensusStuck`): `log.Warn`, throttled to once per slot through a new atomic `lastCheckpointAheadWarnSlot` (`warnOnceCheckpointAheadOfSlotIndex`). The throttle is not optional: `CheckFork` runs on every 5 ms sync-loop iteration while the node is not synchronized, and a not-synchronized node with a stepped-back clock would otherwise emit ~200 lines per second. The message names the local slot index, the checkpoint slot and the checkpoint nonce.
- **baseSync helper** (`slotsSinceLastCommittedBlock`): `log.Debug` rather than warn, for two reasons. It can run under `mutNodeState`, the same lock your point 2 moves log formatting out of, and in the frozen state the fork-detector warning already fires in the same slot via `CheckFork`, so a second warn would double-log every occurrence of one condition. If you want warn in both regardless, say so and I flip it.

Your scenario is pinned by `TestMetaForkDetector_CheckpointAheadWarnsOncePerSlot`: checkpoint at slot 100, index stepped back to 50, first `CheckFork` warns, second in the same slot stays silent, next slot warns again.

### Required 2: move the warn out of the mutNodeState critical section

Done, via a `stalledWhileSynced bool` captured at the spawn site in `computeNodeState` and passed to `requestHeadersIfSyncIsStuck`. The predicate is side-effect-free again.

One nuance on top of your suggestion: warning unconditionally inside `requestHeadersIfSyncIsStuck` would spam during ordinary catch-up, because on the not-synchronized path that goroutine is spawned on every `computeNodeState` pass, every 5 ms, and the lag is legitimately above the threshold the whole time. The flag keeps the warning exclusive to the state it describes. Pinned by `TestRequestHeadersIfSyncIsStuck_WarnsOnlyWhenStalledWhileSynced`: warns with the flag, stays silent without it while the burst still fires, and shares the once-per-slot cap.

The independent pass over this increment caught that the spawn-site wiring itself was the one unpinned link (hard-wiring the argument to false kept every test green). `TestComputeNodeState_StalledWhileSyncedWiresWarnIntoBurst` closes that: it drives `computeNodeState` end to end with a frozen fork detector and asserts the warning through the log observer, and it fails on the hard-wired mutant.

### Suggested: `lastStuckRequestSlot`

The self-enforcing half is in, exactly as you described: atomic swap on the slot index inside `requestHeadersIfSyncIsStuck`, so the once-per-slot bound no longer depends on `computeNodeState`'s memoization or on the defer position in `syncBlock` (their comments now say the cap holds regardless), and concurrent callers in the same slot collapse to one burst. Pinned by `TestRequestHeadersIfSyncIsStuck_FiresAtMostOncePerSlot`.

The multi-slot backoff I deliberately did not add yet. Re-firing only every N slots re-slows exactly the recovery this PR exists for: if the first burst lands while the epoch window is still open, per-slot retry heals within one slot of the window closing, and an N=20 backoff can add up to 80 s on top. The field makes backoff a one-line change, and there is now a node on chain 109 (10-minute epochs) set up to measure the real request cadence across boundaries. If the data shows sustained multi-slot bursts, backoff goes in on top of this field; merge stays on hold for that run either way, per the Slack thread.

### Out-of-scope notes

Filed so they carry numbers: #135 for the pooled-header-during-slot-skew stall variant, #136 for the integration test pinning that an epoch-start header stays verifiable against epoch N-1 while `nodesConfig[N]` is missing (the TODO dependency you flagged). The klc-1920/klc-2389 reconciliation I left to that branch's owner.

### Optional cleanup

`tools.SafeSubUint64` folded into both sites, hand-rolled comparisons gone. The other one (deduplicating the last-committed-slot derivation with `computeNodeState`) I left: the helper serves callers on both sides of the lock, and threading the value through would couple the goroutine to the locked section's locals. Happy to do it if you feel strongly.

### Threads

The 8 self-annotation threads are resolved.

### Verification

`gofmt` clean, `go build ./...` clean, `go vet` clean, `golangci-lint` 0 issues on both packages, `go test -race` green on both, `go test ./core/... ./sharding/...` 75 ok, 0 failures. Coverage on the touched functions: `warnOnceCheckpointAheadOfSlotIndex` 100%, `shouldTryToRequestHeaders` 100%, `slotsSinceLastCommittedBlock` 100%, `requestHeadersIfSyncIsStuck` 93.3%. An independent adversarial review pass ran over the increment before pushing.

core/process/sync/baseForkDetector.go
```diff
@@ -616,6 +634,31 @@ func (bfd *baseForkDetector) isConsensusStuck() bool {
...
+// iteration while the node is not synchronized.
+//
+// The format does happen under mutNodeState, since CheckFork's only production
+// caller holds it. That is accepted here, unlike for the stall warning that was
+// moved out of that lock: the throttle caps this at one format per slot, and
+// the state has no out-of-lock observer to move it to.
+func (bfd *baseForkDetector) warnOnceCheckpointAheadOfSlotIndex(currentSlot uint64, checkpointSlot uint64) {
+	slotIndex := bfd.slotManager.Index()
+	if bfd.lastCheckpointAheadWarnSlot.Swap(slotIndex) == slotIndex {
+		return
```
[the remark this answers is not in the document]
MathijsBok: Fixed in 90db4bd, and the regression test you asked for surfaced a second instance of the same collision one field over: lastSlotWithForcedFork also zero-initializes, and its first check in isConsensusStuck made slot index 0 unreachable for the entire function, so the sentinel alone was not enough. Both fields now start at math.MinInt64, as does lastStuckRequestSlot in the bootstrapper for the same pattern. Pinned by TestMetaForkDetector_CheckpointAheadWarnsAtSlotIndexZero, which fails without either sentinel.

MathijsBok:
## Live test results (KleverChain testnet, chain 109)

Per the Slack thread and the review hold: this branch has now run on a live network. Setup: observer node on testnet (10-minute epochs, 150 slots x 4 s), image `v1.7.21-rc1-130` built from this branch at `90db4bd`, deployed via the standard container path, log level `*:INFO,process/sync:DEBUG,process/headerCheck:DEBUG` during the measurement so every line this PR adds or relies on was capturable. Baseline first: the same node on `v1.7.21-rc1-110` tracked one epoch boundary per-slot with zero stalls.

### Six epoch boundaries, healthy operation

900 API samples at 4 s intervals, 60 minutes, boundaries 79745 through 79751:

- Nonce advanced every sample, no stall of 8 s or longer anywhere, including across all six boundaries.
- `klv_is_syncing` stayed 0 throughout.
- Log signal counts over the whole hour, with debug enabled: `header rejected, epoch consensus config not built yet` **0**, stall warning **0**, checkpoint-ahead warning **0**, `requestHeadersIfSyncIsStuck` **0**, rollbacks **0**.

So on a healthy node: no regression, no request-cadence change, no log noise, and the issue #90 epoch-config window did not even open on any of the six boundaries. The epoch-start block commits and the new config is built before any new-epoch header reaches validation.

### Induced 16-slot stall (container pause)

To exercise the recovery path live rather than only in unit tests: `docker pause` for 65 s (about 16 slots, well past the 10-slot threshold), then unpause.

- The slot-lag trigger fired on the first evaluation after unpause, 6 seconds in: `requestHeadersIfSyncIsStuck from nonce = 11946957 to nonce = 11946973`, which is exactly `min(20, lag-1)` = 17 headers for the lag at that moment.
- **Exactly one burst.** The once-per-slot cap held; gossip took over immediately and no second burst was needed.
- Full catch-up within 45 seconds of unpause, `klv_is_syncing` back to 0, and no rollback lines in the log (captured at debug level), so the forced-rollback loop this PR describes never engaged.
- The stall warning correctly stayed silent: after unpause gossip flowed, `probableHighestNonce` rose, so the node knew it was behind and took the honest catch-up path. The warning is reserved for the frozen-intake state, which this test deliberately does not produce.

### Backoff decision, per the review thread

Data says no backoff, and the field stays as the hook: zero bursts across six healthy boundaries, and one single burst sufficing for a real 16-slot stall. There is no observed state in which re-firing every N slots would have reduced traffic, and the latency cost of backoff in the epoch-window case remains as described. If a future measurement shows sustained multi-slot bursting, it is a one-line addition on `lastStuckRequestSlot`.

### Honest limitation

The frozen-intake false-synchronized state itself (probable frozen while the chain moves, the exact #90 incident) is not reproducible on a healthy testnet node without an instrumented build, so that path is covered by the unit and wiring tests plus the mutation checks, not by this live run. The two adjacent gaps found during review are tracked in #135 and #136.

The node has been reverted to the canonical run line (normal log view) after the measurement.

core/process/sync/baseSync.go
```diff
@@ -358,21 +363,97 @@ func (boot *baseBootstrap) shouldTryToRequestHeaders() bool {
...
+	// threshold and fires on process.SlotModulusTrigger (5), so every slot that
+	// satisfied the old modulus of 20 also produced a forced-rollback ForkInfo,
+	// which the isForcedRollBackOneBlock guard above short-circuits on.
+	// Requesting on the slots in between is what keeps the fork detector out of
+	// that forced-rollback loop: once a requested header lands, isSyncing() turns
+	// true and the next stuck check stands down.
+	//
+	// Import mode replays historical blocks, so the slot lag is unbounded by
+	// construction and there are no peers to request from.
+	if boot.isInImportMode {
```
fbsobreira, line 385: This guard is only reached when `isNodeSynchronized` is true, because the `if !boot.isNodeSynchronized { return true }` above returns first. Import-db nodes are normally not synced, so they still spawn `requestHeadersIfSyncIsStuck` and send header bursts every slot. That contradicts the comment ("there are no peers to request from"). Should the import-mode check move above the not-synced early return? The new test only covers `isNodeSynchronized=true`, so please add the not-synced case too.

core/process/sync/baseSync.go
```diff
@@ -358,21 +363,97 @@ func (boot *baseBootstrap) shouldTryToRequestHeaders() bool {
...
+		return
+	}
+
+	if stalledWhileSynced {
+		// Warn level, and emitted here rather than in shouldTryToRequestHeaders:
+		// this goroutine runs outside mutNodeState, which consensus contends on
+		// through GetNodeState, so log formatting stays out of that lock. It
+		// cannot be driven by a peer, since both operands are local (own slot
+		// index, own last committed block).
+		log.Warn("node believes it is synchronized but has not committed a block for a while",
```
fbsobreira, line 451: During a genuine network-wide halt, or any run of more than `MaxSlotsWithoutNewBlockReceived` empty slots, every node hits this path every slot. Each one logs a Warn that describes a local false-sync fault, and each one sends header-by-nonce requests that no peer can answer. The load grows with node count × slots for the whole halt. Could we tell the two cases apart, for example by only warning and requesting when peers advertise a higher nonce (`ProbableHighestNonce` > current), or at least log the Warn only once when the node enters the state?

core/process/sync/baseSync.go
```diff
@@ -358,21 +363,97 @@ func (boot *baseBootstrap) shouldTryToRequestHeaders() bool {
...
 		return
 	}
 
+	// Self-enforcing once-per-slot cap. Before this field the bound depended on
+	// computeNodeState's per-slot memoization, which in turn depended on the
+	// position of a defer in syncBlock; a refactor there would have silently put
+	// this burst on the 5 ms sync loop. The swap also keeps concurrent callers
+	// in the same slot down to one burst.
+	currentSlot := boot.slotManager.Index()
+	if boot.lastStuckRequestSlot.Swap(currentSlot) == currentSlot {
```
fbsobreira, line 441: `lastStuckRequestSlot` now also rate-limits the not-synchronized recovery path, not just the new synced path. It's claimed before the burst is sent. If the first burst in a slot is lost, the retry waits a full slot, where before it went out on the next `syncBlock` iteration. Also, if a not-synced call claims the slot first, a synced call later in the same slot loses its Warn. Can the cap be limited to the synced path?

core/process/sync/baseSync.go
```diff
@@ -358,21 +363,97 @@ func (boot *baseBootstrap) shouldTryToRequestHeaders() bool {
...
+	// Import mode replays historical blocks, so the slot lag is unbounded by
+	// construction and there are no peers to request from.
+	if boot.isInImportMode {
+		return false
+	}
+
+	// The resulting burst and its warning are capped to once per slot by
+	// lastStuckRequestSlot inside requestHeadersIfSyncIsStuck, independently of
+	// how often this predicate is evaluated.
+	return boot.slotsSinceLastCommittedBlock() > process.MaxSlotsWithoutNewBlockReceived
```
fbsobreira, line 392: `RequestHeaderByNonce` dedups through `requestedItemsHandler`, and that handler's TTL equals the slot duration (`cmd/node/startup.go`). Bursts one slot apart for the same nonces can be dropped silently if the cache hasn't swept yet, so the effective cadence may be every other slot. That weakens the "requesting on the slots in between" argument in the comment above. Please double-check this interaction.

core/process/sync/baseSync.go
```diff
@@ -358,21 +363,97 @@ func (boot *baseBootstrap) shouldTryToRequestHeaders() bool {
 		return true
 	}
 
-	return boot.slotManager.Index()%process.SlotModulusTriggerWhenSyncIsStuck == 0
+	// The node believes it is synchronized, but that belief comes from the fork
```
fbsobreira, line 366: Design note: this works around the symptom. `NsSynchronized` and `MetricIsSyncing` still come only from the fork detector's counters, so a stalled node keeps reporting itself as synced to monitoring and the API. Did you consider either of these?
- Include the local slot lag in the reported sync state (metric and API only, consensus gating unchanged).
- Queue headers rejected with `ErrEpochNodesConfigDoesNotExist` and retry them once the epoch-start block commits.

Not necessarily blocking, but I'd like to hear the reasoning.

core/process/sync/baseSync.go
```diff
@@ -340,7 +345,7 @@ func (boot *baseBootstrap) computeNodeState() {
 		"isNodeSynchronized", boot.isNodeSynchronized)
 
 	if boot.shouldTryToRequestHeaders() {
-		go boot.requestHeadersIfSyncIsStuck()
+		go boot.requestHeadersIfSyncIsStuck(boot.isNodeSynchronized)
```
fbsobreira, line 348: The once-per-slot check happens inside the goroutine, so a not-synced node still spawns a goroutine on every sync-loop iteration, only for it to lose the Swap. Claiming the slot in `shouldTryToRequestHeaders` (already under `mutNodeState`) before the `go` would avoid those spawns.

core/process/sync/baseSync.go
```diff
@@ -358,21 +363,97 @@ func (boot *baseBootstrap) shouldTryToRequestHeaders() bool {
...
+}
+
+// requestHeadersIfSyncIsStuck fires the recovery burst once the node has not
+// committed a block for more than MaxSlotsWithoutNewBlockReceived slots.
+// stalledWhileSynced tells it whether the node believed it was synchronized at
+// the moment the burst was decided; that state deserves a warning, because the
+// node advertises NsSynchronized and MetricIsSyncing reads 0 throughout, while
+// an honestly syncing node passing through here is just catching up.
+func (boot *baseBootstrap) requestHeadersIfSyncIsStuck(stalledWhileSynced bool) {
+	slotDiff := boot.slotsSinceLastCommittedBlock()
```
fbsobreira, line 430: `slotsSinceLastCommittedBlock()` and the threshold check are repeated here after the predicate already did them. Passing the computed lag in, e.g. `requestHeadersIfSyncIsStuck(lag, stalledWhileSynced)`, removes the duplicate work and the duplicate Debug log on underflow.

core/process/sync/baseForkDetector.go
```diff
@@ -616,6 +634,31 @@ func (bfd *baseForkDetector) isConsensusStuck() bool {
...
+// signal. Throttled to once per slot because CheckFork runs on every sync-loop
+// iteration while the node is not synchronized.
+//
+// The format does happen under mutNodeState, since CheckFork's only production
+// caller holds it. That is accepted here, unlike for the stall warning that was
+// moved out of that lock: the throttle caps this at one format per slot, and
+// the state has no out-of-lock observer to move it to.
+func (bfd *baseForkDetector) warnOnceCheckpointAheadOfSlotIndex(currentSlot uint64, checkpointSlot uint64) {
+	slotIndex := bfd.slotManager.Index()
+	if bfd.lastCheckpointAheadWarnSlot.Swap(slotIndex) == slotIndex {
```
fbsobreira, line 652: This is throttled per slot, not once. After a clock step-back it logs a Warn every slot for as long as the state lasts (for example ~150 lines for a 10-minute NTP step), and it formats while holding `mutNodeState`. Latching on the state change instead (warn on entering the state, clear once `slotsDifference` succeeds) gives the same signal in one line.

core/process/sync/metaForkDetector.go
```diff
@@ -47,6 +47,13 @@ func NewMetaForkDetector(
 	bfd.fork.rollBackNonce = math.MaxUint64
 	bfd.fork.probableHighestNonce = bfd.genesisNonce
 	bfd.fork.highestNonceReceived = bfd.genesisNonce
+	// Sentinels below any real slot index: the zero values would otherwise
+	// collide with slot index 0. For the warn throttle that swallows the first
+	// checkpoint-ahead warning; for lastSlotWithForcedFork it makes
+	// isConsensusStuck treat the genesis slot as if a forced fork just happened
+	// and skip its entire body.
+	bfd.lastCheckpointAheadWarnSlot.Store(math.MinInt64)
+	bfd.fork.lastSlotWithForcedFork = math.MinInt64
```
fbsobreira, line 56: Correctness now depends on `MinInt64` sentinels that are seeded only in the meta constructors. Any other constructor (test helpers, a future shard detector) gets zero values and silently drops the first event at slot 0. Changing `lastSlotWithForcedFork`'s initial value also changes behavior: `isConsensusStuck` no longer skips slot 0. Storing `slot+1` (0 = unset) would avoid the per-constructor seeding.

core/process/headerCheck/headerSignatureVerify.go
```diff
@@ -293,9 +295,42 @@ func (hsv *HeaderSigVerifier) getLeader(header data.HeaderHandler) (crypto.Publi
...
 	headerConsensusGroup, err := hsv.nodesCoordinator.ComputeConsensusGroup(prevRandSeed, header.GetSlot(), epoch)
 	if err != nil {
+		logIfEpochConfigMissing(err, header, epoch, "leader")
 		return nil, err
 	}
 
 	leaderPubKeyValidator := headerConsensusGroup[0]
 	return hsv.keyGen.PublicKeyFromByteArray(leaderPubKeyValidator.PubKey())
 }
+
```
fbsobreira, line 305: Nit: several comments in this PR are much longer than the codebase norm:
- the doc comment on `logIfEpochConfigMissing`
- the ~20-line block in `shouldTryToRequestHeaders`
- the doc comment on `warnOnceCheckpointAheadOfSlotIndex`
- the comments that restate the adjacent `log.Warn`

Please trim them to a line or two. The design rationale fits better in the PR description.

fbsobreira (changes requested):
Thanks for tracking this down. Requesting changes for the inline items below.

Also, please **rebase onto the current `develop`**. The branch is 62 commits behind, and several recent merges touch nearby sync and consensus code, so we should review and test against the current base.

The main points are the import-mode guard (it doesn't cover the not-synced path) and the behavior of the new synced trigger during a network-wide halt. The rest are smaller cleanups.

Issue #128, opened 2026-08-11
PeerShardMapper classifies waiting-list validators as observers


MathijsBok:
## Context

Split off from the assessment of #109 as a behaviour issue, deliberately kept out of that pure no-behaviour-change cleanup.

> **Body revised 2026-08-13**, second revision. It now folds in a full assessment made on 2026-08-11 that was never posted here, plus today's re-verification. Two things changed relative to the first version of this issue:
> - The `## Verified consequences` section claimed waiting validators lose connection slots through `listsSharder`. **That was wrong for the shipped configuration.** Corrected below.
> - The sequencing advice ("pick up after #126 merges") was wrong in the other direction. #126 is not a technical dependency. **#109 is the real coupling**, because it deletes the exact helper this fix modifies.

Verified against `develop` at `e48c2c2`.

## Problem

`publicKeyToValidatorMap` is built from `electedList` and `eligibleList` only (`sharding/nodesCoordinator.go:397`, helper at `:419-434`). `waitingList` is never merged in, even though the epoch config carries it (`epochNodesConfig.waitingList`, `:53`, filled by the switch at `:933`).

The lists really are disjoint, so waiting cannot leak in by another route: the switch at `:928-940` puts each validator in exactly one list, and the shuffler only ever receives `Elected` and `Eligible` (`sharding/hashValidatorShuffler.go:85-91`).

`GetValidatorWithPublicKey` (`:437-449`) reads only that map and has exactly one production caller: `PeerShardMapper.getPeerInfoWithNodesCoordinator` (`sharding/networksharding/peerShardMapper.go:127`). Everything else is the interface, the disabled stub (`core/bootstrap/disabled/disabledNodesCoordinator.go:68`) and mocks.

So for a peer whose announced public key belongs to a waiting-list validator:

1. `getPeerInfoWithNodesCoordinator` gets `ErrValidatorNotFound` and returns `UnknownPeer, false` (`peerShardMapper.go:127-133`).
2. `GetPeerInfo` falls through to the fallback cache (`peerShardMapper.go:99-107`, `:141-163`). The heartbeat processor puts every verified sender into `fallbackPidShard` (`node/heartbeat/process/messageProcessor.go:70-72`), so a waiting validator that sends heartbeats reliably lands on **`ObserverPeer`**, not `UnknownPeer`.

The code has been in this shape since the initial import: `git log -L 419,435:sharding/nodesCoordinator.go` returns only `21147fa` (2024-03-13).

[...]

PR #131, opened 2026-08-13
refactor: build the nodesCoordinator public key map in a single pass

### LINKED ISSUE #109: refactor: eliminate double-pass allocation in computePublicKeyToValidatorMap
## Context

Identified during review of PR #108 by @fbsobreira ([comment](https://github.com/klever-io/klever-go/pull/108#discussion_r3741578922)). Intentionally scoped out of that PR to keep the diff focused.

> **Body revised 2026-08-13** after measuring the change and tracing the code path end to end. The proposed fix is unchanged and still correct. What changed is the justification: the original `## Impact` section sold this as a performance win, and that claim did not survive measurement. It is replaced below by what was actually measured. The earlier verification [comment](https://github.com/klever-io/klever-go/issues/109#issuecomment-5256102398) is now folded into this body.

Verified against `develop` at `e48c2c2` on 2026-08-13. `1b59fd5` (#108) is still the last commit touching `sharding/nodesCoordinator.go`, and both functions are unchanged.

## Problem

`computePublicKeyToValidatorMap` (`sharding/nodesCoordinator.go:389-417`) builds the validator lookup map in two passes:

1. First pass: builds `mapAllValidators`, a `map[uint32]map[string]Validator` with one full per-epoch map each (via `createPublicKeyToValidatorMap`).
2. Second pass: iterates the sorted epoch list and copies every entry from each per-epoch map into the final `publicKeyToValidatorMap` result.

The intermediate `mapAllValidators` structure is allocated, populated, and then discarded. The helper `createPublicKeyToValidatorMap` (`sharding/nodesCoordinator.go:419-434`) has exactly one caller, line 397, so it becomes deletable.

## Proposed fix

Sort the epoch list first, then merge `electedList` and `eligibleList` directly into the result map in ascending epoch order.

`slices` is already imported and used at line 592, and `sort` stays in use at lines 951-954, so `slices.Sort` needs no import change and drops the comparator closure:

```go
// computePublicKeyToValidatorMap merges the elected and eligible validators of
// all given epoch configs into one lookup map, later epochs taking precedence
func (ihgs *indexHashedNodesCoordinator) computePublicKeyToValidatorMap(
    nodesConfig map[uint32]*epochNodesConfig,
) map[string]Validator {
    epochList := make([]uint32, 0, len(nodesConfig))
    for epoch := range nodesConfig {
        epochList = append(epochList, epoch)
    }
    slices.Sort(epochList)

    publicKeyToValidatorMap := make(map[string]Validator)
    for _, epoch := range epochList {
        epochConfig := nodesConfig[epoch]
        epochConfig.mutNodesMaps.RLock()
        for _, v := range epochConfig.electedList {
            publicKeyToValidatorMap[string(v.PubKey())] = v
        }
        for _, v := range epochConfig.eligibleList {
            publicKeyToValidatorMap[string(v.PubKey())] = v
        }
        epochConfig.mutNodesMaps.RUnlock()
    }

    return publicKeyToValidatorMap
}
```

## What this is actually worth (measured)

Both implementations were benchmarked side by side over 11 epochs, which is the steady-state retention (`nodeCoordinatorStoredEpochs = 10` plus the epoch being prepared before `EpochStartAction` prunes at line 211). go1.26.5, darwin/arm64, Apple M4 Max, `-benchtime=2s -count=3`, variation across runs under 1%.

Steady state, meaning the validator set is the same in every retained epoch, so every later epoch overwrites every earlier entry:

| elected + eligible | current | proposed | time | memory |
|---|---|---|---|---|
| 21 + 10 | 24.8 us / 59.5 KB | 7.5 us / 10.0 KB | 3.3x | 6.0x |
| 500 + 250 | 535 us / 1.07 MB | 181 us / 210 KB | 3.0x | 5.1x |

Full churn, a completely different validator set every epoch, which is the case least favourable to the change:

| elected + eligible | current | proposed | time | memory |
|---|---|---|---|---|
| 21 + 10 | 32.8 us / 95 KB | 14.4 us / 45.6 KB | 2.3x | 2.1x |
| 500 + 250 | 911 us / 2.37 MB | 573 us / 1.51 MB | 1.6x | 1.6x |

So "halves allocations" understates the ratio. It is roughly 3x the time and 6x the memory in the realistic case.

**The ratio is misleading and the absolute numbers are what matter.** This runs once per epoch boundary (`EpochStartPrepare` -> `fillPublicKeyToValidatorMap`, line 800) and once per node start (`LoadState` line 289, constructor line 124). At `slotsPerEpoch: 5400` and `slotInterval: 4000` an epoch is 6 hours, and at `consensusGroupSize: 21` the saving is **17 microseconds per 6 hours**, plus one restart's worth. Even at 2000 validators it is 2.2 ms versus 0.8 ms.

**This is therefore not a performance fix and should not be approved as one.** What it is worth: 16 lines and one dead helper deleted, on a file that has been edited twice in two months (#108, #126), and a function that becomes readable in a single pass.

## Why this code still deserves care

Worth recording, because it is not obvious from the function and it raises the bar for changes here. `publicKeyToValidatorMap` gates whether gossiped block headers are accepted:

```
publicKeyToValidatorMap                        (nodesCoordinator.go:65)
  -> GetValidatorWithPublicKey                 (nodesCoordinator.go:437)
  -> PeerShardMapper.getPeerInfoWithNodesCoordinator  (peerShardMapper.go:127)
  -> registered as the antiflood peer validator mapper (factory/networkSharding.go:36)
  -> p2pAntiflood.IsOriginatorElectedForTopic  (p2pAntiflood.go:96-111)
  -> SingleDataInterceptor on BlocksTopic      (singleDataInterceptor.go:135-142)
```

`PrepareOpenTopics` opens only `HeartbeatTopic` and `TransactionTopic` (`factory/process.go:773`), so `txBlock` is validator-only, and `antiflood.enabled` is `true` in the shipped config, so this is the real handler and not the disabled stub. A wrong or stale map means a node drops every gossiped block header until the map is rebuilt, which without a restart is the next epoch boundary, up to 6 hours later. That is the same silent-stall shape described in #90 and #130.

The escape hatch is that explicitly requested data is whitelisted (`requestHandlers/requestHandler.go:208`) and bypasses the originator check, which is precisely the path #130 restores.

## Why the change itself is low risk

- **Proven output-equivalent.** Both implementations were compared on steady state, full churn, empty config, empty lists, a duplicate pubkey inside one epoch, and non-contiguous out-of-order epochs, the last repeated 200 times against Go's randomised map iteration order. Identical in every case.
- **Precedence is currently unobservable in production.** The only production caller discards the returned validator: `_, err := psm.nodesCoordinator.GetValidatorWithPublicKey(pkBuff)`. Only the key set is observed, and that key set is the union of all epochs regardless of merge order. The sort must still be kept, because it becomes behaviour-defining the moment any caller uses the returned value, but it means even a precedence regression could not change production behaviour today.
- **No new lock risk.** Ordering stays `mutNodesConfig` -> `mutNodesMaps`, and never more than one `mutNodesMaps` is held at a time, so taking the per-epoch RLocks in ascending epoch order rather than random map order introduces no deadlock path. Work done under the lock is unchanged: the first map insert already happens inside `mutNodesMaps.RLock()` at lines 396-398 today. The write lock on `mutNodesConfig` blocks `ComputeConsensusGroup` (line 517), and this change shortens that window from about 25 us to about 7.5 us once per epoch.

## Plan

1. Apply the fix above and delete `createPublicKeyToValidatorMap`.
2. Add a direct unit test on `computePublicKeyToValidatorMap`. Coverage today is only indirect via `LoadState`. `TestNodesCoordinator_LoadStateMultiEpochPrecedence` pins the cross-epoch rule, but the within-epoch rule (elected written first, then eligible, so eligible wins on a duplicate pubkey) is pinned by nothing. The test must build `map[uint32]*epochNodesConfig` directly rather than going through `SetNodes`, see sequencing below.
3. Verify with `go vet ./sharding/`, `go test ./sharding/... -count=1 -race`, and the existing #108 tests as the regression net.

Size: roughly 30 lines removed and 20 added in production code, plus about 40 lines of test.

## Sequencing

- **#128** (waiting-list validators classified as observers) is directly coupled, and in a way that makes this issue the one to land first: the change here **deletes `createPublicKeyToValidatorMap`**, which is exactly the helper that #128's Option A would modify. Once this lands, that fix shrinks to a third loop over `waitingList` inside the per-epoch block, placed before elected and eligible so those keep precedence within an epoch. Combining both in a single PR is equally fine. What must not happen is the two landing independently in the wrong order.
- **#126** touches the same file but not these functions. Its hunks are at 360, 626-690, 785 and 823-850, so there is no production-code conflict. It does change the `SetNodes` signature and touches `nodesCoordinator_test.go`, which is why the new test should not call `SetNodes`. There is no semantic interaction either: #126 adds `leavingList` to the registry and to `SetNodes`, but not to this map, so the inputs (`electedList`, `eligibleList`) are unchanged. Note that jailed validators promoted to `eligibleList` by `computeNodesConfigFromList` (lines 943-949) already do land in this map, in both versions.
- **#130** does not touch `sharding/` at all. No conflict. As noted above it strengthens the requested-header path that bypasses this map, so it reduces the blast radius of any future defect here.

## Open design question this raises (not in scope here)

The deeper issue is not two passes versus one. It is that accepting a block header from a validator depends on a locally rebuilt public key cache, so any defect in that cache is a silent stall with a detection time measured in hours. #108, #130 and this issue are three symptoms of that one design. Worth its own issue.

MathijsBok:
Closes #109.

## What was wrong

`computePublicKeyToValidatorMap` built its lookup in two passes:

1. one full `map[string]Validator` per epoch, collected into an intermediate `map[uint32]map[string]Validator` via `createPublicKeyToValidatorMap`;
2. a second pass copying every entry of every per-epoch map into the result.

The intermediate structure was allocated, populated and discarded on every call, and `createPublicKeyToValidatorMap` had exactly one caller, so it existed only to feed a structure that was thrown away.

## What changed

**`sharding/nodesCoordinator.go:387-421`** now merges `electedList` and `eligibleList` straight into the result in ascending epoch order, and `createPublicKeyToValidatorMap` is deleted. `sort.Slice` becomes `slices.Sort`: no import change, because `slices` was already imported (used at `:580`) and `sort` stays in use at `:939-942`.

The output is unchanged. Epochs are still merged in ascending order so a later epoch wins, and elected is still written before eligible so eligible wins a duplicate public key.

The doc comment was rewritten to state the *mechanism* rather than the two outcomes: insertion order is what defines precedence, so a list that has to lose against the ones already handled is added before them, never after. That matters for the follow-up in #128, whose `waitingList` loop reads naturally as an append after eligible, which would silently invert precedence with nothing failing. The comment also now states the locking precondition, because the new shape reads `nodesConfig` in two passes and an epoch removed in between would be a nil dereference. Both current callers hold `mutNodesConfig`, so this is documentation of an existing contract, not a new constraint.

## What this is worth, honestly

**This is a readability change, not a performance one.** Measured over 11 epochs (the steady-state retention), the new version is about 3x faster and allocates about 6x less. That ratio is misleading. The function runs once per epoch boundary and once per node start, and at the network's `consensusGroupSize` the saving is roughly **17 microseconds per six-hour epoch**. Please do not approve this as an optimization. What it buys is 35 lines and one dead helper gone from a file that has been edited twice in two months.

## Test

**`sharding/nodesCoordinator_test.go:763-871`** adds `TestNodesCoordinator_computePublicKeyToValidatorMap`. Coverage was previously indirect through `LoadState` only: `TestNodesCoordinator_LoadStateMultiEpochPrecedence` pinned the cross-epoch rule, but the within-epoch rule and the interaction between the two rules were pinned by nothing.

Six subtests, each verified to be load-bearing by applying the corresponding mutant to the production code and confirming the subtest fails:

| Subtest | Mutant it kills |
|---|---|
| `within one epoch eligible takes precedence over elected` | swapping the elected and eligible loops |
| `the highest epoch wins and earlier epochs still contribute` | dropping `slices.Sort`; merging only the highest epoch |
| `the epoch order outranks the list order` | hoisting the loops to "every epoch's elected, then every epoch's eligible" |

That last one is the reason this test exists in this shape. With epoch 5 holding a public key in `eligibleList` and epoch 9 holding it in `electedList`, the hoisted variant returns the stale epoch 5 record while passing every other assertion in the package. The cross-epoch subtests use non-contiguous epochs (2, 9, 11, 40) so that insertion order and lexicographic order both pick a different winner than the correct one, and the assertions repeat 100 times because Go randomises the range start offset on every execution rather than once per process.

## Deliberate decision, so it reads as a choice and not an oversight

`computePublicKeyToValidatorMap` no longer references its receiver. It is kept as a method rather than made package-level, to stay symmetric with `fillPublicKeyToValidatorMap` and to keep the follow-up diff in #128 small. `.golangci.yml` enables only `unused, gosec, govet, staticcheck, ineffassign, errcheck, errorlint`, so revive's `unused-receiver` does not run and CI will not flag it. Happy to change it if a reviewer prefers the free function.

## Note on the coverage number

Package coverage for `sharding` moves from 70.4% to 70.0%. That is arithmetic, not lost tests: roughly 13 fully covered statements were deleted and no uncovered ones were added. Coverage of `computePublicKeyToValidatorMap` itself is 100% before and after.

## Sequencing

- **#128** must land after this. It proposes extending `createPublicKeyToValidatorMap`, which this PR deletes. Once this is in, that fix becomes a third loop over `waitingList` inside the per-epoch block, placed before elected and eligible so those keep precedence. A comment restating that is posted on #128.
- **#126** touches the same file but different hunks and does not route `leavingList` into this map, so the inputs here are unchanged. Verified by trial merge rather than by inspection: merging `fix/116-jailed-peer-type` into this branch is conflict-free and the merged tree builds and passes `go test ./sharding/...`. The new test deliberately builds `map[uint32]*epochNodesConfig` directly instead of calling `SetNodes`, whose signature #126 changes, which is what keeps that merge clean.

## Verification

| Check | Result |
|---|---|
| `gofmt -l sharding/` | clean |
| `go vet ./sharding/...` | clean |
| `go build ./...` | clean |
| `go test ./sharding/... ./core/process/peer/... ./node/heartbeat/... -count=1 -race` | all ok |
| Coverage of the changed function | 100% |
| SonarQube open issues on the touched files | 0 |
| Equivalence of old and new implementation | differential fuzz over randomised configs, no divergence |
| Mutation tests | 4 mutants applied, all 4 killed |
| Trial merge with #126 | conflict-free, builds, tests pass |

## Summary

- Refactors `computePublicKeyToValidatorMap` to build the validator map in one pass.
- Preserves validator precedence across epochs and within each epoch.
- Removes obsolete intermediate maps and `createPublicKeyToValidatorMap`.
- Uses `slices.Sort` for epoch ordering.
- Adds tests for ordering, precedence, empty inputs, and validator retention.

## Impact

- Affects sharding node coordination and validator selection.
- Supports consensus-related state without changing transaction processing, KVM, or networking.
- Preserves data integrity through explicit precedence rules.
- Retains per-epoch read-lock protection. The refactor introduces no reported concurrency or error-handling changes.
- Reported formatting, build, vet, test, race, fuzz, mutation, and merge checks pass.

### DIFF
--- a/sharding/nodesCoordinator.go
+++ b/sharding/nodesCoordinator.go
@@ -384,50 +384,36 @@ func (ihgs *indexHashedNodesCoordinator) fillPublicKeyToValidatorMap() {
 	ihgs.publicKeyToValidatorMap = ihgs.computePublicKeyToValidatorMap(ihgs.nodesConfig)
 }
 
-// computePublicKeyToValidatorMap merges the elected and eligible validators of
-// all given epoch configs into one lookup map, later epochs taking precedence
+// computePublicKeyToValidatorMap merges the elected and eligible validators of all
+// given epoch configs into one lookup map. Insertion order is what defines precedence,
+// the last write for a public key winning: epochs are merged in ascending order so a
+// later epoch wins, and elected is written before eligible so eligible wins. A list
+// that has to lose against the ones already handled is added before them, never after.
+//
+// The caller must hold ihgs.mutNodesConfig. nodesConfig is read in two passes, so an
+// epoch removed in between the passes would be a nil dereference.
 func (ihgs *indexHashedNodesCoordinator) computePublicKeyToValidatorMap(
 	nodesConfig map[uint32]*epochNodesConfig,
 ) map[string]Validator {
-	index := 0
-	epochList := make([]uint32, len(nodesConfig))
-	mapAllValidators := make(map[uint32]map[string]Validator)
-	for epoch, epochConfig := range nodesConfig {
-		epochConfig.mutNodesMaps.RLock()
-		mapAllValidators[epoch] = ihgs.createPublicKeyToValidatorMap(epochConfig.electedList, epochConfig.eligibleList)
-		epochConfig.mutNodesMaps.RUnlock()
-
-		epochList[index] = epoch
-		index++
+	epochList := make([]uint32, 0, len(nodesConfig))
+	for epoch := range nodesConfig {
+		epochList = append(epochList, epoch)
 	}
 
-	sort.Slice(epochList, func(i, j int) bool {
-		return epochList[i] < epochList[j]
-	})
+	slices.Sort(epochList)
 
 	publicKeyToValidatorMap := make(map[string]Validator)
 	for _, epoch := range epochList {
-		validatorsForEpoch := mapAllValidators[epoch]
-		for pubKey, vInfo := range validatorsForEpoch {
-			publicKeyToValidatorMap[pubKey] = vInfo
+		epochConfig := nodesConfig[epoch]
+
+		epochConfig.mutNodesMaps.RLock()
+		for _, validator := range epochConfig.electedList {
+			publicKeyToValidatorMap[string(validator.PubKey())] = validator
 		}
-	}
-
-	return publicKeyToValidatorMap
-}
-
-func (ihgs *indexHashedNodesCoordinator) createPublicKeyToValidatorMap(
-	elected []Validator,
-	eligible []Validator,
-) map[string]Validator {
-	publicKeyToValidatorMap := make(map[string]Validator)
-
-	for i := 0; i < len(elected); i++ {
-		publicKeyToValidatorMap[string(elected[i].PubKey())] = elected[i]
-	}
-
-	for i := 0; i < len(eligible); i++ {
-		publicKeyToValidatorMap[string(eligible[i].PubKey())] = eligible[i]
+		for _, validator := range epochConfig.eligibleList {
+			publicKeyToValidatorMap[string(validator.PubKey())] = validator
+		}
+		epochConfig.mutNodesMaps.RUnlock()
 	}
 
 	return publicKeyToValidatorMap
--- a/sharding/nodesCoordinator_test.go
+++ b/sharding/nodesCoordinator_test.go
@@ -760,6 +760,115 @@ func TestNodesCoordinator_IsInterfaceNil(t *testing.T) {
 	require.False(t, check.IfNil(ihgs3))
 }
 
+func TestNodesCoordinator_computePublicKeyToValidatorMap(t *testing.T) {
+	t.Parallel()
+
+	makeValidator := func(pubKey string, index uint32) Validator {
+		pk := []byte(pubKey)
+		return mock.NewValidatorMock(pk, pk, DefaultSelectionChances, index)
+	}
+
+	ihgs := &indexHashedNodesCoordinator{}
+
+	t.Run("empty config yields an empty map", func(t *testing.T) {
+		t.Parallel()
+
+		require.Len(t, ihgs.computePublicKeyToValidatorMap(map[uint32]*epochNodesConfig{}), 0)
+	})
+
+	t.Run("epoch without validators yields an empty map", func(t *testing.T) {
+		t.Parallel()
+
+		require.Len(t, ihgs.computePublicKeyToValidatorMap(map[uint32]*epochNodesConfig{7: {}}), 0)
+	})
+
+	t.Run("elected and eligible are both present", func(t *testing.T) {
+		t.Parallel()
+
+		nodesConfig := map[uint32]*epochNodesConfig{
+			3: {
+				electedList:  []Validator{makeValidator("elected_pk", 1)},
+				eligibleList: []Validator{makeValidator("eligible_pk", 2)},
+			},
+		}
+
+		publicKeyToValidatorMap := ihgs.computePublicKeyToValidatorMap(nodesConfig)
+
+		require.Len(t, publicKeyToValidatorMap, 2)
+		require.Contains(t, publicKeyToValidatorMap, "elected_pk")
+		require.Contains(t, publicKeyToValidatorMap, "eligible_pk")
+		require.Equal(t, uint32(1), publicKeyToValidatorMap["elected_pk"].Index())
+		require.Equal(t, uint32(2), publicKeyToValidatorMap["eligible_pk"].Index())
+	})
+
+	t.Run("within one epoch eligible takes precedence over elected", func(t *testing.T) {
+		t.Parallel()
+
+		sharedPK := "pk_present_in_both_lists"
+		nodesConfig := map[uint32]*epochNodesConfig{
+			3: {
+				electedList:  []Validator{makeValidator(sharedPK, 1)},
+				eligibleList: []Validator{makeValidator(sharedPK, 2)},
+			},
+		}
+
+		publicKeyToValidatorMap := ihgs.computePublicKeyToValidatorMap(nodesConfig)
+
+		require.Len(t, publicKeyToValidatorMap, 1)
+		require.Contains(t, publicKeyToValidatorMap, sharedPK)
+		require.Equal(t, uint32(2), publicKeyToValidatorMap[sharedPK].Index())
+	})
+
+	t.Run("the highest epoch wins and earlier epochs still contribute", func(t *testing.T) {
+		t.Parallel()
+
+		sharedPK := "pk_present_in_every_epoch"
+		earlyOnlyPK := "pk_present_only_in_the_earliest_epoch"
+		// non-contiguous epochs, so ordering by insertion or lexicographically would pick another one
+		nodesConfig := map[uint32]*epochNodesConfig{
+			2:  {electedList: []Validator{makeValidator(sharedPK, 2), makeValidator(earlyOnlyPK, 200)}},
+			9:  {electedList: []Validator{makeValidator(sharedPK, 9)}},
+			11: {electedList: []Validator{makeValidator(sharedPK, 11)}},
+			40: {electedList: []Validator{makeValidator(sharedPK, 40)}},
+		}
+
+		// the range start offset is randomised on every execution, so repeat to make an
+		// order-dependent regression fail reliably instead of flaking
+		for i := 0; i < 100; i++ {
+			publicKeyToValidatorMap := ihgs.computePublicKeyToValidatorMap(nodesConfig)
+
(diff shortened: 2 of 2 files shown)


fbsobreira (approved):
Approving. Reviewed as consensus-relevant (this map feeds `GetValidatorWithPublicKey` → `PeerShardMapper` peer classification → message admission), and verified the refactor is output-equivalent.

**Output equivalence — proven, not argued.** Reconstructed the pre-refactor implementation verbatim and ran a differential test: 3000 randomized epoch configs with a deliberately small key space so within-epoch (elected∩eligible) and cross-epoch collisions occur constantly. Old vs new produced identical maps every time — same key set and same winning validator per key (pointer identity, not just index). The harness is discriminating: inverting the elected/eligible write order fails it immediately, so the pass is real. This matches the ascending-epoch + elected-before-eligible last-write-wins ordering being preserved.

**Only new precondition is satisfied.** The new version reads the outer `nodesConfig` twice (list epochs, then merge) where the old snapshotted in one pass, so a concurrent epoch removal between passes would nil-deref. Both direct callers hold `mutNodesConfig` across the call — `LoadState` (287–295) and `fillPublicKeyToValidatorMap` (381–384) — so no epoch can be removed mid-function. The doc comment documents a real, held contract.

**Consumer note:** the sole production consumer (`peerShardMapper.go:127`) discards the validator value and branches only on key presence, so precedence can't affect it regardless.

Verified locally at head `a2f54dd`: `go build`/`go vet` clean, `slices`/`sort` both still used (no import churn), full `sharding/...` clean under `-race`, and the new within-epoch precedence subtest is load-bearing (dies under the swap mutant).

No blocking or in-scope issues. Sequencing is correct: #128 must land after this, #126 touches disjoint hunks.

PR #134, opened 2026-08-14
refactor: replace hand-built SliceHeader in wasmer2 memory view with unsafe.Slice

### LINKED ISSUE #129: kvm/wasmer2: go vet unsafe.Pointer findings are suppressed rather than resolved (dead code, SliceHeader construction)
## What this is, and what it is not

`go vet ./kvm/wasmer2/` reports findings on `develop` that already carry a `#nosec G103` suppression, so the static-analysis signal has been silenced rather than resolved. This is a code-hygiene issue in a safety-sensitive package: nobody will look at these lines again on their own, and the suppressions also mask any future vet finding on the same lines.

**This is not a reported vulnerability.** No attacker-reachable path was found, and none is claimed.

The two items below are safe, mechanical cleanups. They are listed separately because they differ in effort and in how confident this report is about each. A third vet finding in this package needs a deeper lifetime review and is being tracked separately, not here.

---

## Item 1: dead code (verified, zero risk)

`kvm/wasmer2/wasmer2InstanceMemory.go:30-32`

```go
var header reflect.SliceHeader
// #nosec G103: unsafe.Pointer is used to convert to reflect.SliceHeader
header = *(*reflect.SliceHeader)(unsafe.Pointer(&header))
```

The second statement takes the address of `header`, casts it back to `*reflect.SliceHeader`, dereferences it, and assigns the result to `header` itself. It is a self-assignment of a freshly declared zero value, which is then fully overwritten on lines 35 to 37 anyway. It does nothing.

It reads like a copy-paste remnant of the pattern where a `SliceHeader` is read out of an *existing* slice, where the source and destination are different variables.

**Action:** delete the line and its suppression. Accounts for two of the vet findings (`32:11` and `32:50`).

**Verification status:** read directly, no behavioural change possible.

---

## Item 2: `reflect.SliceHeader` construction (verified, low risk, mechanical)

`kvm/wasmer2/wasmer2InstanceMemory.go:35-40`, the body of `Wasmer2Memory.Data()`

```go
header.Data = uintptr(unsafe.Pointer(data))
header.Len = int(length)
header.Cap = int(length)

return *(*[]byte)(unsafe.Pointer(&header))
```

This builds a `[]byte` view over WebAssembly linear memory owned by wasmer on the C side. `go vet` flags it because a hand-built `reflect.SliceHeader` is invisible to the garbage collector: the `Data` field is a `uintptr` and keeps nothing alive.

In this specific case the referent is C-owned, so the GC was never going to be involved, which is why the pattern works in practice. It is still the officially discouraged form, and `unsafe.Slice((*byte)(data), length)` expresses exactly the same thing while being vet-clean. The module is on Go 1.25, so `unsafe.Slice` is available.

Two consumers were checked and both are well behaved: `MemLoadFromMemory` (`kvm/executor/executorMemory.go:37`) and `MemStoreToMemory` (`:73`) fetch `Data()`, bounds-check against `Length()`, and copy in or out within the same call without retaining the slice. So the aliasing window is short and bounded at those sites.

**Action:** replace with `unsafe.Slice`, drop the suppression. Accounts for the `40:35` finding.

**Verification status:** the two call sites named above were read. Other consumers of `executor.Memory.Data()` outside `kvm/executor` were not enumerated, so a retained-slice consumer elsewhere is not ruled out.

---

## Steps to reproduce

```bash
go vet ./kvm/wasmer2/
```

## Environment

- Go version: 1.25.7 (per `go.mod`)
- Node version: `develop` @ `e48c2c2`
- Applies to all node types; this is build-time static analysis, not runtime behaviour.

## Suggested handling

Items 1 and 2 are safe to do as one small mechanical PR that removes the `#nosec` suppressions on those lines and makes those findings clean, which also stops them from masking any future vet finding in the package.

MathijsBok:
## Summary

Resolves the two mechanical items of #129: the `go vet` findings on the hand-built `reflect.SliceHeader` in `Wasmer2Memory.Data()` are resolved instead of suppressed.

## What changed, per item

**Item 1, dead code (`kvm/wasmer2/wasmer2InstanceMemory.go`)**
The self-assignment `header = *(*reflect.SliceHeader)(unsafe.Pointer(&header))` was a no-op on a freshly declared zero value, fully overwritten directly after, and carried two `#nosec G103` suppressions. Deleted.

**Item 2, SliceHeader construction (`kvm/wasmer2/wasmer2InstanceMemory.go:24-32`)**
`Data()` now builds the `[]byte` view over the wasmer-owned linear memory with `unsafe.Slice(data, length)`, the canonical replacement for the `reflect.SliceHeader` pattern (deprecated since Go 1.21, staticcheck SA1019). The `reflect` import and the blanket `// nolint` are gone.

**Ride-alongs in the same file**
- The empty `Destroy()` body now carries a nested comment explaining the ownership. This resolves the open SonarQube `go:S1186` issue on this file (open since 2024).
- The type doc comment misnamed `Wasmer2Memory` as "Wasmer2Instance"; corrected.

## Why this is safe

- The construction is semantically identical for every reachable input class: non-empty memory, empty memory, and the nil/zero pairing all produce byte-identical views (`unsafe.Slice` sets cap equal to len, matching the old explicit `Cap` assignment).
- The only divergent case, a nil base pointer with a non-zero length, requires the prebuilt wasmer library to contradict its own documented NULL contract. The old code produced a poisoned slice and an undefined-behaviour segfault under wasmer's signal handlers; the new code fails with a deterministic panic that lands in the existing recover path (`kvm/vmhost/hostCore/host.go:502` and `:575`, `ErrExecutionPanicked`), so the node survives and the transaction fails.
- No gas, output, or state change; old and new binaries interoperate on a mixed network, so no epoch gate is needed.
- All four consumers of `Data()` were enumerated: `executor.MemLoadFromMemory` and `executor.MemStoreToMemory` (`kvm/executor/executorMemory.go:37`, `:73`; both bounds-checked and copying), test-only `MemDump`, and a mock that never touches this type. Every production path copies under the instance mutex behind the cleaned-instance guards, and none retains the view.

## Static-analysis delta, stated honestly

- Resolved: the three `go vet` SliceHeader findings and the staticcheck SA1019 deprecation.
- Replaced: one blanket `// nolint` plus three `#nosec G103` comments make way for a single targeted `#nosec G103` with a written justification on the `unsafe.Slice` line, because gosec's audit rule also matches `unsafe.Slice`. Zero open linter findings on the changed files under the repo config.
- `go vet ./kvm/wasmer2/` now reports only the context-pointer finding in `wasmer2ImportsCgoHelper.go`, which is tracked separately (see #129).

## Test that pins the behaviour

New `kvm/wasmer2/wasmer2InstanceMemory_test.go` (`TestWasmer2Memory_DataRoundTrip`) runs against a live native instance: a store/load round-trip through separately obtained views (proves `Data()` returns a live aliasing view, not a per-call copy), `len` and `cap` both equal to `Length()`, and a re-read after the no-op `Destroy()`. This brings the changed lines to 100% in-package coverage under the CI coverage command.

## Verification

- `go build ./...`, `gofmt`, and the CI-identical `go vet -unsafeptr=false` over all packages: clean.
- `go test -count=1` on `kvm/wasmer2` (also with `-race`, which enables checkptr), `kvm/executor/...`, and the full `kvm/vmhost/hosttest` contract suite: green.
- SonarQube analysis of both final files: 0 issues; the pre-existing open `S1186` on this file is resolved by this PR.
- Audited pre-push by four independent review passes (security, code review, Go specialist, and an adversarial full-diff pass); every finding was fixed before this PR was opened.

Closes #129

## Summary

- Updates the KVM `kvm/wasmer2` memory bridge.
- Replaces manual `reflect.SliceHeader` construction with `unsafe.Slice`.
- Removes unused code, imports, suppressions, and blanket lint directives.
- Documents native Wasmer memory ownership and corrects `Wasmer2Memory` documentation.
- Adds a parallelized native-instance round-trip test for aliasing, length, capacity, cleanup, and post-`Destroy()` reads.
- Resolves related `go vet`, staticcheck, and SonarQube findings while retaining targeted `#nosec G103` justification.

## Impact

- Consensus, transaction processing, state management, and networking logic are unchanged.
- The change affects KVM memory access only.
- The memory view preserves existing pointer, length, capacity, and aliasing behavior.
- The test verifies data integrity and continued read access after the no-op `Destroy()` call.
- Native Wasmer retains ownership of the memory and releases it during `Destroy()`.
- No node stability, error-handling, or production concurrency behavior changes.
- Build, formatting, vet, race tests, package tests, and host contract-suite checks pass.

### DIFF
--- a/kvm/wasmer2/wasmer2InstanceMemory.go
+++ b/kvm/wasmer2/wasmer2InstanceMemory.go
@@ -26,8 +26,10 @@ func (memory *Wasmer2Memory) Data() []byte {
 	var data = (*uint8)(cWasmerMemoryData(memory.cgoInstance))
 
 	// #nosec G103: bounded []byte view over wasmer-owned linear memory;
-	// pointer and length come from the same live instance and every caller
-	// copies in or out under the instance mutex without retaining the view.
+	// pointer and length come from the same live instance. The production
+	// callers (MemLoadFromMemory, MemStoreToMemory) copy in or out under the
+	// instance mutex without retaining the view; test-only MemDump returns
+	// the view directly but only within the instance lifetime.
 	return unsafe.Slice(data, length)
 }
 
--- a/kvm/wasmer2/wasmer2InstanceMemory.go
+++ b/kvm/wasmer2/wasmer2InstanceMemory.go
@@ -2,7 +2,6 @@ package wasmer2
 
 import (
 	"fmt"
-	"reflect"
 	"unsafe"
 
 	"github.com/klever-io/klever-go/kvm/executor"
@@ -10,7 +9,7 @@ import (
 
 var _ = (executor.Memory)((*Wasmer2Memory)(nil))
 
-// Wasmer2Instance represents a WebAssembly instance.
+// Wasmer2Memory represents the memory of a WebAssembly instance.
 type Wasmer2Memory struct {
 	// The underlying WebAssembly instance.
 	cgoInstance *cWasmerInstanceT
@@ -22,22 +21,14 @@ func (memory *Wasmer2Memory) Length() uint32 {
 }
 
 // Data returns a slice of bytes over the WebAssembly memory.
-// nolint
 func (memory *Wasmer2Memory) Data() []byte {
 	var length = memory.Length()
 	var data = (*uint8)(cWasmerMemoryData(memory.cgoInstance))
 
-	var header reflect.SliceHeader
-	// #nosec G103: unsafe.Pointer is used to convert to reflect.SliceHeader
-	header = *(*reflect.SliceHeader)(unsafe.Pointer(&header))
-
-	// #nosec G103: unsafe.Pointer is used to convert to uintptr
-	header.Data = uintptr(unsafe.Pointer(data))
-	header.Len = int(length)
-	header.Cap = int(length)
-
-	// #nosec G103: unsafe.Pointer is used to convert to []byte
-	return *(*[]byte)(unsafe.Pointer(&header))
+	// #nosec G103: bounded []byte view over wasmer-owned linear memory;
+	// pointer and length come from the same live instance and every caller
+	// copies in or out under the instance mutex without retaining the view.
+	return unsafe.Slice(data, length)
 }
 
 // Grow the memory by a number of pages (65kb each).
@@ -62,6 +53,8 @@ func (memory *Wasmer2Memory) Grow(numberOfPages uint32) error {
 
 // Destroy destroys inner memory. Does nothing in wasmer2.
 func (memory *Wasmer2Memory) Destroy() {
+	// Intentionally empty: the linear memory is owned by the native wasmer
+	// instance and is released by cWasmerInstanceDestroy via Clean.
 }
 
 // IsInterfaceNil returns true if underlying object is nil
--- a/kvm/wasmer2/wasmer2InstanceMemory_test.go
+++ b/kvm/wasmer2/wasmer2InstanceMemory_test.go
@@ -1,0 +1,45 @@
+package wasmer2
+
+import (
+	"testing"
+
+	"github.com/klever-io/klever-go/kvm/executor"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestWasmer2Memory_DataRoundTrip pins the behaviour of the []byte view that
+// Data() builds over the wasmer-owned linear memory: bytes stored through the
+// view must read back identically, and the view must span the full memory.
+func TestWasmer2Memory_DataRoundTrip(t *testing.T) {
+	t.Parallel()
+
+	instance := newLiveInstance(t)
+	defer func() {
+		require.True(t, instance.Clean())
+	}()
+
+	memoryLength := instance.MemLength()
+	require.NotZero(t, memoryLength)
+
+	payload := []byte{0xDE, 0xAD, 0xBE, 0xEF}
+	require.NoError(t, instance.MemStore(executor.MemPtr(0), payload))
+
+	loaded, err := instance.MemLoad(executor.MemPtr(0), executor.MemLength(len(payload)))
+	require.NoError(t, err)
+	assert.Equal(t, payload, loaded)
+
+	dump := instance.MemDump()
+	require.Equal(t, int(memoryLength), len(dump))
+	require.Equal(t, int(memoryLength), cap(dump))
+	assert.Equal(t, payload, dump[:len(payload)])
+
+	// Destroy is a documented no-op; the direct call (bypassing the instance
+	// mutex) is safe because this test is sequential and the body is empty.
+	// The memory must remain readable until Clean destroys the instance.
+	instance.memory.Destroy()
+
+	reloaded, err := instance.MemLoad(executor.MemPtr(0), executor.MemLength(len(payload)))
+	require.NoError(t, err)
+	assert.Equal(t, payload, reloaded)
+}
--- a/kvm/wasmer2/wasmer2InstanceMemory_test.go
+++ b/kvm/wasmer2/wasmer2InstanceMemory_test.go
@@ -34,12 +34,19 @@ func TestWasmer2Memory_DataRoundTrip(t *testing.T) {
 	require.Equal(t, int(memoryLength), cap(dump))
 	assert.Equal(t, payload, dump[:len(payload)])
 
+	// Prove the returned view aliases live linear memory rather than copying
+	// it: a write through the view must be observable via MemLoad.
+	dump[0] ^= 0xFF
+	loaded, err = instance.MemLoad(executor.MemPtr(0), executor.MemLength(len(payload)))
+	require.NoError(t, err)
+	assert.Equal(t, dump[:len(payload)], loaded)
+
 	// Destroy is a documented no-op; the direct call (bypassing the instance
-	// mutex) is safe because this test is sequential and the body is empty.
-	// The memory must remain readable until Clean destroys the instance.
+	// mutex) is safe because the instance is local to this test goroutine and
+	// the body is empty. Memory must stay readable until Clean destroys it.
 	instance.memory.Destroy()
 
 	reloaded, err := instance.MemLoad(executor.MemPtr(0), executor.MemLength(len(payload)))
 	require.NoError(t, err)
-	assert.Equal(t, payload, reloaded)
+	assert.Equal(t, dump[:len(payload)], reloaded)
 }


kvm/wasmer2/wasmer2InstanceMemory_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/klever-io/klever-go/kvm/executor"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestWasmer2Memory_DataRoundTrip pins the behaviour of the []byte view that
+// Data() builds over the wasmer-owned linear memory: bytes stored through the
+// view must read back identically, and the view must span the full memory.
+func TestWasmer2Memory_DataRoundTrip(t *testing.T) {
```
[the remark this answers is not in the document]
MathijsBok: Valid finding, applied in 1165e28.

What was wrong: the round-trip only proved that writes through the MemStore view are visible through the MemLoad and MemDump views; every assertion compared against the original payload, so a Data() implementation returning a fresh copy of the memory contents would still have passed the dump assertions.

What changed (kvm/wasmer2/wasmer2InstanceMemory_test.go, lines 36 to 46): the test now flips dump[0] through the returned view and reads it back via MemLoad, asserting equality with the mutated dump both before and after the no-op Destroy, as suggested. This pins that Data() returns a live alias of linear memory rather than a copy.

Verification: go test -count=1 -race ./kvm/wasmer2/ is green (checkptr active under -race), golangci-lint with the repo config reports 0 issues, and the changed lines remain at 100 percent in-package coverage.

kvm/wasmer2/wasmer2InstanceMemory.go
```diff
@@ -22,22 +21,16 @@ func (memory *Wasmer2Memory) Length() uint32 {
...
-	header.Cap = int(length)
-
-	// #nosec G103: unsafe.Pointer is used to convert to []byte
-	return *(*[]byte)(unsafe.Pointer(&header))
+	// #nosec G103: bounded []byte view over wasmer-owned linear memory;
+	// pointer and length come from the same live instance. The production
+	// callers (MemLoadFromMemory, MemStoreToMemory) copy in or out under the
+	// instance mutex without retaining the view; test-only MemDump returns
+	// the view directly but only within the instance lifetime.
+	return unsafe.Slice(data, length)
```
MathijsBok, line 33: Audit note (automated self-review), properties verified before posting:

1. `unsafe.Slice(data, length)` builds the same `{ptr, len, len}` header the removed `reflect.SliceHeader` code built; `cap == len` exactly as before, pinned by the new test.
2. The one divergent input class, a nil pointer with non-zero length, requires the native library to contradict its documented NULL contract (`libvmexeccapi.h`: `memory_data` returns NULL and `memory_data_length` returns 0 for a null memory). It now fails as a deterministic panic that lands in the existing recover layer (`kvm/vmhost/hostCore/host.go:502` and `:575`, `ErrExecutionPanicked`), where the old code produced undefined behaviour under the installed signal handlers.
3. Caller enumeration is complete: `MemLoadFromMemory` and `MemStoreToMemory` copy in or out under the instance mutex behind the cleaned-instance guards; test-only `MemDump` returns the live view within the instance lifetime, exactly as the suppression comment above states.
4. `go test -race` (checkptr validates this construction at runtime) is green; the changed lines are at 100 percent in-package coverage under the CI coverage command; `golangci-lint` with the repo config reports 0 issues on this file.

No behavioural, gas, or state difference exists between old and new binaries; mixed-version operation is safe and no epoch gate is needed.

MathijsBok (commented):
## Automated self-review (posted for transparency)

This PR was audited with a multi-agent review (code quality, security, code metrics; only confidence 8+ findings), on top of the pre-push gate that ran before the PR was opened. This comment is the audit outcome, posted so the human reviewer can see what was already checked and fixed. It is a self-review driven by the PR author, not an independent human review.

### Findings

| ID | Severity | Finding | Resolution |
|----|----------|---------|------------|
| R1/S1 | suggestion | The #nosec G103 justification claimed every caller copies the view; test-only MemDump returns the live view, so the claim was too absolute for a security suppression | Fixed in 1165e28: claim scoped to the production callers, MemDump exception named |
| R2 | suggestion | The comment on the direct Destroy call said the test is sequential, contradicting t.Parallel() | Fixed in 1165e28: states the actual safety basis (instance local to the test goroutine, empty body) |
| CR | minor (CodeRabbit) | The round-trip did not pin that the returned slice itself aliases linear memory; a copying implementation would have passed | Fixed in 1165e28: a byte is flipped through the dump view and read back via MemLoad, before and after the no-op Destroy |

Zero critical and zero warning findings. Both suggestions and the CodeRabbit finding were fixed before this review was posted.

### Verified properties (detail in the inline note on the unsafe.Slice line)

- Semantic equivalence with the removed SliceHeader construction for every reachable input class; cap equals len as before.
- The only divergent input class (nil pointer with non-zero length, which requires the native library to contradict its documented NULL contract) now fails as a deterministic panic into the existing recover layer instead of undefined behaviour.
- Complete caller enumeration; all production paths copy under the instance mutex behind the cleaned-instance guards.
- Metrics: Data() goes from 8 to 3 executable statements, from 4 unsafe conversion sites to 1, from 4 lint suppressions to 1 motivated one, and from 0 to 100 percent in-package coverage of the changed lines.
- go test with -race (checkptr active) green; golangci-lint under the repo config reports 0 issues; go vet on the package now reports only the pre-existing helper finding that is tracked separately.

kvm/wasmer2/wasmer2InstanceMemory.go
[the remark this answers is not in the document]
MathijsBok: Verified against the actual toolchain and skipping the move, with evidence:

1. The suppression is demonstrably active in its current position: `golangci-lint run --no-config --default=none --enable=gosec ./kvm/wasmer2/` reports no G103 on this file, and before this directive was added the same command reported G103 at exactly this line. So the leading-comment placement works with the gosec bundled in golangci-lint 2.12.2 (the toolchain CI runs), which associates a leading comment group with the statement it documents (the comment-map handling your own analysis chain cites).

2. Leading-line `#nosec` is the repo-wide convention: every existing directive in this package sits on the line above the flagged statement (see wasmer2Executor.go), including the pre-refactor version of this very function, whose suppressions were provably effective. Moving this one occurrence to end-of-line would deviate from the codebase style without changing behavior, and a three-line justification does not fit a trailing comment.

3. The `--` separator is only enforced under `nosec-require-justification`, which `.golangci.yml` does not enable, as your finding itself notes.

Residual risk assessment: if a future gosec version stopped honoring leading placement, the failure mode is a visible new finding rather than a silently missing suppression, so nothing can rot unnoticed. Leaving this thread open for the human reviewer to weigh in.

fbsobreira, line 33: Non-blocking. The justification states the wrong invariant. It says the view is usable "within the instance lifetime", but `libvmexeccapi.h` (lines 428-429) says the data pointer can be invalidated when memory grows. The real contract is: valid until the next memory grow or `Clean`, and consumed immediately. The comment also names `MemLoadFromMemory` / `MemStoreToMemory` from another file, which will silently rot on a rename, and every other `#nosec G103` in this package is a short one-line note.

Suggest stating the invariant instead of the caller list:

```suggestion
	// #nosec G103: bounded view over wasmer-owned linear memory. The pointer is
	// only valid until the next memory grow or Clean, so callers must consume
	// the view immediately and never retain it.
	return unsafe.Slice(data, length)
```

kvm/wasmer2/wasmer2InstanceMemory_test.go
```diff
@@ -0,0 +1,52 @@
...
+	assert.Equal(t, dump[:len(payload)], loaded)
+
+	// Destroy is a documented no-op; the direct call (bypassing the instance
+	// mutex) is safe because the instance is local to this test goroutine and
+	// the body is empty. Memory must stay readable until Clean destroys it.
+	instance.memory.Destroy()
+
+	reloaded, err := instance.MemLoad(executor.MemPtr(0), executor.MemLength(len(payload)))
+	require.NoError(t, err)
+	assert.Equal(t, dump[:len(payload)], reloaded)
```
fbsobreira, line 51: Non-blocking. This section asserts that an empty function does nothing: `Destroy` has zero callers anywhere in the repo, and the assertion at line 51 repeats the one at line 42 with only the empty call in between. If `Destroy` ever gains a body, this becomes the one call site that touches native state outside the instance mutex, which the comment itself concedes is only safe because the body is empty.

Suggest dropping the section (or reducing it to a bare `instance.memory.Destroy()` without the mutex justification):

```suggestion
```

fbsobreira (commented):
Verified locally on the PR head and on a clean merge with current `develop`: plain `go vet` on the package now reports only the pre-existing `wasmer2ImportsCgoHelper.go:13` finding, the package and executor tests pass with and without `-race`, `golangci-lint` under the repo config is clean, and gosec honours the `#nosec G103` in its current position (the rule does fire on `unsafe.Slice` when suppressions are disabled, so the directive is warranted). The `unsafe.Slice` construction is byte-identical to the old header for every reachable input, and the nil-pointer-with-length case now recovers instead of throwing an unrecoverable fault for offsets beyond the nil page.

Two non-blocking suggestions inline, both about comment accuracy and test cleanup rather than correctness.

One note for the record: CI runs `go vet -unsafeptr=false`, and this PR does not change that. The remaining helper finding is a different class (a `uintptr` context handle cast back to a Go pointer) and stays tracked separately per #129, so the CI flag has to stay until that one is resolved.

fbsobreira (approved):

PR #139, opened 2026-08-17
fix(node): require auth on /node/debug and surface key-file load errors

fbsobreira:
## Summary

Remediation for the three items reported in `GHSA-h97x-7h5v-wjx4`, triaged as **security hardening** rather than a published advisory. All three are bounded by the loopback default the node ships with; none is an auth bypass or a fund/consensus issue.

Companion to #138, which adds the policy these were classified under. **Merge this first** — #138 states that the shipped YAML is a developer checkout rather than a hardening guide, which should not land while `/debug` is still open.

### 1. `/node/debug` now requires authentication

`config/node/api.yaml` registered `/debug` as `open: true` with no `secured` key — four lines above `/log`, which has one. The interceptor/resolver debug store is enabled by default (`config.yaml`, `cacheSize: 10000`), so the route answered unauthenticated with cached request/resolve state.

This is a default-posture gap, not an auth bypass: the route wrapper enforces `secured: true` correctly, `/debug` was simply never marked. Marked it, matching `/log`.

Exposed data is operational telemetry — event type, topic, hash, counters, last error, timestamp — with no peer identifier.

### 2. `getSkPk()` returns key-file load errors

`factory/cryptoSigningParams.go` only returned an error for the file-not-found case. Any other failure — corrupt pem, wrong `KEY_PASSWORD`, unreadable file — returned **`nil` error with empty key material**, which also skipped the `len(readPk) > 0` public-key comparison in `readCryptoParams`.

Startup still failed closed (the BLS suite rejects an empty scalar, verified directly), so this was diagnostic rather than a security exposure — but the operator got `err mclBnFr_deserialize` naming neither the file nor the cause. Now reports `pem file is invalid while reading <path>`.

Also replaced the `strings.Contains` match on error text with `errors.Is(err, os.ErrNotExist)`.

### 3. Key generation is logged, not blocked

The reporter suggested making a missing pem fatal. **Deliberately not doing that** — auto-generation is what lets an observer start without operator-provided key material, and removing it would break fresh observers, container images and CI.

What was wrong is that it was effectively silent, and the node cannot tell at that point whether it was meant to carry a registered validator identity. Now logs at `WARN` with the generated public key and instructions to restore the key file if the node is a registered validator.

### 4. Credential field documented; constant-time compare

`api.yaml` said `password: hashed password`, which reads as an instruction to type a password. Now states it holds a digest, with a command to generate one and a note on permissions and TLS.

`authHandler.go` now uses `subtle.ConstantTimeCompare`.

### 5. Username enumeration closed

Review flagged that hardening the compare left the wider gap untouched: an unknown username returned a distinct `username does not exist` body and bailed *before* the hash ran, so usernames were enumerable one request each with no timing analysis needed. Both failures now answer `invalid credentials`, and the hash always runs against a stand-in digest of the same length so the compare cannot short-circuit.

### 6. `/debug` warning for existing installs

Setting `secured: true` only changes the default for fresh installs — upgrading the binary does not rewrite an operator's `api.yaml`, and an absent flag reads as false. The node now warns at startup when `/debug` is open but not secured, matching the existing warning for the secured-but-not-open `/subscribe` case.

## Tests

Key-file invariants live in `factory/cryptoSigningParams_test.go`:

- **An unloadable pem is never overwritten** with a generated key.
- **An absent pem still creates one**, so the observer path is locked in against a future "fix".
- **Not-found is matched by error type, not by message text.** Review caught that `isSkPemFileNotFound` used a substring match on `"no such file or directory"`, which is satisfied by any PEM error whose *path* contains that text — sending a corrupt key file down the generate-and-replace branch. Reproduced: the file was destroyed. Now `errors.Is(err, os.ErrNotExist)` only, with a regression test.

Route and auth coverage:

- **The shipped `api.yaml` is pinned.** `TestShippedAPIConfig_SecuredRoutes` loads the real file through `LoadAPIConfig` and asserts `/debug` and `/log` are secured. Nothing previously read the shipped yaml — every other case built a synthetic config — so losing the flag to a reformat or a merge resolution kept CI green. Verified it fails on revert.
- **The gate itself is pinned.** `TestQueryDebug_SecuredRejectsUnauthenticated` builds the node router *with* an auth handler and asserts 401 on an unauthenticated `POST /node/debug`.
- **Enumeration parity.** `TestUnknownUserIsIndistinguishableFromWrongPassword` asserts the two failure responses are byte-identical, not merely both 401.

> **Correction:** an earlier revision of this description claimed a validator's key material "was never at risk" under the old code. That was wrong — the substring hazard above pre-exists on `develop` and does destroy the key file. It is fixed here.

## Operator impact

Both need a release-notes line:

- **`/debug` stays open on existing nodes** until an operator adds `secured: true` to `/debug` in their `api.yaml`. The binary upgrade does not do it for them; the new startup warning is the only signal.
- **Auth error responses changed.** `username does not exist` and `invalid password` are both now `invalid credentials`. Anything parsing the old strings needs updating — that consolidation is the enumeration fix.

## Not in this PR

- **Validator identity pinning.** Cannot be unconditional — observers are absent from `nodesSetup.json`, so refusing to start on a mismatch would refuse every observer. Needs node-type detection.
- **`Blake2b.Compute` empty-input data race.** Unsynchronized check-then-write on `emptyHash`, plus the shared slice returned to callers. Raised in review; filed as #144 rather than grown into this diff. Latent — shipped config is `sha256`.
- **Salted/work-factored KDF for credentials.** Declined. Basic Auth hashes on every request, so a work factor is reachable by an unauthenticated caller and converts an offline-crack concern into remote CPU exhaustion. Credential entropy is what defeats cracking here.

## Summary

- **Networking:** Secures `/node/debug` with authentication. Unauthenticated requests now return `401`.
- **Security:** Uses constant-time credential comparison and prevents username enumeration.
- **Key management:** Generates keys only when the PEM file is absent. Invalid, unreadable, or corrupt files now return descriptive errors and remain unchanged.
- **Error handling:** Uses `errors.Is(err, os.ErrNotExist)` for missing-file detection.
- **Operations:** Logs the generated public key and recovery instructions when automatic key generation occurs. Warns when `/node/debug` remains unauthenticated.
- **Blockchain impact:** No changes affect consensus, transaction processing, state management, or KVM execution. The changes improve node security and startup safety without changing blockchain data integrity or node stability behavior.
- **Testing and documentation:** Adds regression tests for route authentication, credential handling, and key-file preservation. Updates credential digest and generation guidance.

### DIFF
--- a/config/node/api.yaml
+++ b/config/node/api.yaml
@@ -85,8 +85,10 @@ apiPackages:
         open: true
       - name: /p2pstatus
         open: true
+      # /debug returns cached interceptor and resolver state; keep it behind auth.
       - name: /debug
         open: true
+        secured: true
       - name: /peerinfo
         open: true
       - name: /statistics
@@ -114,10 +116,19 @@ apiPackages:
       - name: /query
         open: true
 
+# `password` holds a DIGEST, not the password itself: the hex-encoded hash of the
+# password, using the algorithm named under `hasher` below. A plaintext value here
+# will never authenticate.
+#
+# Generate one with:
+#   printf '%s' 'your-password' | shasum -a 256 | cut -d' ' -f1
+#
+# Anyone who can read this file can attempt an offline crack against these digests,
+# so keep it readable only by the node user, and terminate TLS in front of the API.
 credentials:
   - username: example
-    password: hashed password
+    password: replace-me-with-a-sha256-digest
   - username: example2
-    password: hashed password
+    password: replace-me-with-a-sha256-digest
 hasher:
   type: sha256
--- a/config/node/api.yaml
+++ b/config/node/api.yaml
@@ -121,7 +121,8 @@ apiPackages:
 # will never authenticate.
 #
 # Generate one with:
-#   printf '%s' 'your-password' | shasum -a 256 | cut -d' ' -f1
+#   printf '%s' 'your-password' | sha256sum | cut -d' ' -f1     # GNU coreutils
+#   printf '%s' 'your-password' | shasum -a 256 | cut -d' ' -f1 # macOS / perl
 #
 # Anyone who can read this file can attempt an offline crack against these digests,
 # so keep it readable only by the node user, and terminate TLS in front of the API.
--- a/factory/cryptoSigningParams.go
+++ b/factory/cryptoSigningParams.go
@@ -3,6 +3,7 @@ package factory
 import (
 	"bytes"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"os"
 	"strings"
@@ -109,14 +110,27 @@ func (cspf *cryptoSigningParamsLoader) getSkPk() ([]byte, []byte, error) {
 	skIndex := cspf.skIndex
 	encodedSk, pkString, err := tools.LoadSkPkFromPemFile(cspf.skPemFileName, skIndex, os.Getenv("KEY_PASSWORD"))
 	if err != nil {
-		if strings.Contains(err.Error(), ErrFileNotFound.Error()) {
-			keyGen := signing.NewKeyGenerator(cspf.suite)
-			encodedSk, pkString, err = tools.CreateWallet(cspf.skPemFileName, os.Getenv("KEY_PASSWORD"), keyGen, cspf.pubkeyConverter)
-			if err != nil {
-				return nil, nil, err
-			}
+		// Only a missing file is recoverable by generating a key. Anything else
+		// (corrupt pem, wrong KEY_PASSWORD, permissions) must surface here rather
+		// than fall through and fail later against an empty key.
+		if !isSkPemFileNotFound(err) {
+			return nil, nil, fmt.Errorf("%w while loading %s", err, cspf.skPemFileName)
 		}
 
+		keyGen := signing.NewKeyGenerator(cspf.suite)
+		encodedSk, pkString, err = tools.CreateWallet(cspf.skPemFileName, os.Getenv("KEY_PASSWORD"), keyGen, cspf.pubkeyConverter)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		// Generating a key here is intentional: it lets an observer start without
+		// operator-provided key material. It is only a problem when the node was
+		// meant to run under an already-registered validator identity, which we
+		// cannot distinguish at this point, so say so rather than staying silent.
+		log.Warn("no key file found - generated a new node identity",
+			"file", cspf.skPemFileName,
+			"public key", pkString,
+			"note", "expected for a new observer; if this node should run as a registered validator, stop it and restore its key file")
 	}
 
 	skBytes, err := hex.DecodeString(string(encodedSk))
@@ -130,4 +144,11 @@ func (cspf *cryptoSigningParamsLoader) getSkPk() ([]byte, []byte, error) {
 	}
 
 	return skBytes, pkBytes, nil
+}
+
+// isSkPemFileNotFound reports whether the key file is simply absent. The string
+// check is kept as a fallback for loaders that flatten the wrapped os error.
+func isSkPemFileNotFound(err error) bool {
+	return errors.Is(err, os.ErrNotExist) ||
+		strings.Contains(err.Error(), ErrFileNotFound.Error())
 }
--- a/factory/cryptoSigningParams.go
+++ b/factory/cryptoSigningParams.go
@@ -114,7 +114,7 @@ func (cspf *cryptoSigningParamsLoader) getSkPk() ([]byte, []byte, error) {
 		// (corrupt pem, wrong KEY_PASSWORD, permissions) must surface here rather
 		// than fall through and fail later against an empty key.
 		if !isSkPemFileNotFound(err) {
-			return nil, nil, fmt.Errorf("%w while loading %s", err, cspf.skPemFileName)
+			return nil, nil, fmt.Errorf("loading validator key: %w", err)
 		}
 
 		keyGen := signing.NewKeyGenerator(cspf.suite)
--- a/factory/cryptoSigningParams.go
+++ b/factory/cryptoSigningParams.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"strings"
 
 	"github.com/klever-io/klever-go/common"
 	"github.com/klever-io/klever-go/core"
@@ -146,9 +145,10 @@ func (cspf *cryptoSigningParamsLoader) getSkPk() ([]byte, []byte, error) {
 	return skBytes, pkBytes, nil
 }
 
-// isSkPemFileNotFound reports whether the key file is simply absent. The string
-// check is kept as a fallback for loaders that flatten the wrapped os error.
+// isSkPemFileNotFound reports whether the key file is simply absent. Matched by
+// type only: a substring match on the not-found text is satisfied by any error
+// carrying a path that happens to contain it, which would send a corrupt key
+// file down the generate-and-replace branch.
 func isSkPemFileNotFound(err error) bool {
-	return errors.Is(err, os.ErrNotExist) ||
-		strings.Contains(err.Error(), ErrFileNotFound.Error())
+	return errors.Is(err, os.ErrNotExist)
 }
--- a/factory/cryptoSigningParamsKeyFile_test.go
+++ b/factory/cryptoSigningParamsKeyFile_test.go
@@ -1,0 +1,56 @@
+package factory
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/klever-io/klever-go/common/mock"
+	"github.com/stretchr/testify/require"
+)
+
+const corruptPem = "this is not a pem file at all\n"
+
+func newLoaderFor(t *testing.T, pemPath string) *cryptoSigningParamsLoader {
+	t.Helper()
+	cspf, err := NewCryptoSigningParamsLoader(
+		&mock.PubkeyConverterStub{},
+		0,
+		pemPath,
+		&mock.SuiteStub{CreateKeyPairStub: createKeyPair},
+		false,
+	)
+	require.NoError(t, err)
+	return cspf
+}
+
+// SAFETY: an existing-but-unloadable pem (corrupt file, wrong KEY_PASSWORD,
+// bad permissions) must NEVER be replaced by a freshly generated key. Doing so
+// would destroy a validator's key material irrecoverably.
+func TestKeyFileSafety_CorruptPemIsNeverOverwritten(t *testing.T) {
+	dir := t.TempDir()
+	pem := filepath.Join(dir, "validatorKey.pem")
+	require.NoError(t, os.WriteFile(pem, []byte(corruptPem), 0o600))
+
+	_, _, err := newLoaderFor(t, pem).getSkPk()
+	t.Logf("corrupt pem -> err=%v", err)
+
+	after, rerr := os.ReadFile(pem)
+	require.NoError(t, rerr)
+	require.Equal(t, corruptPem, string(after),
+		"CATASTROPHIC: an unloadable pem was overwritten with a generated key")
+}
+
+// The intentional observer path: a genuinely absent pem DOES create a key file.
+func TestKeyFileSafety_MissingPemDoesCreateKey(t *testing.T) {
+	dir := t.TempDir()
+	pem := filepath.Join(dir, "validatorKey.pem")
+
+	_, _, err := newLoaderFor(t, pem).getSkPk()
+	require.NoError(t, err, "observer auto-generation must keep working")
+
+	info, serr := os.Stat(pem)
+	require.NoError(t, serr, "a key file should have been created")
+	require.Greater(t, info.Size(), int64(0))
+	t.Logf("missing pem -> created %s (%d bytes)", pem, info.Size())
+}
--- a/factory/cryptoSigningParamsKeyFile_test.go
+++ b/factory/cryptoSigningParamsKeyFile_test.go
@@ -34,12 +34,32 @@ func TestKeyFileSafety_CorruptPemIsNeverOverwritten(t *testing.T) {
 
 	_, _, err := newLoaderFor(t, pem).getSkPk()
 	require.Error(t, err, "an unloadable pem must surface an error, not be swallowed")
-	t.Logf("corrupt pem -> err=%v", err)
+	require.ErrorContains(t, err, "loading validator key",
+		"the error must name what failed, not surface as a bare deserialize error")
 
 	after, rerr := os.ReadFile(pem)
 	require.NoError(t, rerr)
 	require.Equal(t, corruptPem, string(after),
 		"CATASTROPHIC: an unloadable pem was overwritten with a generated key")
+}
+
+// Regression: the not-found check must match by error TYPE, not by searching the
+// message for "no such file or directory". That text appears in every PEM error
+// whose path happens to contain it, which sent a corrupt key file down the
+// generate-and-replace branch and destroyed it.
+func TestKeyFileSafety_NotFoundIsMatchedByTypeNotMessage(t *testing.T) {
+	dir := filepath.Join(t.TempDir(), "no such file or directory")
+	require.NoError(t, os.MkdirAll(dir, 0o755))
+	pem := filepath.Join(dir, "validatorKey.pem")
+	require.NoError(t, os.WriteFile(pem, []byte(corruptPem), 0o600))
+
+	_, _, err := newLoaderFor(t, pem).getSkPk()
+	require.Error(t, err, "a corrupt pem must fail even when its path contains the not-found text")
+
+	after, rerr := os.ReadFile(pem)
+	require.NoError(t, rerr)
+	require.Equal(t, corruptPem, string(after),
+		"CATASTROPHIC: a corrupt pem on a path containing the not-found text was replaced")
 }
 
 // The intentional observer path: a genuinely absent pem DOES create a key file.
--- a/factory/cryptoSigningParamsKeyFile_test.go
+++ b/factory/cryptoSigningParamsKeyFile_test.go
@@ -33,6 +33,7 @@ func TestKeyFileSafety_CorruptPemIsNeverOverwritten(t *testing.T) {
 	require.NoError(t, os.WriteFile(pem, []byte(corruptPem), 0o600))
 
 	_, _, err := newLoaderFor(t, pem).getSkPk()
+	require.Error(t, err, "an unloadable pem must surface an error, not be swallowed")
 	t.Logf("corrupt pem -> err=%v", err)
 
 	after, rerr := os.ReadFile(pem)
--- a/network/api/middleware/authHandler.go
+++ b/network/api/middleware/authHandler.go
@@ -1,6 +1,7 @@
 package middleware
 
 import (
+	"crypto/subtle"
 	"encoding/hex"
 	"net/http"
 
@@ -60,7 +61,8 @@ func NewAuthenticationFunc(credentialsConfig config.APIRoutesConfig) gin.Handler
 			return
 		}
 
-		if userPassword != hex.EncodeToString(hasher.Compute(pass)) {
+		expected := hex.EncodeToString(hasher.Compute(pass))
+		if subtle.ConstantTimeCompare([]byte(userPassword), []byte(expected)) != 1 {
 			c.AbortWithStatusJSON(http.StatusUnauthorized, shared.GenericAPIResponse{
 				Data:  nil,
 				Error: "invalid password",


factory/cryptoSigningParamsKeyFile_test.go
```diff
@@ -0,0 +1,56 @@
...
+	pem := filepath.Join(dir, "validatorKey.pem")
+	require.NoError(t, os.WriteFile(pem, []byte(corruptPem), 0o600))
+
+	_, _, err := newLoaderFor(t, pem).getSkPk()
+	t.Logf("corrupt pem -> err=%v", err)
+
+	after, rerr := os.ReadFile(pem)
+	require.NoError(t, rerr)
+	require.Equal(t, corruptPem, string(after),
+		"CATASTROPHIC: an unloadable pem was overwritten with a generated key")
```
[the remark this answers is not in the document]
fbsobreira: Fixed — added `require.Error(t, err, "an unloadable pem must surface an error, not be swallowed")` before the preservation assertions.

You are right that this was a real gap. Surfacing the load error is the central change in this PR and nothing covered it — the throwaway probe I used to verify the behaviour during development was not kept. Confirmed the new assertion actually bites: against the pre-fix code `getSkPk()` returned `err=nil` for a corrupt pem, so a regression restoring the swallow now fails this test rather than passing it.

factory/cryptoSigningParamsKeyFile_test.go
```diff
@@ -0,0 +1,56 @@
...
+// SAFETY: an existing-but-unloadable pem (corrupt file, wrong KEY_PASSWORD,
+// bad permissions) must NEVER be replaced by a freshly generated key. Doing so
+// would destroy a validator's key material irrecoverably.
+func TestKeyFileSafety_CorruptPemIsNeverOverwritten(t *testing.T) {
+	dir := t.TempDir()
+	pem := filepath.Join(dir, "validatorKey.pem")
+	require.NoError(t, os.WriteFile(pem, []byte(corruptPem), 0o600))
+
+	_, _, err := newLoaderFor(t, pem).getSkPk()
+	t.Logf("corrupt pem -> err=%v", err)
```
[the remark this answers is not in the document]
fbsobreira: Fixed — same change as the one CodeRabbit flagged on this file: `require.Error` now runs before the file-preservation assertions.

Agreed on the reasoning. The PR's stated goal is surfacing key-file load errors, and the test only pinned the file-safety half of that. Verified the assertion is meaningful: against the pre-fix code `getSkPk()` returned `err=nil` for an unloadable pem, so this now catches a regression instead of silently passing.

factory/cryptoSigningParams.go
```diff
@@ -109,14 +110,27 @@ func (cspf *cryptoSigningParamsLoader) getSkPk() ([]byte, []byte, error) {
...
-			encodedSk, pkString, err = tools.CreateWallet(cspf.skPemFileName, os.Getenv("KEY_PASSWORD"), keyGen, cspf.pubkeyConverter)
-			if err != nil {
-				return nil, nil, err
-			}
+		// Only a missing file is recoverable by generating a key. Anything else
+		// (corrupt pem, wrong KEY_PASSWORD, permissions) must surface here rather
+		// than fall through and fail later against an empty key.
+		if !isSkPemFileNotFound(err) {
+			return nil, nil, fmt.Errorf("%w while loading %s", err, cspf.skPemFileName)
 		}
```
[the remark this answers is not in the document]
fbsobreira: Fixed — and it was worse than the ordering. `LoadSkPkFromPemFile` already embeds the path in its own error, so my added `%s` duplicated it.

Before:
```
pem file is invalid while reading <path> file, error decoding while loading <path>
```
After:
```
loading validator key: pem file is invalid while reading <path> file, error decoding
```

Now `fmt.Errorf("loading validator key: %w", err)` — context first, wrapped error last, path only once.

config/node/api.yaml
```diff
@@ -114,10 +116,19 @@ apiPackages:
       - name: /query
         open: true
 
+# `password` holds a DIGEST, not the password itself: the hex-encoded hash of the
+# password, using the algorithm named under `hasher` below. A plaintext value here
+# will never authenticate.
+#
+# Generate one with:
+#   printf '%s' 'your-password' | shasum -a 256 | cut -d' ' -f1
+#
```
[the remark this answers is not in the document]
fbsobreira: Fixed — documented both, with `sha256sum` (GNU coreutils) listed first since Linux is the common node platform, and `shasum -a 256` noted for macOS.

network/api/middleware/authHandler.go
```diff
@@ -60,7 +61,8 @@ func NewAuthenticationFunc(credentialsConfig config.APIRoutesConfig) gin.Handler
 			return
 		}
 
-		if userPassword != hex.EncodeToString(hasher.Compute(pass)) {
+		expected := hex.EncodeToString(hasher.Compute(pass))
+		if subtle.ConstantTimeCompare([]byte(userPassword), []byte(expected)) != 1 {
```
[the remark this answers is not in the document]
fbsobreira: Thanks — the general point about `subtle.ConstantTimeCompare` is correct, but I do not think it changes anything in this specific call, so I am leaving it as is.

`expected` is `hex.EncodeToString(hasher.Compute(pass))`, which is invariantly 64 characters for sha256. The caller controls `pass`, but `pass` never changes the *length* of `expected`. So the early-return-on-length-mismatch branch depends only on the length of the digest in `api.yaml` — a static, operator-set property — and not on anything an attacker can vary per request. An explicit length check would behave identically and would read as security-relevant while doing nothing.

Worth flagging the larger point though: the username lookup immediately above this returns a distinct `username does not exist` error and skips the hash computation entirely, so username existence is already observable both by response body and by timing. That is the enumeration leak actually worth closing, and it is a deliberate change to API responses (with test assertions on the exact JSON) rather than a drive-by — so I would rather do it as its own change than fold it in here.

Happy to revisit if you see a path where `expected` can vary in length.

factory/cryptoSigningParamsKeyFile_test.go
```diff
@@ -0,0 +1,57 @@
...
+// bad permissions) must NEVER be replaced by a freshly generated key. Doing so
+// would destroy a validator's key material irrecoverably.
+func TestKeyFileSafety_CorruptPemIsNeverOverwritten(t *testing.T) {
+	dir := t.TempDir()
+	pem := filepath.Join(dir, "validatorKey.pem")
+	require.NoError(t, os.WriteFile(pem, []byte(corruptPem), 0o600))
+
+	_, _, err := newLoaderFor(t, pem).getSkPk()
+	require.Error(t, err, "an unloadable pem must surface an error, not be swallowed")
+	t.Logf("corrupt pem -> err=%v", err)
```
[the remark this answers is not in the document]
fbsobreira: Fixed — added `require.ErrorContains(t, err, "loading validator key")` alongside the non-nil check.

Agreed on the reasoning: a bare `require.Error` would pass on any error, including the pre-fix `mclBnFr_deserialize` that named neither the file nor the cause, which is precisely the thing this PR set out to replace.

fbsobreira:
@coderabbitai — the outside-diff finding on `isSkPemFileNotFound` is correct, and it is the most serious thing found on this PR. Fixed.

I reproduced it before changing anything. With a key file at a path containing `no such file or directory`:

```
BEFORE:  err = <nil>      preserved = false   <- corrupt pem REPLACED by a generated key
AFTER:   err = loading validator key: pem file is invalid ...
                          preserved = true
```

The substring branch matched on the *path*, not on the error type, so a corrupt or unreadable key file took the generate-and-replace branch and `CreateWallet` destroyed it. That is unrecoverable loss of a validator's key material.

Three notes for the record:

1. **It pre-exists on `develop`.** I verified the same reproduction against `origin/develop`, so this is not a regression introduced here — the substring check was in the original code and I preserved it as a fallback.
2. **It is still in scope for this PR.** This PR modifies `getSkPk` and adds a test asserting the key file is never overwritten. A hazard that the PR's own test claims to exclude is not something to defer.
3. **The fallback was never load-bearing.** `LoadSkPkFromPemFile` returns the `*os.PathError` from `os.Open` unwrapped, so `errors.Is(err, os.ErrNotExist)` matches on its own — `TestCryptoSigningParamsLoader_GetSkPk_PathNotFound_CreateNew` passes without the substring. It bought nothing and cost this.

`isSkPemFileNotFound` now matches by type only, with a comment recording why, so the substring is not reinstated later.

Also added the regression test you asked for: `TestKeyFileSafety_NotFoundIsMatchedByTypeNotMessage` places a corrupt pem on a path containing the legacy message and asserts both that the error surfaces and that the file is preserved. Confirmed it fails against the previous commit and passes with the fix.

I have corrected the PR description as well — it previously claimed key material "was never at risk" under the old code, which this finding disproves.

config/node/api.yaml
```diff
@@ -85,8 +85,10 @@ apiPackages:
         open: true
       - name: /p2pstatus
         open: true
+      # /debug returns cached interceptor and resolver state; keep it behind auth.
       - name: /debug
         open: true
+        secured: true
```
nickgs1337, line 91: Nothing pins this flag anywhere. No test loads the shipped yaml (config/api_test.go only builds synthetic route configs), and routes_test.go:554 still declares /debug as `open: true` with no `Secured`, so if this line gets dropped in a reformat or a merge conflict resolution, CI stays green. Can we add a test that parses config/node/api.yaml and asserts `IsRouteSecured("node", "/debug")`? factory/process_test.go:107 already reads a shipped config that way. Worth flipping that routes_test case to `Secured: true` too and covering the 401 for an unauthenticated request, otherwise /log is the only secured route ever exercised end to end.

config/node/api.yaml
```diff
@@ -85,8 +85,10 @@ apiPackages:
         open: true
       - name: /p2pstatus
         open: true
+      # /debug returns cached interceptor and resolver state; keep it behind auth.
       - name: /debug
```
nickgs1337, line 89: Strictly this only changes the default for fresh installs. `routeHasFlag` returns false when the flag is absent, and upgrading the binary doesn't rewrite an operator's existing api.yaml, so every node already out there keeps /debug open and unauthenticated with no signal that anything changed. Could we log a warn at startup when /debug is open but not secured? api.go:148 already does that for the secured-but-not-open /subscribe case, and open-but-not-secured is the direction that actually leaves state exposed. The required config edit is worth a line in the release notes too.

network/api/middleware/authHandler.go
```diff
@@ -60,7 +61,8 @@ func NewAuthenticationFunc(credentialsConfig config.APIRoutesConfig) gin.Handler
 			return
 		}
 
-		if userPassword != hex.EncodeToString(hasher.Compute(pass)) {
+		expected := hex.EncodeToString(hasher.Compute(pass))
```
nickgs1337, line 66: Read your reply to Copilot on this, but I'd push back on deferring it, since this PR is exactly what makes this middleware the gate on /node/debug. The branch just above returns `username does not exist` instead of `invalid password` and bails before the hash ever runs, so usernames are enumerable straight off the response body with one request each, and that timing gap is much wider than the compare you just hardened. Can we collapse both failures into a single `invalid credentials` response and always compute the hash, comparing against a fixed dummy digest of the same length so the length short-circuit doesn't split the paths again? `TestIncorrectUser` asserts the current message verbatim, so it has to move with it.

nickgs1337, line 66: Separate thing on this line, and not yours: `hasher` is a single instance captured by the closure and shared across every request, and `Blake2b.Compute` does an unsynchronized check-then-write on `emptyHash` for zero-length input, which any unauthenticated caller can reach by sending Basic Auth with an empty password. Shipped config is sha256 so it doesn't bite today. The fix belongs in the blake2b package (a `sync.Once`, like `Sha256` already has), so I'd file it separately rather than grow this diff. Flagging it so it doesn't get lost.

factory/cryptoSigningParamsKeyFile_test.go
```diff
@@ -0,0 +1,77 @@
+package factory
```
nickgs1337: Any reason for a new file here? cryptoSigningParams_test.go already covers `getSkPk`, and this one reaches into it for `createKeyPair`, so the split buys nothing and leaves a cross-file dependency that reads as accidental. Moving the cases over there, and calling the exported `GetSkPk()` from export_test.go like the neighbours do instead of the unexported one, would match the package. Not asking for less coverage, the corrupt-pem cases are the best part of this PR.

factory/cryptoSigningParamsKeyFile_test.go
```diff
@@ -0,0 +1,77 @@
...
+		false,
+	)
+	require.NoError(t, err)
+	return cspf
+}
+
+// SAFETY: an existing-but-unloadable pem (corrupt file, wrong KEY_PASSWORD,
+// bad permissions) must NEVER be replaced by a freshly generated key. Doing so
+// would destroy a validator's key material irrecoverably.
+func TestKeyFileSafety_CorruptPemIsNeverOverwritten(t *testing.T) {
```
nickgs1337: nit: this one and the next are near-identical bodies, only the pem's parent directory differs. Could be a single table with two cases, something like "plain dir" and "dir named 'no such file or directory'". If you do that, keep the `ErrorContains` as a per-case field since only this one asserts it, and keep the regression intent in the case name so it isn't lost.

factory/cryptoSigningParamsKeyFile_test.go
```diff
@@ -0,0 +1,77 @@
...
+	require.Error(t, err, "a corrupt pem must fail even when its path contains the not-found text")
+
+	after, rerr := os.ReadFile(pem)
+	require.NoError(t, rerr)
+	require.Equal(t, corruptPem, string(after),
+		"CATASTROPHIC: a corrupt pem on a path containing the not-found text was replaced")
+}
+
+// The intentional observer path: a genuinely absent pem DOES create a key file.
+func TestKeyFileSafety_MissingPemDoesCreateKey(t *testing.T) {
```
nickgs1337: nit: `TestCryptoSigningParamsLoader_GetSkPk_PathNotFound_CreateNew` in cryptoSigningParams_test.go already pins the absent-pem path, and what's genuinely new here is that a wallet file gets written, which is `CreateWallet`'s contract rather than `getSkPk`'s. Thin value, your call whether it stays. Either way the `t.Logf` at the bottom should go, passing tests should be silent.

factory/cryptoSigningParams.go
```diff
@@ -109,14 +109,27 @@ func (cspf *cryptoSigningParamsLoader) getSkPk() ([]byte, []byte, error) {
...
+			return nil, nil, fmt.Errorf("loading validator key: %w", err)
 		}
 
+		keyGen := signing.NewKeyGenerator(cspf.suite)
+		encodedSk, pkString, err = tools.CreateWallet(cspf.skPemFileName, os.Getenv("KEY_PASSWORD"), keyGen, cspf.pubkeyConverter)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		// Generating a key here is intentional: it lets an observer start without
```
nickgs1337: nit: this comment says roughly what the `note` field three lines down says, and the log is the version that actually reaches operators. Could trim it to the one bit the log doesn't carry, that we can't tell the two cases apart at this point.

factory/cryptoSigningParams.go
```diff
@@ -131,3 +144,11 @@ func (cspf *cryptoSigningParamsLoader) getSkPk() ([]byte, []byte, error) {
 
 	return skBytes, pkBytes, nil
 }
+
+// isSkPemFileNotFound reports whether the key file is simply absent. Matched by
+// type only: a substring match on the not-found text is satisfied by any error
+// carrying a path that happens to contain it, which would send a corrupt key
+// file down the generate-and-replace branch.
+func isSkPemFileNotFound(err error) bool {
```
nickgs1337, line 150: nit: one `errors.Is` with a single caller, so this could just be inlined. I don't feel strongly, `!isSkPemFileNotFound(err)` does read well at the call site. If you do inline it, keep the why-not-substring rationale near the call, that's the whole regression this PR exists to prevent.

nickgs1337:
`ErrFileNotFound` (factory/errors.go:48) is dead now. The `strings.Contains(err.Error(), ErrFileNotFound.Error())` you removed was its only reference, grep shows nothing else uses it. Its value is the bare "no such file or directory" text and nothing ever returns it, so leaving it parked in the package is an open invitation to redo the substring match this PR just fixed. It's exported, so maybe a quick check for out-of-tree users before it goes.

nickgs1337:
docs/node_swagger.yaml:2013, and the matching annotation at docs/node_docs.go:790, still document /node/debug as open with only 200/400 responses. With `secured: true` it can now return 401, plus 500 when the server has no credentials configured, so anyone generating a client off those docs won't send Basic Auth or handle the new codes. /log has the same gap but that one's pre-existing, just /debug here since this PR is what changes its contract.

config/node/api.yaml
fbsobreira: Fixed — you're right that nothing pinned it.

Added `TestShippedAPIConfig_SecuredRoutes` in `config/api_test.go`, which loads the real `config/node/api.yaml` through `LoadAPIConfig` and asserts `IsRouteSecured("node", "/debug")` (and `/log` alongside it). Verified it fails on revert: dropping the `secured: true` line makes it fail with `/debug must require Basic Auth`.

Also flipped `node/routes_test.go:554` to `{Name: "/debug", Open: true, Secured: true}` and added `TestQueryDebug_SecuredRejectsUnauthenticated`, which builds the wrapper *with* an auth handler and asserts 401 on an unauthenticated POST.

Worth recording for whoever reads this next: the existing `/debug` cases keep passing after the flip because `RegisterHandler` only applies the auth wrapper when one is passed, and `startNodeServerWithFacade` passes none. That is also exactly why the flag could go missing unnoticed.

config/node/api.yaml
fbsobreira: Fixed — agreed, the yaml edit only helps fresh installs, and a silent no-op on every existing node is the worse half of this.

Added a startup warn in `RegisterRoutes`, directly below the `/subscribe` one and in the same shape:

```go
if routesConfig.IsRouteEnabled(nodePackage, debugRoute) && !routesConfig.IsRouteSecured(nodePackage, debugRoute) {
	log.Warn("node debug route is open but not secured; /node/debug answers unauthenticated with cached interceptor and resolver state. Add secured:true to /debug in api.yaml.")
}
```

Kept it scoped to `/debug` rather than a general open-but-not-secured sweep — that would fire on `/status`, `/metrics` and the rest of the open routes on every start and get tuned out within a week, which costs the warning that matters.

Adding the required config edit to the PR description as a rollout note.

network/api/middleware/authHandler.go
fbsobreira: Fixed, and you're right that deferring it was the wrong call — this PR is what makes this middleware the gate on `/node/debug`, so it is in scope by exactly that reasoning.

Both failures now answer `invalid credentials`, and the hash always runs:

```go
unknownUserDigest := hex.EncodeToString(make([]byte, hasher.Size()))
...
userPassword, userExists := accounts[user]
if !userExists {
	userPassword = unknownUserDigest
}

expected := hex.EncodeToString(hasher.Compute(pass))
if subtle.ConstantTimeCompare([]byte(userPassword), []byte(expected)) != 1 || !userExists {
```

The stand-in digest is the same length as a real one so the compare cannot split on length as you noted, and the `|| !userExists` is belt-and-braces so an unknown user fails even if some password hashed to the all-zero digest.

`TestIncorrectUser` and `TestIncorrectPassword` moved to the new message. Added `TestUnknownUserIsIndistinguishableFromWrongPassword`, which asserts the two responses are byte-identical rather than merely both 401 — a later change that reintroduces a distinct body fails there rather than passing two separate 401 assertions.

This is an operator-visible response change, so it is going in the PR description too.

fbsobreira: Good catch, and agreed it belongs outside this diff — filed #144.

Confirmed both halves while writing it up. `Compute` does the unsynchronized check-then-write you describe, and it also hands back the instance's own `emptyHash` slice rather than a copy, so a caller that mutates the returned digest corrupts every later empty-input result. `EmptyHash()` has the same check-then-write and needs the same fix. `Sha256` already has the shape to copy from: package-level `sync.Once` plus a copy on return.

Recorded in the issue that the shipped config is `sha256` so nothing is exposed today, and that the trie hashes concurrently too, so the fix is worth making independently of the API path.

factory/cryptoSigningParamsKeyFile_test.go
fbsobreira: Fixed — no reason beyond how it got written. Moved all three cases into `cryptoSigningParams_test.go` next to `createKeyPair`, deleted `cryptoSigningParamsKeyFile_test.go`, and they call the exported `GetSkPk()` from `export_test.go` like the neighbours instead of reaching for the unexported one. Coverage is unchanged; the cross-file dependency is gone because there is no longer a second file.

factory/cryptoSigningParamsKeyFile_test.go
fbsobreira: I'd like to keep these as two tests, though it's a fair reading either way.

They look near-identical but they pin different things: the first asserts the error text (`loading validator key`), the second asserts that a path *containing* `no such file or directory` still fails — which is the actual regression this PR exists to prevent. In a table those become one case with an optional `ErrorContains` field set and one with it empty, and the reader has to reconstruct from that which case is the regression. Two names say it outright.

Happy to convert if you'd still prefer the table — say the word and I'll do it.

factory/cryptoSigningParamsKeyFile_test.go
fbsobreira: Half taken.

`t.Logf` is gone — agreed, passing tests should be silent.

Kept the test but retargeted it on your point. `GetSkPk_PathNotFound_CreateNew` does pin the absent-pem path, but it only asserts `sk` and `pk` are non-nil; it never checks anything reached disk. That side effect is `CreateWallet`'s contract as you say, and it is the part that could disappear without a test noticing. Renamed to `TestKeyFileSafety_MissingPemWritesKeyFile` and its comment now names the existing test, so the division between the two is explicit rather than accidental.

factory/cryptoSigningParams.go
fbsobreira: Fixed — trimmed to the one thing the log line doesn't carry:

```go
// We cannot tell here whether this node was meant to carry a registered
// validator identity, so the warning has to cover both cases.
```

factory/cryptoSigningParams.go
fbsobreira: Leaving the helper, on the strength of your own caveat.

Inlining `errors.Is(err, os.ErrNotExist)` reads fine, but it leaves the why-not-substring rationale nowhere natural to live except a comment floating above an `if` — and that rationale is the whole regression this PR exists to prevent. A named predicate carries it in a doc comment attached to the thing it explains, which is where it survives a future edit.

Your other half was the argument too: `!isSkPemFileNotFound(err)` does read well at the call site. Happy to inline it if you'd rather.

PR #146, opened 2026-09-17
[KLC-2673][KLR-17] Bound unknown heartbeat identity state and persistence

RomuloSiebra:
## Summary
Follow-up to the KLR-17 fix (KLC-2524). The heartbeat monitor now enforces the unknown-identity cap on live in-memory state at eviction time, restricts every persistence path to admitted identities, and bounds how often received heartbeats trigger a full recomputation of the heartbeat map.

## Problem
The KLC-2524 change tracked unknown heartbeat identities in a bounded transient cache, but two gaps remained:

- Evicting or expiring an identity only removed it from the transient cache; its `heartbeatMessages` and `doubleSignerPeers` entries stayed until the next `Cleanup()`, so live state could exceed the configured cap between cleanup runs.
- `SaveMultipleHeartbeatMessageInfos` did not check admission, so an unknown identity that transitioned to inactive during recomputation was still written to the heartbeat storer. Those records were never added to the persisted key list, which is what startup pruning iterates, so they could not be removed on restart.

Additionally, every received heartbeat scheduled an independent full-map recomputation, including heartbeats rejected by the caps.

## Solution
- Sweep and eviction now return the identities they drop, and the caller releases their live state under `mutHeartbeatMessages`. The transient mutex is never held while acquiring the map lock, preserving the existing lock order (`mutHeartbeatMessages` → `mutTransientUnknownHeartbeatPubKeys`) with no inversion.
- Before any live entry is deleted or created, the current tracked/admitted status is re-checked under the map lock, so an identity re-tracked concurrently keeps its state and an identity evicted mid-flight does not create orphaned state.
- `SaveMultipleHeartbeatMessageInfos` skips any identity that is not admitted, which gates both recomputation paths.
- `ProcessReceivedMessage` now runs a single goroutine that recomputes only after the heartbeat was actually applied, and recomputations are coalesced through a dirty-flag single runner: a burst of accepted heartbeats during a walk yields one further walk, not one per message. The runner observes `Close()`.
- `Cleanup()` snapshots the tracked set once instead of re-locking per entry, with a live re-check only for deletion candidates.

## Key Changes
- **Updated:** `node/heartbeat/process/monitor.go` — eviction/sweep return dropped identities; `dropLiveHeartbeatStateLocked`; admission gate in `SaveMultipleHeartbeatMessageInfos`; `processValidatedHeartbeat`, `scheduleHeartbeatRecompute`, `runScheduledRecomputes`; re-verification under lock in `addHeartbeatMessageToMap`; `Cleanup` snapshot; `isStopped`.
- **Updated:** `node/heartbeat/process/monitor_test.go` — regression coverage for the cap before cleanup, no persistence across an unknown identity's lifecycle, rejected heartbeats not triggering recomputation, coalescing under a burst, live/tracked consistency under concurrent churn, and recomputes stopping after close.
- **Updated:** `node/heartbeat/process/export_test.go` — test hooks for the above.
- **Removed:** `transientUnknownHeartbeatInfo` accessor (replaced by `isTransientUnknownHeartbeatPubKeyTracked`).

## Testing
  - `go test ./node/heartbeat/... -race -count=3`
  - Each new test verified to fail against the code it guards.
  - Stress run driving `ProcessReceivedMessage` from 64 goroutines over cap with concurrent `Cleanup`, `GetHeartbeats`, refresh and admission changes: no data races, no map faults, no hangs.
  - Local devnet (seednode + validator + 2 observers): on `develop` an observer record was persisted through the inactive-transition path outside the key list and survived restart; on this branch nothing was persisted. Observer visibility, block production, p2p mesh, shutdown and restart unaffected.

## Configuration Changes
```yaml
None
```

## Breaking Changes
None

## Related Issues
Related to KLC-2673
Related to KLC-2524

## Checklist

## Summary

- Bounds transient unknown heartbeat identity state with global and per-origin limits.
- Removes stale heartbeat and double-signer state after eviction or expiration.
- Persists heartbeat data only for admitted identities.
- Rechecks admission under lock during processing and cleanup.
- Recomputes heartbeat state only after accepted heartbeats.
- Coalesces concurrent recomputations and stops scheduled work after shutdown.
- Adds regression tests for caps, persistence, recomputation, concurrency, cleanup, and shutdown.

## Blockchain impact

- **Consensus, transaction processing, and KVM:** No direct changes.
- **State management:** Limits transient identity state and removes stale live entries.
- **Networking:** Changes heartbeat admission, persistence, cleanup, and recomputation.

## Stability and data integrity

The changes reduce memory-growth risk from unknown identities and race risk during concurrent heartbeat processing and cleanup. Rejected heartbeats no longer trigger recomputation.

Test execution results are not available from the supplied evidence.


node/heartbeat/process/monitor.go
```diff
@@ -373,22 +380,76 @@ func (m *Monitor) ProcessReceivedMessage(message p2p.MessageP2P, fromConnectedPe
...
 	pubKeyStr := string(hb.Pubkey)
 	isAdmittedPubKey := m.isAdmittedHeartbeatPubKey(pubKeyStr)
+	droppedPubKeys := make([]string, 0)
 	if !isAdmittedPubKey {
-		if !m.trackTransientUnknownHeartbeatPubKey(pubKeyStr, fromConnectedPeer) {
-			return
+		tracked, dropped := m.trackTransientUnknownHeartbeatPubKey(pubKeyStr, fromConnectedPeer)
+		droppedPubKeys = dropped
+		if !tracked {
+			m.mutHeartbeatMessages.Lock()
```
n4t4l: maybe add a if len(droppedPubKeys > 0) here, with these changes this rejection path always takes the lock GetHeartbeats and the recompute walks also use

n4t4l (dismissed):
lgtm, one optional nit inline 
✋ ✋

node/heartbeat/process/monitor.go
```diff
@@ -373,22 +380,76 @@ func (m *Monitor) ProcessReceivedMessage(message p2p.MessageP2P, fromConnectedPe
...
+}
+
+func (m *Monitor) addHeartbeatMessageToMap(hb *data.Heartbeat, fromConnectedPeer core.PeerID) bool {
 	pubKeyStr := string(hb.Pubkey)
 	isAdmittedPubKey := m.isAdmittedHeartbeatPubKey(pubKeyStr)
+	droppedPubKeys := make([]string, 0)
 	if !isAdmittedPubKey {
-		if !m.trackTransientUnknownHeartbeatPubKey(pubKeyStr, fromConnectedPeer) {
-			return
+		tracked, dropped := m.trackTransientUnknownHeartbeatPubKey(pubKeyStr, fromConnectedPeer)
```
Beroni: **low · logic/race** (exploit-digger, confidence: medium)

A key can be both admitted and tracked as transient-unknown, so it holds a cap slot it shouldn't.

`addHeartbeatMessageToMap` reads `isAdmittedHeartbeatPubKey` without holding a lock, then calls `trackTransientUnknownHeartbeatPubKey`. `markHeartbeatPubKeyAsAdmitted` does its work in two steps under two separate locks: it adds the key to the admitted set, then deletes it from the transient map. If the transient insert happens after that delete, a stale transient entry for an admitted key stays behind.

The new re-check under `mutHeartbeatMessages` (line ~447) keeps the live map correct. Nothing removes the transient entry, though. Later heartbeats for that key take the admitted path and never touch the entry again. It keeps using one global slot and one per-origin slot until the sweep removes it after `maxDurationPeerUnresponsive`.

**Trigger:** a heartbeat for pubkey X arrives at the same moment the refresh loop admits X through `computeInactiveHeartbeatMessages` → `markHeartbeatPubKeyAsAdmitted`. The goroutine sees X as not admitted before the admit, and `trackTransient…` runs after the admit's delete. X then counts against its origin's cap and the global cap. When the global cap is full, this can evict a real unknown identity early.

**Fix:** in `trackTransientUnknownHeartbeatPubKey`, re-check `isAdmittedHeartbeatPubKey` while holding `mutTransientUnknownHeartbeatPubKeys`. `markHeartbeatPubKeyAsAdmitted` already takes the locks in admitted → transient order, so this adds no new lock ordering. Alternatively, when the under-lock re-check in `addHeartbeatMessageToMap` finds the key admitted, delete it from the transient map.

node/heartbeat/process/monitor.go
```diff
@@ -373,22 +380,76 @@ func (m *Monitor) ProcessReceivedMessage(message p2p.MessageP2P, fromConnectedPe
...
+func (m *Monitor) processValidatedHeartbeat(hb *data.Heartbeat, fromConnectedPeer core.PeerID) {
+	if !m.addHeartbeatMessageToMap(hb, fromConnectedPeer) {
+		return
+	}
+
+	m.scheduleHeartbeatRecompute()
+}
+
+func (m *Monitor) scheduleHeartbeatRecompute() {
+	if m.isStopped() {
```
Beroni: **cut · reuse** (lazy-senior-dev + exploit-digger) · ~30 lines saved

Swap the hand-rolled coalescing runner for a 1-slot channel in `runRefreshLoop`.

Right now the runner is built from `recomputeDirty` and `recomputeRunning` atomics, a CAS loop in `runScheduledRecomputes`, and an `isStopped` helper. Its only jobs are to keep `computeAllHeartbeatMessages` from overlapping with itself and to stop on `Close()`.

**Fix:**
1. Add `recomputeCh chan struct{}` with capacity 1.
2. In `scheduleHeartbeatRecompute`, do a non-blocking send: `select { case m.recomputeCh <- struct{}{}: default: }`.
3. Add `case <-m.recomputeCh: m.computeAllHeartbeatMessages()` to the existing `select` in `runRefreshLoop`.

`runRefreshLoop` already coalesces work, observes `stopCh`, and is tracked by `m.wg`. Moving the recompute there also resolves the concern that the runner goroutine isn't tracked by `m.wg` (**`exploit-digger` and `lazy-senior-dev` point at this independently**). It also runs recomputes one at a time alongside `refreshHeartbeatMessageInfo`.

node/heartbeat/process/monitor.go
```diff
@@ -373,22 +380,76 @@ func (m *Monitor) ProcessReceivedMessage(message p2p.MessageP2P, fromConnectedPe
...
+		if !m.recomputeRunning.CompareAndSwap(false, true) {
+			return
+		}
+	}
+}
+
+func (m *Monitor) addHeartbeatMessageToMap(hb *data.Heartbeat, fromConnectedPeer core.PeerID) bool {
 	pubKeyStr := string(hb.Pubkey)
 	isAdmittedPubKey := m.isAdmittedHeartbeatPubKey(pubKeyStr)
+	droppedPubKeys := make([]string, 0)
```
Beroni: **cut · shrink** (lazy-senior-dev) · ~7 lines saved

The `!tracked` early-return branch duplicates the re-check under the lock.

The `!tracked` branch locks, drops, unlocks and returns false. The `!isAdmitted && !isTransientUnknownHeartbeatPubKeyTracked` re-check under `mutHeartbeatMessages` a few lines below already covers that case. Collapsing the two also changes one behaviour: a key that gets admitted concurrently would now be accepted instead of rejected. That is arguably the better outcome.

The same spot can also use `var droppedPubKeys []string` instead of `make([]string, 0)`, because the value is overwritten right away.

Beroni (commented):
## Automated review — Bazel

**Verdict: 💬 COMMENT.** There are no blocking issues. One low-severity logic race and two simplification cuts are left for the author to decide on.

| | |
|---|---|
| Files (3) | `node/heartbeat/process/monitor.go`, `node/heartbeat/process/monitor_test.go`, `node/heartbeat/process/export_test.go` |
| Lenses run | `senior-code-reviewer` (approve, 0 findings), `exploit-digger` (comment), `lazy-senior-dev` (comment) |

**Inline:**
1. **[LOW]** A key can be both admitted and tracked as transient-unknown, so it holds a cap slot it shouldn't (`monitor.go:435`)
2. **C1 · reuse** — swap the hand-rolled coalescing runner for a 1-slot channel in `runRefreshLoop` (`monitor.go:397`, ~30 lines saved)
3. **C2 · shrink** — the `!tracked` early-return branch duplicates the re-check under the lock (`monitor.go:433`, ~7 lines saved)

### Coverage (`exploit-digger`, condensed)

- **Lock ordering:** every path takes `mutHeartbeatMessages` → `mutAdmittedHeartbeatPubKeys` / `mutTransientUnknownHeartbeatPubKeys`. The transient-only paths release the transient lock before taking the map lock. No reverse order exists, so there is no deadlock.
- **TOCTOU between the transient cache and the live map:** inserts and drops both re-check admitted/tracked status under `mutHeartbeatMessages`. Every dropped list reaches `dropLiveHeartbeatStateLocked`. Once processing settles, live unknown entries are a subset of the tracked set and stay bounded by the cap.
- **Dirty-flag runner:** no lost wakeup and no case where two runners run at once. It stops after `Close()` once the walk already in progress finishes.
- **Persistence gate:** `SaveMultipleHeartbeatMessageInfos` skips non-admitted keys, which covers both recompute paths. `addHeartbeatMessageToMap` and `addPeerToFullPeersSlice` persist only admitted keys. No bypass found.
- **Error handling:** every early return releases `mutHeartbeatMessages`. The `newHeartbeatMessageInfo` failure path returns false, so no recompute is scheduled.
- **Resources:** recompute goroutines are down from one per message to one runner. Live unknown state is bounded by the cap.
- **Checked and clean:** injection, authz (admission comes only from `pubKeysList`, the storer and `peerTypeProvider`), message validation (unchanged), secrets, crypto, and dependencies (the only new import is stdlib `sync/atomic`, which needs Go ≥ 1.19).
- **Tests:** the new tests cover the cap before cleanup, no persistence for unknown keys, no recompute on rejection, coalescing, churn consistency, and stop-after-close. They call `processValidatedHeartbeat` synchronously, so the async spawn in `ProcessReceivedMessage` isn't exercised end-to-end.
- **PR description:** matches the code.

Reviewed at `5045a77`.

node/heartbeat/process/monitor_test.go
```diff
@@ -1282,3 +1283,355 @@ func TestMonitor_ProcessReceivedMessageConcurrentNearCap(t *testing.T) {
...
+	assert.Equal(t, 1, afterRejected)
+}
+
+func waitForScheduledRecomputes(t *testing.T, mon *process.Monitor) {
+	deadline := time.Now().Add(5 * time.Second)
+	for mon.HasPendingRecompute() {
+		require.True(t, time.Now().Before(deadline), "scheduled recompute did not finish")
+		time.Sleep(time.Millisecond)
+	}
+}
```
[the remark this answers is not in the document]
RomuloSiebra: Replaced the sleep loop with a channel signalled from the status handler stub; the three scheduler tests now wait on it with a bounded select.

node/heartbeat/process/monitor_test.go
```diff
@@ -1282,3 +1283,355 @@ func TestMonitor_ProcessReceivedMessageConcurrentNearCap(t *testing.T) {
...
+
+	arg := createMockArgHeartbeatMonitor()
+	arg.PubKeysList = []string{}
+	arg.HeartbeatRefreshIntervalInSec = 3600
+	arg.MaxDurationPeerUnresponsive = 10 * time.Second
+	arg.MaxUnknownHeartbeatPubKeys = 64
+	arg.MaxUnknownHeartbeatPubKeysPerOrigin = 64
+
+	mon, err := process.NewMonitor(arg)
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
RomuloSiebra: Added, and did the same for the other scheduler tests while there.

node/heartbeat/process/monitor.go
RomuloSiebra: Good catch, added the guard. The over-cap rejection path no longer takes the map lock.

node/heartbeat/process/monitor.go
RomuloSiebra: Right, took the second option: when the under-lock re-check finds the key admitted it now drops the transient entry. Added a racing admit/heartbeat test for it.

node/heartbeat/process/monitor.go
RomuloSiebra: Agreed, swapped it for a 1-slot channel consumed in runRefreshLoop, plus a priority stop check at the loop top so no new walk starts after Close. The runner is wg-tracked now.

node/heartbeat/process/monitor.go
RomuloSiebra: Took the var nit. Kept the early return though: collapsing it would make every over-cap rejection take the map lock for the re-check, which is the contention the len(droppedPubKeys) guard removes on the attacker-controlled path.

node/heartbeat/process/monitor.go
```diff
@@ -810,14 +926,18 @@ func (m *Monitor) getNumInstancesOfPublicKey(pubKeyStr string) uint64 {
 // Cleanup will delete all the entries in the heartbeatMessages map
 func (m *Monitor) Cleanup() {
```
Beroni, line 975: **cut · shrink** (lazy-senior-dev) · ~6 lines saved · optional

Drop the `Cleanup()` snapshot. `Cleanup()` copies `transientUnknownHeartbeatPubKeys` into `trackedUnknownPubKeys`, but it still calls `isTransientUnknownHeartbeatPubKeyTracked(k)` on every miss, so the live check makes every deletion decision. The snapshot only saves a lock on keys that hit.

**Fix:** keep the sweep under the transient lock, remove the snapshot map and its loop, and in the loop use `if !m.isTransientUnknownHeartbeatPubKeyTracked(k) { delete…; delete… }`.

**Other side:** the PR says the snapshot is deliberate — it avoids re-locking for each entry, and in steady state most unknown entries are hits, so it does save locks. **Author's call.**

node/heartbeat/process/monitor.go
```diff
@@ -373,36 +379,54 @@ func (m *Monitor) ProcessReceivedMessage(message p2p.MessageP2P, fromConnectedPe
...
-func (m *Monitor) addHeartbeatMessageToMap(hb *data.Heartbeat, fromConnectedPeer core.PeerID) {
+func (m *Monitor) processValidatedHeartbeat(hb *data.Heartbeat, fromConnectedPeer core.PeerID) {
+	if !m.addHeartbeatMessageToMap(hb, fromConnectedPeer) {
+		return
+	}
+
+	m.scheduleHeartbeatRecompute()
+}
+
+func (m *Monitor) scheduleHeartbeatRecompute() {
```
Beroni, line 413: **cut · yagni** (lazy-senior-dev) · ~5 lines saved · optional

The `stopCh` pre-check in `scheduleHeartbeatRecompute` has no visible effect. `recomputeCh` has a 1-slot buffer and nothing reads it after the refresh loop exits, so a send after `Close()` is harmless. The pre-check exists only so a test can assert `HasPendingRecompute() == false`, which is also why the `DrainPendingRecompute` export exists.

**Fix:** reduce the function to the non-blocking send. Drop `DrainPendingRecompute` and the last three assertions of `TestMonitor_ScheduledRecomputesStopAfterClose`. `requireNoWalkQueued` already proves that no recompute runs after `Close`.

Beroni (commented):
## Automated review — Bazel

**Verdict: 💬 COMMENT.** No confirmed defects: the precision (senior-code-reviewer: approve) and adversarial (exploit-digger: comment) lenses both came back clean. lazy-senior-dev left two optional simplification cuts inline (~11 lines total); neither touches code the other lenses marked as load-bearing.

### Coverage (exploit-digger)

- **Injection:** no query, command, template or path building. Pubkeys are used only as map keys and storer keys.
- **Authz:** only admitted keys are persisted or added to `fullPeersSlice`. `SaveMultipleHeartbeatMessageInfos` now skips non-admitted keys; `confirmHeartbeatAdmissionLocked` re-checks admission under the map lock before creating any live entry; no path promotes an unknown identity to admitted.
- **Validation:** antiflood, `CreateHeartbeatFromP2PMessage` and the pid == originator check are unchanged and still run before the async `processValidatedHeartbeat`.
- **Secrets:** nothing new. The new debug logs only include error strings.
- **Logic, live state vs tracked set:** live unknown state always stays within the tracked set. Every removal (sweep, evict, untrack on promotion, `Cleanup`) is followed by a drop under `mutHeartbeatMessages` with a live re-check; the creating goroutine re-checks under the map lock, so an entry evicted between tracking and locking is rejected; if a stale `Cleanup` snapshot keeps an evicted entry, the evicting goroutine drops it once it takes the map lock.
- **Logic, lock order:** `mutHeartbeatMessages` is always taken before `mutAdmitted…` and `mutTransient…`. No function holds the transient lock while taking another lock; no inversion and no recursive `RLock`.
- **Error handling:** if `getOrCreateHeartbeatMessageInfoLocked` fails, the function unlocks, returns false and schedules no recompute — no orphaned state.
- **Resources:** `recomputeCh` has a 1-slot buffer with non-blocking sends, so bursts collapse into a single walk (replacing one compute goroutine per message, including rejected ones). The sweep is O(cap), which is bounded.
- **Crypto and dependencies:** nothing touched. `sync/atomic` is added in tests only.
- **Regression:** the removed accessor had no other callers. None of the three callers of `SaveMultipleHeartbeatMessageInfos` loses a write it needs. `Close` still waits for the loop.
- **Test gaps (not blocking):** the churn test uses `maxDuration=1h`, so the time-based sweep path in `trackTransient` never runs; eviction racing with a concurrent promotion through `computeInactiveHeartbeatMessages` is not covered.
- **PR vs code:** the implementation matches the PR description.

Reviewed at `ce411f7` by senior-code-reviewer, exploit-digger and lazy-senior-dev.

node/heartbeat/process/monitor.go
```diff
@@ -373,22 +379,59 @@ func (m *Monitor) ProcessReceivedMessage(message p2p.MessageP2P, fromConnectedPe
...
+	select {
+	case <-m.stopCh:
+		return
+	default:
+	}
+
+	select {
+	case m.recomputeCh <- struct{}{}:
```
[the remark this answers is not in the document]
RomuloSiebra: Right, the select could still pick the token when both were ready. Both arms now re-check stop before doing any work, so shutdown wins deterministically. No clean way to pin the parked-in-select interleaving from a test without hooks in production code, so this one rides on the existing close test and the race runs.

node/heartbeat/process/monitor_test.go
```diff
@@ -1282,3 +1283,386 @@ func TestMonitor_ProcessReceivedMessageConcurrentNearCap(t *testing.T) {
...
+	drainWalks(walks)
+
+	origin := core.PeerID("origin-a")
+	mon.ProcessValidatedHeartbeat(&data.Heartbeat{Pubkey: []byte("unknown-accepted"), Pid: []byte("pid-1")}, origin)
+	waitForWalk(t, walks)
+	requireNoWalkQueued(t, walks)
+
+	mon.ProcessValidatedHeartbeat(&data.Heartbeat{Pubkey: []byte("unknown-rejected"), Pid: []byte("pid-2")}, origin)
+	assert.False(t, mon.HasPendingRecompute())
+	requireNoWalkQueued(t, walks)
```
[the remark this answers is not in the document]
RomuloSiebra: On the rejected-heartbeat test: one token means one walk, and a rejected heartbeat never sends a token, so once the first walk has signalled there is nothing left to be in flight, both assertions are deterministic. On the admission race: rather than adding checkpoints to production code, the admitted path now always releases any transient slot, and there is a deterministic test that builds the admitted-plus-tracked state through the export hooks and checks a single heartbeat clears it. The racing test stays as well.

node/heartbeat/process/monitor.go
RomuloSiebra: Done as suggested, the snapshot is gone and each unknown entry is decided on the live check.

node/heartbeat/process/monitor.go
RomuloSiebra: Done, schedule is just the non-blocking send now and the drain hook and the trailing assertions are gone. The loop-side stop checks are what guarantee no walk runs after Close.

Beroni:
## Automated review — Bazel

> Reviewed at `fa7375b` per the report, but Bazel recorded head `56ca864` for this run. Since the two don't match, findings are not line-anchored; `file:line` references point at `fa7375b`.

**Verdict: 💬 COMMENT.** Every lens (`senior-code-reviewer`, `exploit-digger`, `lazy-senior-dev`) returned `comment`. No blocking defect was found.

The core fix holds up. The adversarial lens traced these paths and found each one correct:
- the eviction and sweep interleavings;
- the lock order: `mutHeartbeatMessages` → `mutAdmittedHeartbeatPubKeys` → `mutTransientUnknownHeartbeatPubKeys`, with no inversion;
- the admission gate in `SaveMultipleHeartbeatMessageInfos`;
- the coalesced recompute channel.

`go test ./node/heartbeat/... -race` passes.

---

### 1. **low · logic** · pre-existing, narrowed by this PR — A key admitted after its first heartbeat is not persisted until its next heartbeat
`node/heartbeat/process/monitor.go:438` — *exploit-digger*

`addHeartbeatMessageToMap` fixes the admission status when `confirmHeartbeatAdmissionLocked` runs. After that, if `!isAdmittedPubKey` it returns early and skips `SavePubkeyData` and `addPeerToFullPeersSlice`. There are two ways a key can get admitted and still not be saved:
- `computeInactiveHeartbeatMessages` calls `markHeartbeatPubKeyAsAdmitted` after the map lock is released at `:426` but before `:438`.
- A refresh admits a key that already has an `hbmi`. `computeInactiveHeartbeatMessages` only creates and saves a record when the `hbmi` is nil (`:716`).

**Trigger:** a new validator's first heartbeat arrives before `peerTypeProvider` lists it. The key is tracked as transient and gets an `hbmi`. The next refresh admits it, but nothing is persisted and the key is not added to `fullPeersSlice` until that validator sends another heartbeat.

**Impact:** a short persistence gap that fixes itself on the next heartbeat. `develop` has the same pre-lock flag and early return, so this is not a regression. The PR makes the window smaller by re-checking under the lock.

**Fix:** in `computeInactiveHeartbeatMessages`, add keys that were just admitted and already have an `hbmi` to the save set and to `fullPeersSlice`. Alternatively, re-check `isAdmittedHeartbeatPubKey` before the early return at `:438`. No test covers this window.

### 2. **low · docs** — The PR description no longer matches the code at HEAD
`node/heartbeat/process/monitor.go:930` — *exploit-digger*, confirmed by the orchestrator

The description says `Cleanup()` "snapshots the tracked set once". Commit `fa7375b` removed that snapshot. `Cleanup` now calls `isTransientUnknownHeartbeatPubKeyTracked` for each non-admitted entry while it holds the map write lock. That is correct and bounded by `maxUnknownHeartbeatPubKeys`.

The description also mentions a "dirty-flag single runner" and `isStopped`. The code uses `recomputeCh` (buffered 1, non-blocking send), drains it on the existing `runRefreshLoop` goroutine, and names the helper `isStopping`.

**Fix:** update the Solution and Key Changes sections of the PR body to match `fa7375b`.

### 3. **suggestion · docs** — The doc comment on `SaveMultipleHeartbeatMessageInfos` is stale
`node/heartbeat/process/monitor.go:221` — *senior-code-reviewer* (confidence 8)

The comment still says it "stores all heartbeatMessageInfos". The function now silently skips every key that is not admitted, so a caller relying on the comment would expect unknown or transient entries to be persisted.

**Fix:**
```go
// SaveMultipleHeartbeatMessageInfos stores the given heartbeatMessageInfos to the storer,
// skipping any public key that is not admitted (transient/unknown identities are never persisted)
```

---

### Cuts (lazy-senior-dev)

**C2. delete** — Remove the `isStopping()` check at the top of the loop in `runRefreshLoop` (~3 lines)
`node/heartbeat/process/monitor.go:851`

When `stopCh` is closed, the `select` either takes the `stopCh` case or takes a case that already checks `isStopping()` before doing any work. The checks inside the cases are what give shutdown priority, so keep those.

*C1 (inline `admitHeartbeatPubKey` and merge its second lock acquisition on the rejection path into the locked re-check) was already raised in an earlier round and is not repeated here.*

---

<details>
<summary>Coverage (exploit-digger)</summary>

| Class | Result |
|---|---|
| injection | Clean. The pubkey is used only as a map key and as the storer key for admitted identities. |
| authz | Clean. Admission comes only from trusted sources (`peerTypeProvider`, `pubKeysList`, storage filter). It is re-checked under the map lock before any live state is created. `SaveMultipleHeartbeatMessageInfos` is now gated, so unknown identities cannot reach storage through the changed-to-inactive path. |
| validation | Clean. Pid/originator and decoding checks are unchanged, and antiflood runs before the goroutine is spawned. The per-origin cap cannot exceed the global cap. |
| secrets | Clean. No new sensitive logging. |
| logic: eviction races | Clean. A key re-tracked between eviction and drop keeps its state. A key evicted before confirm is rejected. A key evicted after creation is deleted by the later drop. `Cleanup` catches whatever is left, so unknown live entries stay bounded by the cap. |
| logic: lock order | Clean. No inversion on any path, and the race test passes. |
| error handling | Clean. When `newHeartbeatMessageInfo` fails, the code unlocks and schedules no recompute. The transient slot is swept later. |
| resource | Clean. `recomputeCh` has a buffer of 1 with a non-blocking send, so there is no blocking or leak after `Close`. Recomputes are coalesced onto one runner, and `Close` waits on `wg`. |
| crypto / deps | Not touched. No `go.mod` changes and no new imports. |
| regression | Clean. Rejected heartbeats no longer trigger a recompute, which is intended; the ticker still keeps metrics current. The initial `pubKeysList` save still works because those keys are admitted first. |
| test gap | 8 new tests cover the cap, no persistence of unknown keys, coalescing, stop after close, churn consistency and slot release. Finding 1's window has no test. |

</details>

RomuloSiebra:
Following up on the automated review at fa7375b, all four points are in the latest push.

1. Taken the first option: a key admitted by a refresh that already has live state is now persisted and added to the key list by that refresh, with a test covering the window. Only keys the registry admits reach that path, so the admission gate in SaveMultipleHeartbeatMessageInfos still applies.
2. PR description updated to match the code (live tracked check in Cleanup, recomputeCh drained on runRefreshLoop, isStopping).
3. Doc comment on SaveMultipleHeartbeatMessageInfos updated with the suggested wording.
C2. Removed the loop-top check, the in-arm ones stay.

node/heartbeat/process/monitor.go
```diff
@@ -614,20 +715,31 @@ func (m *Monitor) computeInactiveHeartbeatMessages() {
...
+		}
+		if newlyAdmitted {
+			inactiveHbChangedMap[peerTypeInfo.PublicKey] = hbmi
+			newlyAdmittedPubKeys = append(newlyAdmittedPubKeys, []byte(peerTypeInfo.PublicKey))
 		}
 	}
 
 	m.mutHeartbeatMessages.Unlock()
 	m.SaveMultipleHeartbeatMessageInfos(inactiveHbChangedMap)
+	for _, pubKey := range newlyAdmittedPubKeys {
```
Beroni: ### 1. `SaveKeys` is issued once per newly admitted key

**performance** · senior-code-reviewer (conf. 8) + exploit-digger (high) — 🔴 **found independently by two lenses**

`computeInactiveHeartbeatMessages` ends with `for _, pubKey := range newlyAdmittedPubKeys { m.addPeerToFullPeersSlice(pubKey) }`. `addPeerToFullPeersSlice` (monitor.go:499) does a linear `isPeerInFullPeersSlice` scan **and then marshals and writes the entire key list** via `storer.SaveKeys(m.fullPeersSlice)` on every call. Admitting K identities in one refresh costs K full-list marshal+DB writes — O(K·N) bytes.

**Trigger:** this runs on the single `runRefreshLoop` goroutine, so for its whole duration the ticker refresh *and* every heartbeat-driven recompute on `recomputeCh` are blocked. The first refresh after `peerTypeProvider.UpdateCache(epoch)` makes `GetAllPeerTypeInfos()` non-empty, or an epoch boundary promoting nodes previously tracked as unknown. K is bounded by the unknown-identity cap (default 1024).

**Fix** — append under `mutFullPeersSlice`, persist once:

```go
m.mutFullPeersSlice.Lock()
changed := false
for _, pubKey := range newlyAdmittedPubKeys {
    if !m.isPeerInFullPeersSlice(pubKey) {
        m.fullPeersSlice = append(m.fullPeersSlice, pubKey)
        changed = true
    }
}
if changed {
    if err := m.storer.SaveKeys(m.fullPeersSlice); err != nil {
        log.Debug("can't store the keys slice", "error", err.Error())
    }
}
m.mutFullPeersSlice.Unlock()
```

node/heartbeat/process/monitor.go
```diff
@@ -614,20 +715,31 @@ func (m *Monitor) computeInactiveHeartbeatMessages() {
...
+		m.addPeerToFullPeersSlice(pubKey)
```
Beroni: ### 2. The one-shot `newlyAdmitted` edge is the sole gate for key-list persistence

**correctness / error-handling** · exploit-digger

`markHeartbeatPubKeyAsAdmitted` returns `true` exactly once per key for the process lifetime (the admitted set is never pruned), and that single edge gates **both** `SaveMultipleHeartbeatMessageInfos` and `addPeerToFullPeersSlice`. Two ways the edge is consumed without the key reaching the key list:

- **The `hbmi == nil` branch** (line 729): `markHeartbeatPubKeyAsAdmitted` is called *before* the nil check, then the branch creates the live entry and `continue`s — deliberately skipping both `inactiveHbChangedMap` and `newlyAdmittedPubKeys`. The key is now admitted and live but absent from `fullPeersSlice`. If it later trips `getValsForUpdate`, `SaveMultipleHeartbeatMessageInfos` writes a record for it (it passes the new admission gate) while it is still not in the key list. `loadRestOfPubKeysFromStorage` prunes only what `LoadKeys` returns — so that record is unreachable by startup pruning forever. **This is the same failure shape the PR states as its root cause**, surviving for admitted keys.
- **A storer error**: both `SaveKeys` (line 506) and `SavePubkeyData` (line 235) only `log.Debug` their errors. A transient failure at exactly that refresh permanently drops the identity from the key list; no later refresh retries, because `markHeartbeatPubKeyAsAdmitted` now returns `false`. Only a subsequent inbound heartbeat recovers it, so an identity that goes quiet is lost until restart.

**Fix** — make key-list membership a postcondition of every `SavePubkeyData` rather than a consequence of a one-shot edge: call `addPeerToFullPeersSlice` for every key persisted by `SaveMultipleHeartbeatMessageInfos`, and add the `hbmi == nil` branch's key to `newlyAdmittedPubKeys` before `continue`. At minimum, retry on the next refresh when the previous write returned an error.

node/heartbeat/process/monitor.go
```diff
@@ -373,36 +380,48 @@ func (m *Monitor) ProcessReceivedMessage(message p2p.MessageP2P, fromConnectedPe
...
-			m.mutHeartbeatMessages.Unlock()
-			return
-		}
-		m.heartbeatMessages[pubKeyStr] = hbmi
+	hbmi, err := m.getOrCreateHeartbeatMessageInfoLocked(hb)
+	if err != nil {
+		log.Debug("error creating heartbeat message info", "error", err.Error())
+		m.mutHeartbeatMessages.Unlock()
+		return false
 	}
```
Beroni: ### 3. `doubleSignerPeers` entry is created before the `hbmi` that can fail

**resource / invariant** · exploit-digger

In `addHeartbeatMessageToMap`, `addDoubleSignerPeers(hb)` runs *before* `getOrCreateHeartbeatMessageInfoLocked(hb)`. If hbmi creation fails, the function returns having inserted only the `doubleSignerPeers` entry. `Cleanup()` discards its sweep result (`_ = m.sweepTransientUnknownHeartbeatPubKeysLocked()`, line 943) and reclaims live state by iterating `heartbeatMessages` only — `doubleSignerPeers` is never iterated independently. So a desynchronized entry is invisible to `Cleanup` forever, and once its transient slot has been swept no future `dropLiveHeartbeatStateLocked` can ever name it.

**Trigger:** reachable today only with `MaxDurationPeerUnresponsive == 0`, which `NewMonitor` never validates (unlike every other arg) but which `HeartbeatHandler.checkConfigParams` enforces `>= 1s` in production wiring. **Latent invariant hole, not a remotely reachable leak.**

**Fix** — three cheap independent hardenings: create the `doubleSignerPeers` entry only after the hbmi exists; feed the sweep result into `dropLiveHeartbeatStateLocked` inside `Cleanup` (line 943) instead of discarding it; validate `arg.MaxDurationPeerUnresponsive != 0` in `NewMonitor`.

node/heartbeat/process/monitor.go
```diff
@@ -450,42 +516,71 @@ func (m *Monitor) isAdmittedHeartbeatPubKey(pubKey string) bool {
...
 	m.transientUnknownHeartbeatPubKeys[pubKey] = transientUnknownHeartbeatInfo{
 		originPeer: originPeer,
 		lastSeen:   m.timer.Now(),
 	}
 
-	return true
+	return true, droppedPubKeys
+}
+
+func (m *Monitor) dropLiveHeartbeatStateLocked(pubKeys []string) {
```
Beroni, line 603: ### 4. The `...Locked` suffix now denotes two different mutexes in one file

**maintainability** · senior-code-reviewer (conf. 9)

The `...Locked` suffix now denotes two different mutexes in one file. `sweepTransientUnknownHeartbeatPubKeysLocked` / `countTransientUnknownHeartbeatPubKeysForOriginLocked` / `evictOldestTransientUnknownHeartbeatPubKeyLocked` require `mutTransientUnknownHeartbeatPubKeys`; `dropLiveHeartbeatStateLocked` / `confirmHeartbeatAdmissionLocked` / `getOrCreateHeartbeatMessageInfoLocked` require `mutHeartbeatMessages`.

Worse, `dropLiveHeartbeatStateLocked` itself **acquires** `mutAdmittedHeartbeatPubKeys` and `mutTransientUnknownHeartbeatPubKeys` internally — so calling it from a transient-lock-held context self-deadlocks on the non-reentrant `sync.Mutex`. The entire sweep/evict-returns-dropped-keys design exists precisely so that never happens, but nothing in the name says so. Every current call site is correct; the hazard is for the next change.

**Fix** — encode the mutex in the name and document the internal acquisitions, e.g.:

```go
// requires mutHeartbeatMessages held; acquires mutAdmittedHeartbeatPubKeys and
// mutTransientUnknownHeartbeatPubKeys. Never call with mutTransientUnknownHeartbeatPubKeys held.
```

node/heartbeat/process/monitor.go
```diff
@@ -418,16 +437,63 @@ func (m *Monitor) addHeartbeatMessageToMap(hb *data.Heartbeat, fromConnectedPeer
...
+	err = m.storer.SavePubkeyData(hb.Pubkey, hbDTO)
 	if err != nil {
 		log.Debug("cannot save heartbeat to db", "error", err.Error())
 	}
 	m.addPeerToFullPeersSlice(hb.Pubkey)
+
+	return true
+}
+
+func (m *Monitor) admitHeartbeatPubKey(pubKeyStr string, fromConnectedPeer core.PeerID) (bool, []string, bool) {
```
Beroni: ### 5. Unnamed bool returns on the admission hot path

**readability** · senior-code-reviewer (conf. 9)

`admitHeartbeatPubKey` returns `(bool, []string, bool)` and `confirmHeartbeatAdmissionLocked` returns `(bool, bool)` — two bare bools with opposite meanings (`isAdmitted` vs `accepted`/`stillAccepted`). Swapping them at a `return` site compiles silently and flips admission on the heartbeat hot path.

**Fix** — `(isAdmitted bool, droppedPubKeys []string, accepted bool)` and `(isAdmitted bool, stillAccepted bool)`.

Beroni (commented):
## Automated review — Bazel

**Verdict: 💬 COMMENT.** All three lenses (`senior-code-reviewer`, `exploit-digger`, `lazy-senior-dev`) returned `comment`; no blocker found.

| | |
|---|---|
| **Scope** | `git diff origin/develop...HEAD` |
| **Files** | 3 (`monitor.go`, `monitor_test.go`, `export_test.go`) — +665 / −51 |
| **Base / head** | `develop` / `bbd59b9` |
| **Lenses** | `senior-code-reviewer`, `exploit-digger`, `lazy-senior-dev` — all three ran |

The core of the PR holds up. The adversarial lens traced the cap invariant exhaustively and confirmed it: no live `heartbeatMessages` entry can exist for a non-admitted key without a transient slot, every path that removes a slot hands the key to a caller that drops live state under the map lock, and the documented lock order (`mutHeartbeatMessages` → `mutAdmittedHeartbeatPubKeys` → `mutTransientUnknownHeartbeatPubKeys`) is acyclic with no inversion. Goroutine accounting improved rather than regressed. The 5 findings posted inline are all on the periphery of that core.

<details>
<summary><b>Coverage — what was swept and came back clean</b></summary>

| Class | Result |
|---|---|
| **Cap invariant (logic)** | Traced exhaustively. A live entry for a non-admitted key requires `confirmHeartbeatAdmissionLocked` under the map lock + `isTransientUnknownHeartbeatPubKeyTracked`. `markHeartbeatPubKeyAsAdmitted` sets the admitted flag *before* untracking, so `dropLiveHeartbeatStateLocked` can never delete a key mid-admission. Sweep-then-reinsert self-eviction guarded at line 569; `evictOldest` cannot select the incoming key (runs before insertion). |
| **Lock ordering** | Acyclic, no inversion. Every `mutTransientUnknownHeartbeatPubKeys` acquisition releases before any `mutHeartbeatMessages` acquisition; `markHeartbeatPubKeyAsAdmitted` releases `mutAdmittedHeartbeatPubKeys` before calling untrack. `mutFullPeersSlice` only ever taken alone; `hbmi.updateMutex` always innermost. No new lock introduced. |
| **Authz** | Admission never sees attacker data — `markHeartbeatPubKeyAsAdmitted` is reachable only from config `PubKeysList`, filtered storage load, and `peerTypeProvider`. No network-supplied pubkey can reach it. |
| **Resource** | `ProcessReceivedMessage` now spawns **one** goroutine per message instead of two; the unbounded per-message `computeAllHeartbeatMessages` fan-out is replaced by the 1-slot coalescing channel. `recomputeCh` is never closed and every send is non-blocking — sends after `Close` neither panic nor leak. `fullPeersSlice` and `admittedHeartbeatPubKeys` grow only with validator-set-bounded keys. `PeerShardMapper` sibling state is backed by size-bounded caches. |
| **Validation** | Heartbeat fields bounded to 128 bytes by `verifyLengths` before the monitor sees them; `hb.Pid` pinned to `message.Peer()` twice; BLS signature verified before any state is created. Only unvalidated arg is `MaxDurationPeerUnresponsive` (finding 3). |
| **Regression** | `addHeartbeatMessageToMap`, `computeInactiveHeartbeatMessages` and `Cleanup` diffed line by line against `develop`. `getOrCreateHeartbeatMessageInfoLocked`'s `if ok && hbmi != nil` is equivalent to the old guard; the new `continue` skips only new code; `isTransientUnknownHeartbeatPubKeyTracked` is behaviourally identical to the removed `transientUnknownHeartbeatInfo` helper. The one intentional removal (skipping non-admitted keys in `SaveMultipleHeartbeatMessageInfos`) can only ever skip transient unknowns. |
| **Tests** | New tests do cover the risky paths — concurrent admission vs. tracking (3000 iterations), churn with concurrent `Cleanup` asserting exact equality across all three maps, recompute coalescing, drop-on-close. Synchronise on channels, as claimed. `go test -race -count=1 ./node/heartbeat/process/...` passes. **Not covered:** the `getOrCreateHeartbeatMessageInfoLocked` error branch, storer errors on the `newlyAdmitted` path, and a refresh admitting many keys at once — i.e. findings 1, 2 and 3. |
| **Injection / secrets / crypto / deps** | Clean. No string-built queries or paths; storer keys are raw `[]byte`; logs are structured k/v. No credentials handled; `export_test.go` is `_test`-guarded and unreachable from production binaries. No crypto touched. No dependency added; only new import is stdlib `sync/atomic` in tests. |

</details>

**Recommendation:** finding 1 is the one worth fixing before merge — two lenses reached it independently, it's a few lines, and it blocks the refresh loop. Findings 2 and 3 are real but can ship as follow-ups; 4 and 5 are free.

Reviewed at `bbd59b9`.

node/heartbeat/process/monitor.go
fbsobreira: Agreed, this one should be fixed before merge. `addPeerToFullPeersSlice` re-encodes and rewrites the whole key list on every call, and this loop is new in the PR, so admitting K keys in one refresh means K full rewrites on the refresh-loop goroutine. That goroutine also serves `recomputeCh`, so recomputes wait behind it. The suggested fix looks right: append everything under `mutFullPeersSlice`, then call `SaveKeys` once, only if something changed. It would also help to pull that into a helper that takes a slice, so the heartbeat path and this path share it.

node/heartbeat/process/monitor.go
fbsobreira: Both points are right, but I'd handle them in a follow-up, not in this PR.

- **`hbmi == nil` branch:** `develop` already behaves this way. Its `computeInactiveHeartbeatMessages` marks the key admitted, creates the entry and never adds it to the key list, and a later `getValsForUpdate` hit saves it the same way. The keys that reach this branch come only from `peerTypeProvider` (the validator set), not from the network, so this doesn't reintroduce the unbounded, attacker-driven growth this PR fixes. The cleanest fix is the first one you suggested: have `SaveMultipleHeartbeatMessageInfos` add every key it saves to the key list, so being in the list always follows from being saved. That goes well with the batched write from finding 1.
- **Storer errors:** agreed that nothing retries once the one-shot admission edge has fired. `develop` never wrote the key list from the refresh at all, so this PR is still an improvement. The fix above would also cover retries, because the next save of that key would add it to the list again.

node/heartbeat/process/monitor.go
fbsobreira: Confirmed that this can't happen today. `newHeartbeatMessageInfo` fails only on a zero `maxDurationPeerUnresponsive` or a nil timer, `NewMonitor` already checks the timer, and `checkConfigParams` requires the duration to be at least 1s. The `addDoubleSignerPeers` ordering is also the same as on `develop`. Of the three hardenings, the `NewMonitor` check is the one worth adding. It's a single `if`, and it removes the only way to reach the error branch. With a zero duration the transient sweep would also expire every entry immediately. I'm fine with that going in here or in the follow-up.

node/heartbeat/process/monitor.go
fbsobreira: Agreed, it's cheap to do now. The self-deadlock is real: `dropLiveHeartbeatStateLocked` → `isTransientUnknownHeartbeatPubKeyTracked` takes `mutTransientUnknownHeartbeatPubKeys`, so calling it from inside the sweep would hang. A one-line comment per function stating which lock it expects the caller to hold, plus the "never call with the transient lock held" note on `dropLiveHeartbeatStateLocked`, is enough. No rename needed.

node/heartbeat/process/monitor.go
fbsobreira: Agreed, no cost to doing it. Named returns here would make a swap at a `return` site much easier to spot in review.

node/heartbeat/process/monitor.go
RomuloSiebra: Done, and took the helper suggestion: addPeersToFullPeersSlice takes a slice, appends under mutFullPeersSlice and writes the key list once only if something changed. The heartbeat path uses the same helper.

node/heartbeat/process/monitor.go
RomuloSiebra: Went with the cleanest version now since it fell out of the batched helper: SaveMultipleHeartbeatMessageInfos hands every key it actually wrote to the key list, so listing follows from saving, which covers the nil-hbmi branch when that key is later saved. On a SaveKeys error the in-memory append is rolled back so the next save retries. Tests for both.

node/heartbeat/process/monitor.go
RomuloSiebra: Added the NewMonitor check for a zero duration, and moved addDoubleSignerPeers after the hbmi exists since that was three lines. Skipped the Cleanup sweep-result change, the reorder makes it moot.

node/heartbeat/process/monitor.go
RomuloSiebra: Added the one-line lock requirement on each Locked function and the never-with-the-transient-lock note on dropLiveHeartbeatStateLocked. No rename.

node/heartbeat/process/monitor.go
RomuloSiebra: Done, both signatures use named results now.

fbsobreira (approved):

Beroni (approved):

Issue #144, opened 2026-09-06
Blake2b.Compute: unsynchronized write to emptyHash and shared slice returned to callers


fbsobreira:
Raised by @nickgs1337 during review of #139, filed separately so it does not grow that diff.

## Problem

`crypto/hashing/blake2b/blake2b.go:32-43`:

```go
func (b2b *Blake2b) Compute(s string) []byte {
	if len(s) == 0 {
		if len(b2b.emptyHash) == 0 {
			b2b.setEmpty()
		}
		return b2b.emptyHash
	}
	...
}
```

Two defects on the zero-length path:

1. **Unsynchronized check-then-write.** `b2b.emptyHash` is read, tested, and assigned with no synchronization. A `Blake2b` instance shared across goroutines racing on empty input is a data race on both the slice header and the backing array.
2. **The shared slice is returned directly.** Callers receive the instance's own `emptyHash`, so any caller mutating the returned digest corrupts every subsequent empty-input result.

`crypto/hashing/sha256/sha256.go:12-40` already has the right shape for both — a package-level `sync.Once` plus a copy on return:

```go
func (sha Sha256) EmptyHash() []byte {
	sha256Once.Do(func() { ... })
	result := make([]byte, len(sha256EmptyHash))
	copy(result, sha256EmptyHash)
	return result
}
```

`Blake2b.EmptyHash()` (lines 46-51) has the same check-then-write and needs the same treatment.

## Reachability

`NewAuthenticationFunc` builds one hasher instance captured by the closure and shared across every request, so an unauthenticated caller sending Basic Auth with an empty password reaches `Compute("")` concurrently. This does **not** bite today: the shipped `config/node/api.yaml` sets `hasher.type: sha256`, and `Sha256` is already safe. An operator switching `hasher.type` to `blake2b` would activate it.

The trie also uses hashers concurrently (`data/trie/branchNode.go`, `extensionNode.go`), so the fix is worth making regardless of the API path.

## Suggested fix

In `crypto/hashing/blake2b`: guard the empty-hash initialization with a `sync.Once` (per-instance, since `HashSize` is a field) and return a copy from both `Compute("")` and `EmptyHash()`. Add a `-race` test that hammers `Compute("")` from several goroutines on one shared instance.

fbsobreira:
Tracked in Jira as [KLC-2648](https://klever-io.atlassian.net/browse/KLC-2648).

[...]

PR #147, opened 2026-09-18
[KLC-2648] fix(hashing): make blake2b empty-hash init race-free and non-aliasing

fbsobreira:
## Summary

`Blake2b.Compute("")` and `Blake2b.EmptyHash()` did an unsynchronized check-then-write on `b2b.emptyHash` and then returned the instance's own slice. Two defects on the zero-length path:

1. **Unsynchronized check-then-write.** A single `Blake2b` instance shared across goroutines racing on empty input is a data race on both the slice header and the backing array.
2. **The shared slice was returned directly.** Any caller mutating the returned digest corrupted every later empty-input result.

`crypto/hashing/sha256` already had the right shape. This brings blake2b in line: guard the lazy initialization, and return a copy from both entry points.

Jira: KLC-2648 · Issue: #144 · Reported by @nickgs1337 during review of #139.

## Two deliberate deviations from the suggested fix

**1. A mutex instead of `sync.Once`.** The issue suggested `sync.Once`, and sha256/keccak/fnv all use one. blake2b is the only variable-size hasher, so it is the only one whose initialization can fail — and `sync.Once` marks itself done even when its function panics. With a `Once`, a `Blake2b{HashSize: 65}` panicked on the first call and then **silently returned a zero-length digest on every call after**, while `Size()` still reported 65. Re-checking under a mutex keeps that misconfiguration loud on every call. The rationale is documented on the struct field.

**2. `getHasher()` now checks its error.** It discarded the error from `blake2b.New`, which returns a nil `*digest` boxed in a **non-nil** `hash.Hash`, so returning it nil-dereferenced inside `Sum`. It now panics with the cause. `HashSize` is only ever set from compile-time constants or source literals — never config, never input — so this is a programming-error guard, not a reachable DoS.

## Correction to the reachability analysis

Issue #144 states *"this does not bite today: the shipped `config/node/api.yaml` sets `hasher.type: sha256`."* **That is true only of the API auth hasher.** The node's core hasher ships as blake2b:

```
config/node/config.yaml:558   hasher:          type: blake2b
config/node/config.yaml:560   multisigHasher:  type: blake2b
config/node/config.yaml:562   txSignHasher:    type: blake2b
```

`factory/coreComponents.go:39` builds one instance from that and shares it node-wide, and it does reach the empty path in production — `bp.hasher.EmptyHash()` at `core/process/block/baseProcess.go:260,263`, and `mt.hasher.EmptyHash()` at `data/block/block.go:300`. So the path is live under shipped config, not dormant.

**Still hardening, not an advisory**, for different reasons than the issue gave:

- The one attacker-controlled vector — Basic Auth with an empty password — runs on the API hasher, which genuinely is sha256, so it never reaches blake2b.
- The race window is first-initialization only; after the first write every call is a plain read. It cannot be reopened or timed by an outsider.
- All racing writers write the same digest at the same length, so even a torn header yields correct bytes. The real hazard is memory ordering on arm64 — a reader could see the new pointer before the backing array is visible and get a zeroed digest, flipping the `bytes.Equal` at `baseProcess.go:260` and diverging that one node on an empty block. A single-node correctness hazard, not remotely triggerable, not memory-unsafe.
- Neither call site mutates the returned digest, so the aliasing half has no live trigger today.

Not verified: that those two call sites are actually entered from two goroutines simultaneously. Reachability and instance-sharing are confirmed — racy by construction — but a concurrent hit was not observed.

## Tests

- `TestBlake2b_ConcurrentEmptyInputIsRaceFree` — 64 goroutines on one **cold** instance, mixing `Compute("")` and `EmptyHash()`. The shared suite's `testConcurrentEmptyHash` calls `EmptyHash()` once *before* spawning its goroutines, so the cache is already warm and the lazy init never races — which is why it never caught this. Possible follow-up: strengthen that helper for all hashers.
- `TestBlake2b_ComputeEmptyReturnsCopy` — the `Compute("")` entry point, which the shared suite does not exercise.
- `TestBlake2b_InvalidHashSizePanicsOnEveryCall` — a failed init must not be cached.
- blake2b wired into the shared suite (`TestBlake2b_ConcurrentEmptyHash`, `TestBlake2b_EmptyHashReturnsCopy`), which previously covered only keccak/sha256/fnv.

All three new tests were watched failing first: the concurrency test reported `WARNING: DATA RACE` on `setEmpty` vs `Compute`/`EmptyHash`; the copy test returned the XOR-flipped digest; the invalid-size test panicked on call 1 and silently succeeded on calls 2 and 3.

## Verification

- `go test -race -count=2 ./crypto/hashing/...` — all 6 packages pass
- `go test -race ./crypto/... ./storage/bloom/... ./data/trie/...` — pass
- `go build ./...`, `gofmt`, and `go vet` over all 12 packages referencing `Blake2b` — clean (the `sync.Mutex` field makes `Blake2b` non-copyable; no call site copies it)

No hash output changes, so no audit-fork gating.

## Summary

- Fixed race conditions in `Blake2b.Compute("")` and `Blake2b.EmptyHash()`.
- Returned digest copies to prevent cached-slice mutation.
- Preserved repeated panics for invalid hash sizes.
- Added error handling for `blake2b.New`.
- Added concurrency, copy-isolation, invalid-size, and shared hasher tests.

## Impact

- The change affects the hashing utility layer.
- It does not directly change consensus, transaction processing, state management, KVM, or networking.
- It improves node stability and data integrity by preventing data races and cached digest corruption.
- Valid Blake2b digest results remain unchanged.


crypto/hashing/blake2b/blake2b.go
```diff
@@ -10,8 +12,23 @@ import (
...
+	// emptyHash caches the digest of the empty string. It is computed lazily because it
+	// depends on HashSize, which is only known once the caller has built the struct.
+	// emptyMut guards it, and the cached digest is never handed out directly, only copied,
+	// so a caller mutating a result cannot corrupt later calls.
+	//
+	// A mutex rather than a sync.Once: sync.Once marks itself done even when its function
+	// panics, so an out-of-range HashSize would panic on the first call and then silently
+	// yield a zero-length digest on every call after. Re-checking under a mutex keeps that
+	// misconfiguration loud on every call.
+	emptyMut  sync.Mutex
```
Beroni, line 19: **nit · yagni** (lazy-senior-dev) · −18 lines

**The empty-hash cache may not be needed at all.** The cache stores the digest of an empty input, which takes one blake2b compression call to compute. Every call already copies the cached value, so the cache saves little. These could all be deleted:
- the `emptyMut` mutex
- the `emptyHash` field
- the 9-line comment on mutex vs `sync.Once`
- the lazy-init branch
- the `make`+`copy`

**Replacement:**

```go
func (b2b *Blake2b) EmptyHash() []byte {
	return b2b.getHasher().Sum(nil)
}
```

This version has no shared state, so it cannot race. It returns a new slice on every call, and an invalid `HashSize` still panics on every call. That meets all three goals of the PR, and the new tests should still pass. It also removes the need for the "must not be copied after first use" rule on the type.

**The case for keeping it:** sha256, keccak and fnv all cache their empty hash, so dropping it here would make blake2b the only hasher that doesn't. No lens flagged the cache as load-bearing. This is a style choice, so it's up to the author.

Beroni (commented):
## Automated review — Bazel

**Verdict:** comment · no correctness or security defects · 1 nit about the PR description · 1 optional simplification (−18 lines, inline)
**Lenses:** senior-code-reviewer (approve), exploit-digger (comment), lazy-senior-dev (comment)

The fix does what it says. `EmptyHash` and `Compute("")` now share a single locked path, and both return a copy of the cached digest. If `HashSize` is invalid, init panics before anything is cached, and the deferred unlock stops that panic from leaving the mutex locked. The hash output is the same as before.

The exploit-digger lens ran the new tests against the old `blake2b.go` from `origin/develop` using `go test -overlay`:
- `TestBlake2b_ConcurrentEmptyInputIsRaceFree` and the shared-suite test reported `WARNING: DATA RACE`, 7 times across 5 runs.
- Both copy tests failed.

So the new tests do catch the old bug. On HEAD, `go test -race -count=3 ./crypto/hashing/...` passes.

### PR description cites stale line numbers for the `EmptyHash` call sites
**nit · pr-mismatch** (exploit-digger)

The description says the core hasher reaches `EmptyHash()` at `core/process/block/baseProcess.go:260,263`. In the current tree those calls are at **lines 280 and 283** (`bytes.Equal(headerHandler.GetTxRootHash(), bp.hasher.EmptyHash())`). The reachability claim is correct; only the line numbers are out of date.

**Fix:** change the description to cite `baseProcess.go:280,283`.

<details>
<summary>Coverage — swept clean</summary>

- `validation`: `HashSize` is only set in source code, never from config or input (`0` in the literals at `hasherFactory.go:28`, `storageunit.go:506` and `bloom.go:65`; `multisig.BlsHashSize=16` in `factory/cryptoComponents.go:139`). Config only picks the hasher type, never the size, so the new panic in `getHasher` can't be triggered from outside. An invalid size already panicked before this PR, through a nil `*digest` dereference, so there is no new DoS surface.
- `logic`: Existing callers (the `bytes.Equal` checks at `baseProcess.go:280/283` and the root hash at `data/block/block.go:300`) compare the digest's value, not slice identity, so returning a copy is safe. The cache is still filled with `getHasher().Sum(nil)`, so the output is unchanged.
- `error-handling`: If init fails, nothing is cached. The panic fires before `emptyHash` is assigned, the deferred unlock runs, and the next call can't deadlock. `TestBlake2b_InvalidHashSizePanicsOnEveryCall` covers this.
- `resource`: Each empty-input call now takes an uncontended lock and allocates 32 bytes (16 for BLS). The call sites are per-block root-hash checks and occasional empty bloom keys, not hot loops, so the cost is negligible.
- `secrets`: The panic message includes only `HashSize` and the blake2b error text, no input data or key material.
- `crypto`: No primitive or parameter changed. It still uses `golang.org/x/crypto/blake2b` `New256`/`New` with a nil key.
- `deps`: `go.mod`/`go.sum` are unchanged. The only new imports are the stdlib `fmt` and `sync`.
- `regression`: No check was removed. `testComputeOk` uses an addressable variable and makes no copy. The `assert.Equal` in `hasherFactory_test` compares zero-valued structs and passes.
- `injection` / `authz`: Not touched; this diff only changes in-memory hashing.

</details>

Reviewed at `739be80`.

crypto/hashing/blake2b/blake2b_test.go
```diff
@@ -44,3 +45,79 @@ func TestBlake2b_Empty(t *testing.T) {
...
+	assert.Panics(t, func() { _ = hasher.EmptyHash() })
+	assert.Panics(t, func() { _ = hasher.EmptyHash() })
+	assert.Panics(t, func() { _ = hasher.Compute("") })
+}
+
+// The shared suite's testConcurrentEmptyHash calls EmptyHash() once before spawning its
+// goroutines, so the cache is already warm and the lazy initialization itself never races.
+// That is why it never caught this bug. This test leaves the instance cold so the goroutines
+// race the initialization, and mixes both entry points into it.
+func TestBlake2b_ConcurrentEmptyInputIsRaceFree(t *testing.T) {
```
RomuloSiebra, line 80: **[F1 — Test gap, non-blocking]**

This test only fails under `-race`. CI runs `go test` without it (`pr-qa-sec.yaml`), and without the detector every goroutine writes the same 32 bytes, so a torn header never shows up on the fast path. Running this test file against the pre-fix `blake2b.go` without `-race` passed 500/500; with `-race` it failed every run. The two `ReturnsCopy` tests do fail deterministically against the old code, so the aliasing half is guarded — it's only the synchronization half that has no CI guard.

Two cheap ways to make it fail on its own: assert that no two goroutines received slices sharing a backing array (the old code handed out the same field slice, so `&results[i][0] == &results[j][0]` fails deterministically), or add a `-race` job for `./crypto/hashing/...` in CI. Fine as a follow-up; the fix itself is correct.

RomuloSiebra (dismissed):

crypto/hashing/blake2b/blake2b.go
fbsobreira: Thanks — agreed it would be correct and simpler. Keeping the cache for consistency with sha256/keccak/fnv, which all cache their empty digest; the mutex-vs-`sync.Once` rationale is documented on the field.

crypto/hashing/blake2b/blake2b_test.go
fbsobreira: Good catch — added in 3170642 a pairwise check that no two goroutines' results share a backing array. Against the pre-fix `blake2b.go` it now fails 20/20 without `-race`. A `-race` job for `./crypto/hashing/...` is still worth doing separately.

fbsobreira: Follow-up: took the CI option too in 9a480dee. The QA test job now runs `go test -race -short` over every package except the six wasmer2 VM-execution packages, which trip `checkptr` under `-race`. The first full run found two test-only races in `tools/tracing` (a plain `int` counter shared with an httptest handler), which are fixed in the same commit.

fbsobreira: Correction: backed the `-race` CI step out of this PR in 6b991f5. Each CI run found another pre-existing issue (a libp2p shutdown crash, two packages that can't run under `-race` yet, and more), so enabling it is now tracked in KLC-2697. The deterministic backing-array check from 3170642 stays, so this test no longer depends on `-race` to catch the regression.

PR #148, opened 2026-09-18
[KLC-2421] appStatusPolling: join the polling goroutine in Close()

fbsobreira:
## Summary

Follow-up to KLC-2402. `AppStatusPolling.Close()` closed the `done` channel and returned immediately, so a polling handler caught mid-tick could still be reading from `NetMessenger` while `closeAllComponents` moved on to tear it down.

The drain-first comment in `cmd/node/startup.go` already claimed the stronger guarantee:

> Drain background goroutines first so their handlers cannot fire against components being torn down below.

This makes that true, and lines up with the synchronous `Close()` that `peerHonesty` already has in the same `backgroundClosers` slice.

## Why the join is bounded

`closeAllComponents` drains these closers **before** `Store.CloseAll` and the trie persisters, all under one `maxTimeToClose` watchdog. An unbounded wait on a stuck handler could let that watchdog fire and skip the storage close entirely — a worse outcome than the shutdown race the join exists to prevent.

So `Close()` waits up to `closeJoinTimeout` (1s, ~1000x normal handler cost), then returns `ErrCloseTimeout` and gives up. `closeAllComponents` already surfaces that via `LogIfError` and proceeds to close storage.

The bound is not theoretical — not every handler is a cheap in-memory read. `registerMemStatistics` calls `runtime.ReadMemStats` (stop-the-world) plus several gopsutil `/proc` lookups inside the handler, and `GetConnectedPeersInfo` walks every connected peer taking swarm locks. Only the cpu/net/disk handlers are pure atomic reads; their expensive compute runs in separate goroutines.

## WaitGroup misuse guard

`Poll()` and `Close()` now coordinate under a mutex, and `Poll()` after `Close()` is a no-op. Without this, `wg.Go`'s `Add` can land inside a live `wg.Wait`, which `sync` answers with a process-killing panic:

```
panic: sync: WaitGroup is reused before previous Wait has returned
```

Reproduced under `-race` before the guard was added. `closeOnce` was dropped — the `closed` bool under the same mutex already provides idempotency, and two mechanisms for one invariant is worse than one.

## Behaviour notes

- `Close()` is still idempotent, but no longer *always* returns nil — it returns `ErrCloseTimeout` when it gives up. The only production caller already logs the error.
- `Close()` must not be called from inside a registered polling handler; that would deadlock on the join. Documented on the method.

## Testing

- `go test -race -count=3 ./core/appStatusPolling/ ./cmd/node/metrics/` — pass
- `go test -race -count=1 ./cmd/node/... ./core/process/rating/peerHonesty/` — pass
- `go build ./...`, `go vet ./core/appStatusPolling/ ./cmd/node/...` — clean

New coverage: in-flight join, stuck-handler timeout, `Close` before `Poll`, `Poll` after `Close`, and `Poll` racing `Close` (200 iterations).

Negative controls confirm the tests discriminate:
- join removed → `TestAppStatusPolling_Close_WaitsForInFlightHandler` fails
- guard removed → the WaitGroup misuse panic reproduces

## Summary

- Synchronizes `Poll()` and `Close()` with `mutClose`.
- Prevents new polling goroutines after closure.
- Makes `Close()` wait up to one second for active handlers.
- Returns `ErrCloseTimeout` when a handler remains blocked.
- Preserves idempotent `Close()` behavior.

## Impact

- Improves shutdown stability for handlers that read `NetMessenger`, `Store`, and trie persisters.
- Reduces `WaitGroup` misuse and shutdown data races.
- A timeout allows storage shutdown to continue, but a stuck handler may still access components during teardown.
- Consensus, transaction processing, state management, and KVM logic are unchanged.
- No direct data-integrity change is introduced.

## Tests

- Covers in-flight joins, timeout handling, close-before-poll, poll-after-close, idempotent close, and concurrent `Poll()`/`Close()`.
- Race tests, builds, and vet checks pass.


core/appStatusPolling/appStatusPolling_test.go
```diff
@@ -118,12 +119,154 @@ func TestAppStatusPolling_Close_StopsGoroutine(t *testing.T) {
...
+
+	// Close must give up rather than block the caller forever: in production
+	// closeAllComponents still has Store.CloseAll and the trie persisters to
+	// run under a shared watchdog.
+	start := time.Now()
+	err = asp.Close()
+	elapsed := time.Since(start)
+
+	assert.ErrorIs(t, err, appStatusPolling.ErrCloseTimeout)
+	assert.Less(t, elapsed, 5*time.Second, "Close should have given up near closeJoinTimeout")
```
fbsobreira: **[suggestion]** `TestAppStatusPolling_Close_GivesUpOnStuckHandler` does not lock the "gives up near closeJoinTimeout" claim. `release` stays open until `t.Cleanup`, so any `ErrCloseTimeout` returned while the handler is still blocked satisfies `ErrorIs`, and `assert.Less(elapsed, 5s)` also passes if `Close` returns that error immediately. A regression that skips the wait would stay green. The upper bound also never stops a hang; `Close` is called on the test goroutine, so a missing timeout blocks until the package timeout.

**Suggestion:** Assert a lower bound as well (for example `elapsed` greater than a few hundred milliseconds) so an instant `ErrCloseTimeout` fails. Run `Close` on a side goroutine with its own deadline if the test should fail fast when the join is unbounded.

core/appStatusPolling/appStatusPolling.go
```diff
@@ -65,13 +85,48 @@ func (asp *AppStatusPolling) Poll() {
...
+// exists to prevent. Not every handler is a cheap in-memory read:
+// registerMemStatistics calls runtime.ReadMemStats plus several gopsutil /proc
+// lookups, any of which can stall on an IO-loaded node.
+//
+// On timeout Close returns ErrCloseTimeout and gives up rather than blocking;
+// closeAllComponents logs it via LogIfError and proceeds to close storage. The
+// abandoned waiter goroutine exits on its own once the handler finally returns.
+//
+// Idempotent. Must not be called from inside a registered polling handler —
+// that would deadlock on the join.
```
fbsobreira: **[nit]** The godoc says a `Close` from inside a registered handler "would deadlock on the join." With `closeJoinTimeout`, that call returns `ErrCloseTimeout` after one second and the handler continues; the outer `Wait` then completes. It stalls and burns the shutdown budget, but it does not deadlock. The same comment says the handlers read Store and the trie persisters (line 95). The registered ticks read `NetMessenger`, `ReadMemStats`/gopsutil, and in-memory snapshots (`ForkDetector`, redundancy, NTP). They do not touch the store or trie persisters; those closes matter only because they share `maxTimeToClose` and run after the drain.

**Suggestion:** Say re-entrant `Close` returns `ErrCloseTimeout` after `closeJoinTimeout` instead of deadlocking, and name `NetMessenger` (not Store or the tries) as what the in-flight handler can still be reading.

core/appStatusPolling/appStatusPolling.go
```diff
@@ -65,13 +85,48 @@ func (asp *AppStatusPolling) Poll() {
...
-	})
-	return nil
+	}
+	asp.mutClose.Unlock()
+
+	joined := make(chan struct{})
+	go func() {
+		asp.wg.Wait()
+		close(joined)
+	}()
```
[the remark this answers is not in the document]
fbsobreira: Fixed. Shutdown calls Close once, so this was not leaking waiters on the node path. Close is idempotent and can return ErrCloseTimeout, so a retry would. The first Close now starts one waiter; later calls share that channel and still return ErrCloseTimeout after closeJoinTimeout.

core/appStatusPolling/appStatusPolling_test.go
```diff
@@ -118,12 +119,154 @@ func TestAppStatusPolling_Close_StopsGoroutine(t *testing.T) {
...
+	asp, err := appStatusPolling.NewAppStatusPolling(&statusHandler.NilStatusHandler{}, pollingDuration)
+	require.Nil(t, err)
+
+	var once sync.Once
+	err = asp.RegisterPollingFunc(func(_ core.AppStatusHandler) {
+		once.Do(func() {
+			close(handlerStarted)
+			// Hold the tick open long enough that a non-blocking Close would
+			// demonstrably return while this handler is still running.
+			time.Sleep(500 * time.Millisecond)
```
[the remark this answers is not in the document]
fbsobreira: Replaced the in-flight sleep with a release channel: the test fails if Close returns before the handler is released. Removed the settling sleep in TestAppStatusPolling_Close_StopsGoroutine. Once Close returns nil, the goroutine has already exited.

Left the sleep in TestAppStatusPolling_PollAfterCloseIsNoOp. The polling floor is 1s, and that test is an absence check. Dropping the wait would let a still-started goroutine pass.

core/appStatusPolling/appStatusPolling_test.go
```diff
@@ -118,12 +119,154 @@ func TestAppStatusPolling_Close_StopsGoroutine(t *testing.T) {
...
+		wg.Add(3)
+		for j := 0; j < 2; j++ {
+			go func() {
+				defer wg.Done()
+				asp.Poll()
+			}()
+		}
+		go func() {
+			defer wg.Done()
+			_ = asp.Close()
```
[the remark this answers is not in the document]
fbsobreira: Fixed. The concurrent test now requires Close to return nil. A timeout there would mean the goroutine missed done.

core/appStatusPolling/appStatusPolling.go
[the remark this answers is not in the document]
fbsobreira: Fixed. The first Close starts one waiter and stores its completion channel. Every later Close selects on that same channel and still returns ErrCloseTimeout while the handler is stuck.

core/appStatusPolling/appStatusPolling_test.go
fbsobreira: Fixed. Close runs on a side goroutine with its own deadline, and the test requires the wait to last at least 500ms, so an instant ErrCloseTimeout fails.

core/appStatusPolling/appStatusPolling.go
fbsobreira: Fixed. The comment now says a re-entrant Close returns ErrCloseTimeout after closeJoinTimeout, and that in-flight handlers read NetMessenger. The store and trie persisters are closed afterward because they share the shutdown watchdog.

core/appStatusPolling/appStatusPolling_test.go
```diff
@@ -117,13 +118,202 @@ func TestAppStatusPolling_Close_StopsGoroutine(t *testing.T) {
...
+	releaseFn()
+
+	err, elapsed = closeWithin(t, asp, 2*time.Second)
+	assert.NoError(t, err)
+	assert.Less(t, elapsed, time.Second, "Close after the handler finished should return promptly")
+}
+
+func TestAppStatusPolling_PollAfterCloseIsNoOp(t *testing.T) {
+	t.Parallel()
+
```
Beroni, line 247: **low · test-validity** (exploit-digger)

Some tests can't fail when the change they cover is removed, and one assertion was dropped.

- **`PollAfterCloseIsNoOp` still passes if you delete the `closed` guard in `Poll`.** Without the guard, the new goroutine sees `done` already closed and returns before any tick. `wg.Add` after a completed `Wait` is also legal, so nothing panics.
- **`ConcurrentPollAndClose` only catches the panic in a narrow window.** The window is when the counter is 0 and the waiter is parked.
- **The check that no more handlers fire after a successful `Close` was removed** from `Close_StopsGoroutine`. The join now makes that check deterministic, so it can come back.
- **Nothing tests the extra-batch-after-`Close` finding** (see review body).

**Fix:**
- Make the guard test assert on something the guard controls: goroutine count via `goleak` or `runtime.NumGoroutine`, or a test hook.
- Restore the no-more-calls assertion after `Close`.
- Add a test where a handler runs longer than `pollingDuration`, then assert that no second batch runs after `Close`.

core/appStatusPolling/appStatusPolling.go
```diff
@@ -65,13 +86,53 @@ func (asp *AppStatusPolling) Poll() {
 				asp.mutRegisteredFunc.RUnlock()
 			}
 		}
-	}()
+	})
 }
 
-// Close stops the polling goroutine. Idempotent; always returns nil.
+// Close stops the polling goroutine and waits for it to exit, so that a handler
```
Beroni: **low · docs** (exploit-digger, checked against `cmd/node/startup.go:1162-1200`)

The `Close` doc comment and the PR description don't match the code.

1. **Wrong shutdown order in the doc comment.** It says `closeAllComponents` closes "NetMessenger … and then the Store and trie persisters". The real order is: background closers, health service, `Store.CloseAll`, trie persisters, and **`NetMessenger` last**.
2. **"Would deadlock" is wrong in the PR description.** It says calling `Close` from inside a handler would deadlock. It doesn't: the handler holds only `mutRegisteredFunc.RLock`, `Close` takes `mutClose`, and after 1s `Close` returns `ErrCloseTimeout`. The code comment gets this right; the description is stale after the second commit.
3. **"The only production caller" is misleading.** The shutdown loop closes four `AppStatusPolling` instances one after another, so the worst case is about **4s** (4 × `closeJoinTimeout`) before `Store.CloseAll`, not 1s.

**Fix:** correct the order in the doc comment, and update the PR description (no deadlock, four instances, ~4s worst case).

---

**simplification** (lazy-senior-dev) · ~20 lines: the `Close` doc comment is 23 lines, and the `closeJoinTimeout` comment is 5 more. Cut to 4–5 lines and move the reasoning to the PR description (if you keep it, fix the ordering first, see above).

core/appStatusPolling/appStatusPolling_test.go
```diff
@@ -117,13 +118,202 @@ func TestAppStatusPolling_Close_StopsGoroutine(t *testing.T) {
...
+	go func() {
+		errCh <- asp.Close()
+	}()
+
+	select {
+	case err = <-errCh:
+		t.Fatalf("Close returned while the handler was still running: %v", err)
+	case <-time.After(200 * time.Millisecond):
+	}
+
```
Beroni, line 156: **low · test-flakiness** (exploit-digger)

Tests rely on wall-clock margins against a hard-coded 1s timeout.

- **`WaitsForInFlightHandler`** releases the handler 200ms in, which leaves about 800ms of margin before `Close` gives up.
- **`ConcurrentPollAndClose`** calls `require.NoError(Close())` 200 times, each with a 1s budget.
- **`PollAfterCloseIsNoOp`** always sleeps 2s.

On a loaded CI runner under `-race`, these could fail intermittently. They passed locally: `-race -count=3` took about 10s.

**Fix:** turn `closeJoinTimeout` into a variable tests can override (e.g. via `export_test.go`). Use a long timeout in the tests that expect the join to succeed and a short one in the timeout test.

core/appStatusPolling/appStatusPolling_test.go
```diff
@@ -117,13 +118,202 @@ func TestAppStatusPolling_Close_StopsGoroutine(t *testing.T) {
 	}, 3*pollingDuration, 50*time.Millisecond, "expected goroutine to fire at least once before Close")
 
 	assert.NoError(t, asp.Close())
+}
+
+func TestAppStatusPolling_Close_WaitsForInFlightHandler(t *testing.T) {
```
Beroni, line 134: **simplification** (lazy-senior-dev) · ~30 lines

Same `release`/`releaseOnce`/`handlerStarted` setup in two tests. Two tests also rebuild the close-with-deadline check inline.

**Fix:** one `blockOnFirstTick(t, asp) (started, release func())` helper, and reuse the `closeWithin` helper that already exists in the file.

core/appStatusPolling/appStatusPolling.go
```diff
@@ -10,14 +10,23 @@ import (
...
 // AppStatusPolling will update an AppStatusHandler by polling components at a predefined interval
 type AppStatusPolling struct {
 	pollingDuration     time.Duration
 	mutRegisteredFunc   sync.RWMutex
 	registeredFunctions []func(appStatusHandler core.AppStatusHandler)
 	appStatusHandler    core.AppStatusHandler
 	done                chan struct{}
-	closeOnce           sync.Once
+	wg                  sync.WaitGroup
+	mutClose            sync.Mutex
```
Beroni, line 24: **simplification** (lazy-senior-dev) · ~14 lines

`WaitGroup` + `joined` channel + waiter goroutine, all to join one goroutine (all four callers call `Poll` once).

**Fix:** one `stopped chan struct{}` closed by the goroutine. Under the lock, `Poll` returns early if `closed` or already started. `Close` selects on `stopped` against a timeout. This also removes the WaitGroup panic risk and its 4-line comment. The `mutClose` + `closed` guard stays.

**Trade-off:** a second `Poll` becomes a no-op. Author's call.

Beroni (commented):
## Automated review — Bazel

**Verdict: 💬 comment** · lenses: `senior-code-reviewer` (approve), `exploit-digger` (comment), `lazy-senior-dev` (comment)

The core change is sound. `wg.Go` (the `WaitGroup` Add) only runs under `mutClose` while `closed == false`. `Close` sets `closed` before `wg.Wait` starts, so the WaitGroup misuse panic is fixed. Repeated `Close` calls share one waiter. The shutdown caller already logs the new error. One real logic gap remains on the timeout path, plus some doc and test issues. None of them block the merge.

### 🟠 Medium — Polling loop can run one more handler batch after `Close` closes `done`
`core/appStatusPolling/appStatusPolling.go:78` · *exploit-digger* · (line is outside the diff, so it is not anchored inline)

The loop's `select` gives `<-asp.done` and `<-ticker.C` equal weight, and Go picks at random when both are ready. If a handler batch takes longer than `pollingDuration`, a tick is always waiting when the batch returns. After `Close` has closed `done`, the goroutine then has about a 50% chance of running one more **full** batch.

- **Happy path:** this only makes the join slower.
- **Timeout path, which is the one that matters:**
  1. `Close` returns `ErrCloseTimeout`.
  2. `closeAllComponents` moves on to `Store.CloseAll`, then the trie persisters, then `NetMessenger.Close`.
  3. The slow goroutine can start a **new** batch against components that are closing or already closed. That is the race this PR says it removes.
- **Trigger:** `machineStats` polls every 1s, so any `ReadMemStats` or `/proc` stall longer than 1s during SIGTERM sets this up.
- **Doc impact:** the doc line "the waiter exits once the handler returns" does not hold in this case.

**Fix:** make `done` win. Right after `case <-ticker.C:`, before taking the RLock, add a non-blocking check:
```go
select { case <-asp.done: return; default: }
```
Optionally, also check `done` between handlers inside the batch loop.

### Needs human verification (`exploit-digger`)

- **`.github/workflows/tests-dispatch.yaml:40`**
  - **Concern:** this is the repo's first use of `sync.WaitGroup.Go`, which needs Go 1.25. `go.mod` (1.25.7) and the Dockerfiles are fine, but `e2e-tests.yaml` defaults `GO_VERSION` to `1.22` and the runner uses `GOTOOLCHAIN=local`.
  - **To settle:** confirm that `vars.GO_VERSION` and the `go-setup-action` default resolve to Go 1.25 or later.

<details>
<summary>Coverage (checked and found clean)</summary>

- **Deadlocks:** `Close` and the handlers never take `mutClose` and `mutRegisteredFunc` in opposite orders. `Poll` holds `mutClose` only while starting the goroutine. `RegisterPollingFunc` called from inside a handler would deadlock, but that was already true before this PR.
- **WaitGroup:** every Add happens under the lock while `closed == false`, so no Add can overlap a `Wait`. Multiple `Poll` calls all exit on the shared `done`. `Close` before `Poll` returns nil immediately. Repeated `Close` calls share one waiter.
- **Resources:** the timer and ticker are both stopped. At most one waiter goroutine exists, and it only outlives `Close` while a handler is stuck.
- **Error contract:** production code only reaches `Close` through `io.Closer` plus `LogIfError`. Test callers discard the error. Nothing depended on `Close` always returning nil.
- **Regression:** `close(done)` still runs exactly once without `closeOnce`.
- **Deps and build:** `go.mod` and `go.sum` are unchanged. On the local go1.25.7, `go test -race -count=3` and `go vet` pass.
- **Not applicable:** injection, authz, input validation, secrets, crypto.
</details>

Inline: 3 low findings and 3 simplification cuts (~64 lines of potential savings in total).

Reviewed at `0f163b6`.

core/appStatusPolling/appStatusPolling.go
fbsobreira: Taken — done in `2245005b`.

`WaitGroup`, `joined` and the waiter goroutine are all gone. `Poll` publishes a `stopped` channel that the goroutine closes on exit, `Close` captures it under `mutClose` and selects against the timer. Close-before-Poll returns nil on the nil channel, repeated `Close` calls share the channel for free, and the WaitGroup-misuse comment went with the WaitGroup.

On the trade-off: took it. All four production callers `Poll()` once, and a second `Poll` only ever started a duplicate ticker doing duplicate work — so a no-op is the better behaviour, not a loss. `TestAppStatusPolling_SecondPollIsNoOp` pins it, and fails if the already-started guard is removed.

core/appStatusPolling/appStatusPolling.go
fbsobreira: All three correct — fixed in `2245005b`.

1. **Shutdown order.** The doc comment now states the real order: `closeAllComponents` drains the background closers, then `Store.CloseAll` and the trie persisters, and `NetMessenger` **last**.
2. **"Would deadlock".** Wrong, as you say — the handler holds only `mutRegisteredFunc.RLock`, `Close` takes `mutClose`, and the call returns `ErrCloseTimeout`. PR description corrected.
3. **"The only production caller".** Also corrected: four instances closed in sequence, so ~4s worst case before `Store.CloseAll`, not 1s.

The comment is down from 28 lines to 6; the reasoning now lives in the PR description.

core/appStatusPolling/appStatusPolling_test.go
fbsobreira: All four points taken — fixed in `2245005b`.

- **`PollAfterCloseIsNoOp` passed without the guard.** Correct, and a good catch. It now asserts on what the guard actually controls, through an `export_test.go` seam rather than a production hook: `assert.False(t, asp.PollStarted())`. Negative control: deleting the `closed` guard fails it. The 2s sleep is gone as a side effect.
- **No-more-calls assertion restored** in `Close_StopsGoroutine`, deterministic now that the join is real.
- **Extra-batch finding is covered** by the new `Close_RunsNoBatchAfterClose`. Detection is ~50% per attempt since `select` picks at random, so it runs 3 times. Negative control: removing the `done` re-check fails it with `expected: 1, actual: 2`.
- **`ConcurrentPollAndClose` narrow window** is moot now — the WaitGroup is gone, so there is no panic window left to hit. It guards the `Poll`/`Close` field race under `-race` instead, and its comment says so.

core/appStatusPolling/appStatusPolling_test.go
fbsobreira: Fixed in `2245005b`.

`closeJoinTimeout` is now a per-instance field with a `SetCloseJoinTimeout` setter in `export_test.go`. Per-instance rather than a package-level var specifically because these tests are `t.Parallel()` — a shared var would be a data race under `-race`.

Tests that expect the join to succeed use 30s, so a loaded runner cannot turn a pass into `ErrCloseTimeout`. `GivesUpOnStuckHandler` uses 200ms, which also cuts its runtime, with its margin assertions scaled to that. `PollAfterCloseIsNoOp` no longer sleeps 2s at all.

core/appStatusPolling/appStatusPolling_test.go
fbsobreira: Fixed in `2245005b`. One `blockOnFirstTick(t, asp)` helper returning `(started, release, calls)`, used by three tests now, and `Close_BeforePollDoesNotBlock` uses the existing `closeWithin` instead of rebuilding the deadline check inline.

core/appStatusPolling/appStatusPolling_test.go
```diff
@@ -117,13 +118,202 @@ func TestAppStatusPolling_Close_StopsGoroutine(t *testing.T) {
...
+				asp.Poll()
+			}()
+		}
+		closeErr := make(chan error, 1)
+		go func() {
+			defer wg.Done()
+			closeErr <- asp.Close()
+		}()
+		wg.Wait()
+		require.NoError(t, <-closeErr)
```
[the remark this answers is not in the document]
fbsobreira: Not taking this one, though the observation is right: `Close` can take `mutClose` first and both concurrent `Poll` calls then return at the guard.

The reason is cost. Making that ordering deterministic means putting a test-only seam inside production `Poll`, and the 200-iteration loop already provides the coverage the test exists for.

The underlying risk also changed in `2245005b`: the `WaitGroup` is gone. `Poll` now publishes a `stopped` channel and `Close` selects on it, so there is no `Add`-during-`Wait` to misuse — the panic window this test was written for no longer exists, and it now guards the `Poll`/`Close` field race under `-race`.

The `Poll` guards did need real coverage, and they have it deterministically now, via `export_test.go` seams instead of production hooks: `PollAfterCloseIsNoOp` asserts `PollStarted()` is false, `SecondPollIsNoOp` asserts no duplicate ticker. Both fail if their guard is removed.

fbsobreira:
@Beroni — the two review-body items that had no inline anchor, both addressed in `2245005b`.

## 🟠 Medium: extra handler batch after `Close` closes `done`

Confirmed and fixed. The tick case now re-checks `done` before taking the RLock:

```go
case <-ticker.C:
    // A batch slower than pollingDuration leaves a tick queued, so the
    // select above sees two ready cases and picks at random. Re-check
    // done here, or Close can be followed by one more batch.
    select {
    case <-asp.done:
        return
    default:
    }
```

Your reasoning held up exactly. A new test reproduces it — `TestAppStatusPolling_Close_RunsNoBatchAfterClose` holds a batch past a full period so a tick is queued, then closes and releases. With the re-check removed it fails on the first attempt:

```
Error: Not equal: expected: 1, actual: 2
Messages: a handler batch ran after Close closed done
```

Detection is ~50% per attempt because `select` picks at random, so the test runs the scenario 3 times.

I left the between-handlers check out — it would end a batch half-applied, and the batch-entry check already closes the window that matters on the timeout path.

## Needs human verification: Go version in CI

Checked, and it does resolve to `1.22`:

```
$ gh api repos/klever-io/klever-go/actions/variables
{"variables":[],"total_count":0}
```

`vars.GO_VERSION` is unset, so `e2e-tests.yaml:22` fell through to its `1.22` default while every other workflow defaults to `1.25.7`. It was already broken against `go.mod` (`go 1.25.7`) independently of `wg.Go`, and it is `workflow_dispatch`-only so it never gated this PR. Bumped it to `1.25.7` in this commit rather than leaving it stale.

`tests-dispatch.yaml` takes `go_version` as a required input, so it has no stale default to fix.

---

Also worth noting for the WaitGroup coverage in your "checked and found clean" section: the `stopped`-channel refactor from your inline comment landed, so the WaitGroup is gone entirely and the misuse risk with it.

RomuloSiebra (dismissed):

core/appStatusPolling/appStatusPolling_test.go
```diff
@@ -118,12 +120,250 @@ func TestAppStatusPolling_Close_StopsGoroutine(t *testing.T) {
...
+
+	err = asp.RegisterPollingFunc(func(appStatusHandler core.AppStatusHandler) {
+		appStatusHandler.SetInt64Value(core.MetricNumConnectedPeers, int64(10))
+	})
+	require.Nil(t, err)
+
+	// A second goroutine would double every metric update at the same interval.
+	asp.Poll()
+	asp.Poll()
+	t.Cleanup(func() { _ = asp.Close() })
```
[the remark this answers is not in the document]
fbsobreira: Good catch — fixed in `ebc46f52`, with one deliberate change to the suggested patch.

**Applied, using `assert` rather than `require`.** `require.NoError` calls `t.FailNow()`, which is `runtime.Goexit()`. Inside a `t.Cleanup` callback that aborts the cleanup function and skips whatever follows it, so a genuine `ErrCloseTimeout` could end up masking a second failure rather than reporting it cleanly. `assert.NoError` reports the same thing without unwinding:

```go
t.Cleanup(func() { assert.NoError(t, asp.Close()) })
```

Applied to the other discarding cleanup in the file too (`Poll_TestNumOfConnectedAddressesCalled`), which also now sets a long join bound so the new assertion cannot flake on a loaded runner.

---

While here, the **hardcoded sleeps** from the Additional comments section (`:126`, `:226`, `:282`). The repo rule in `.coderabbit.yaml:131` is fair, and two of the three are now gone. A `StoppedChan()` seam in `export_test.go` lets both tests assert the goroutine lifecycle directly instead of sampling over a wall-clock window:

- `Close_StopsGoroutine` — asserts the `stopped` channel is already closed when `Close` returns. That *proves* no handler can fire afterwards, where the old 2s sleep only sampled for it.
- `SecondPollIsNoOp` — compares the channel identity across the second `Poll`, so a dropped already-started guard is caught immediately.

Both still discriminate, and now fail fast: 0.14s and 1.0s respectively with their guard removed, versus multi-second before.

The third (`:226`, in `Close_RunsNoBatchAfterClose`) stays. It has to let a real `time.Ticker` queue a tick behind the held batch, which is the precondition the test exists to create — removing it means injecting a clock into production code purely for test determinism, which is not a good trade here. Note the package wall time is unchanged at ~21.9s either way, since it is bounded by that test running in parallel.

---

On the **zizmor `excessive-permissions`** warning for `e2e-tests.yaml`: real, but pre-existing and repo-wide — `go-setup-lint`, `push`, `release-docker` and `tests-dispatch` all lack a `permissions:` block too, and this PR only touched line 22 to correct a stale Go version. Fixing one of five here would leave the inconsistency in place while widening a polling PR into workflow hardening, so it is better handled as its own change.

core/appStatusPolling/appStatusPolling_test.go
```diff
@@ -104,26 +106,259 @@ func TestAppStatusPolling_Close_StopsGoroutine(t *testing.T) {
...
-	// Close does not synchronously join the goroutine, so a tick already
-	// selected may complete after Close returns. Wait long enough that the
-	// goroutine has observed `done`, then assert the count is stable across a
-	// second polling interval.
-	time.Sleep(2 * pollingDuration)
-	stable := atomic.LoadInt32(&callCount)
-	time.Sleep(2 * pollingDuration)
-	assert.Equal(t, stable, atomic.LoadInt32(&callCount), "expected no further handler invocations after Close settles")
+	// Nothing can fire after Close returned because the goroutine is already
+	// gone. Assert that directly rather than sampling the count over a sleep.
```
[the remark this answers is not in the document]
fbsobreira: The premise is right, but it does not lead to the hazard this PR removes, so I am not changing `AppStatusFacade` here.

**Where you are right:** every facade setter is fire-and-forget (`statusHandler/appStatusFacade.go:60-66`), and production polling does go through the facade. A facade goroutine can still be running when `AppStatusPolling.Close` returns. The assertion at `:126` proves only that the polling goroutine exited, and that is the intent.

**Why it does not reintroduce the race:**

1. **The component reads are already joined.** The shutdown hazard is a handler reading `NetMessenger`, the store, or `/proc` while `closeAllComponents` tears them down. Those reads (`GetConnectedPeersInfo`, `ReadMemStats`, the gopsutil lookups, …) run synchronously inside the polling goroutine, which `Close` now waits for. The facade goroutine only receives the finished value.
2. **The value only reaches sinks that shutdown never closes.** The production facade (`factory/structsStatusHandlers.go:138`) wraps `statusMetrics`, `persistentHandler`, and optionally the TUI presenter. For every setter the polling handlers use, each of these just stores into memory (a `sync.Map` or presenter state), and `closeAllComponents` closes none of them.
3. **The only storage write is not reachable from polling.** `persistentHandler` does write to the `StatusMetricsUnit` storer, which `Store.CloseAll` closes. It does so only on `SetUInt64Value(core.MetricNonce, …)` (`statusHandler/persister/persistentHandler.go:159-173`), which fires on block commit. No polling handler in `cmd/node/metrics/` sets `MetricNonce`; I checked every setter they call.

So after `Close` returns, a lingering facade goroutine can do at most an in-memory map write on an object that is still alive. Making the facade lifecycle-aware would change a type shared by block processing, consensus and every other status writer, to remove a hazard that does not exist on this path. The trade is not worth it inside this PR.

RomuloSiebra (approved):

PR #150, opened 2026-09-18
[KLC-2596] Add panic barriers to the WebSocket client goroutines

fbsobreira:
## Summary

`gin.Recovery()` only covers the gin handler goroutine. A goroutine detached from the request is outside it, so a panic there terminates the node process rather than failing the one connection — the same mechanism as GHSA-g797-9c85-644m, which panicked in the allowance fan-out goroutine.

The API layer already closes this with `shared.SafeGo`, and the allowance fan-out got its own barrier in KLC-2586. The three goroutines a WebSocket **client** owns had none.

## Changes

| Site | Barrier | Teardown |
| --- | --- | --- |
| `websocket/client.go` per-request worker | added | answers the client with an opaque internal error, keeps the connection |
| `websocket/client.go` `loopIn` | added | registered before the existing teardown defer so it unwinds last, also covering a panic raised inside that teardown |
| `websocket/client.go` `loopOut` | added | closes — it owns the only writer, and a half-live client has nothing behind it |
| `network/api/shared/safego.go` | already existed | gains the stack it was missing, plus a scrub of the panic value |

## Notes for review

**Recovering is safe on the request path.** Both mutating hub calls reachable from `HandleClientRequest` — `HandleClientInsertion` and `HandleClientRemoval` — release `h.mu` with `defer`, so the mutex unwinds rather than staying locked. The non-`defer` unlocks in this file are all on the event-broadcast path, which a client request cannot reach.

**Panic logging is bounded so the barrier cannot become a new lever.** Every recovered panic logs at Error and none is folded away — a panic is a node defect, not ordinary peer behaviour. What is bounded is the stack, the only multi-kilobyte field in the package: one per window, and **per barrier** rather than per hub, so a flood on one path cannot spend the stack budget a rarer defect on another path needs. Panic values go through the same bound-and-scrub as errors (`loggablePanic`), because a panic message routinely quotes what it was handed, and a value of `"\nERROR ..."` would otherwise forge a log line.

**`stacksOmitted` is deliberately not the sibling of `similarSinceLastLog`.** At the other `dropWarner` call sites a folded occurrence produces no line at all, so the count is what the operator never saw. Here every occurrence still logs and only its stack is dropped, so the count is reported one lower — this panic's own stack is attached, not omitted.

## Testing

- `go build ./...`, `gofmt`, `go vet` — clean.
- `go test ./network/... ./websocket/` — no failures; `-race` green on every affected package.
- Each new barrier was **mutation-checked**: disabling it in turn makes the matching test crash the test binary with an unrecovered panic, so these are real regression tests rather than vacuous ones.
- The per-barrier budget test was checked the same way — collapsing the selector back to a single shared warner fails it on all three paths.

One existing assertion in `safego_test.go` changed: it pinned the raw panic value, which is now quoted. Updated, and a case covering the log-forging vector was added alongside it.

## Scope

Per `SECURITY.md`, this is the **hardening** bucket — defence in depth, no advisory. The sweep behind it found no currently-exploitable panic on these paths; the exposure being closed is structural, so that every latent panic in the transaction and block read paths is a `500` rather than a node stop.

The hub's **own** detached goroutines (`notifyTxSender` / `notifyTxReceipts` / `postTxHash`, the mirror post worker, and the `StartServer` event loop) are the same class and are deliberately **not** in this PR: they are indexer-driven rather than peer-reachable. Tracked separately in KLC-2682.

## Summary

- Adds panic barriers to WebSocket request workers, `loopIn`, and `loopOut`.
- Keeps recovered request panics isolated to one request and preserves the connection.
- Releases request worker slots after recovery.
- Preserves read-loop teardown and closes clients after write-loop panics.
- Bounds and sanitizes panic reporting in the WebSocket hub to limit log growth and prevent log injection.
- Keeps detached goroutine recovery at Error level. `SafeRun` logs a full stack and closes the connection after recovery.
- Affects networking, concurrency, error handling, and node stability.
- Does not change consensus, transaction processing, state management, KVM behavior, or blockchain data integrity.
- Adds regression tests for recovery, teardown, worker-slot release, stack budgets, and log sanitization.


websocket/panic_barrier_test.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	require.Eventually(t, func() bool { return len(c.sem) == 0 }, 3*time.Second, 20*time.Millisecond,
+		"every recovered request must hand its worker slot back; a leak here wedges the connection")
+}
+
+// TestPanicBarrier_LoopIn_KeepsNodeAlive drives the read loop into a panic and asserts it
+// unwinds to a clean return. newTestClient leaves conn nil, so loopIn's first conn call
+// panics — a stand-in for any latent panic on the read path. killClient runs first so the
+// teardown defer's c.Close() is a no-op on that nil conn, leaving the barrier itself, and
```
Beroni: **minor · test-validity** (senior-code-reviewer)

**The loopIn barrier test doesn't check defer order or teardown.**

`TestPanicBarrier_LoopIn_KeepsNodeAlive` only checks that `loopIn` returns. The test would still pass if the two defers in `loopIn` were swapped: the barrier would recover first, the teardown would run afterwards, and `loopIn` would return either way. So the ordering its comment says it covers is not tested, and the PR's "mutation-checked" claim does not hold for this barrier.

It also never checks that teardown ran:
- The client is never inserted into the hub.
- `killClient` turns `c.Close()` into a no-op.

A barrier that skipped `Close`/`handleClientDelete` would pass too. The property that matters is that a reader panic, once recovered, still closes and deregisters the client. Nothing asserts that, and nothing covers a panic raised inside the teardown itself (exploit-digger's test-gap note says the same).

**Fix:**
1. Use `dialLiveClient` and register the client with `HandleClientInsertion`.
2. Force a panic on the read path, for example with a pong handler that panics.
3. Assert `require.Eventually(!c.IsAlive())`, that `c.Done()` is closed, and that the client is gone from `hub.clients` and the subscription maps.
4. To pin the ordering, make the first teardown step panic and assert that the barrier recovers and logs it.

websocket/websocket.go
```diff
@@ -627,6 +663,61 @@ func loggableError(err error) string {
...
+// saves. What is bounded is the stack — the only multi-kilobyte field in this package. If a
+// latent panic ever does become peer-triggerable, a stack per request would turn this
+// barrier into the log-amplification lever the budgets above exist to deny, so the stack
+// rides the first panic of each window on that barrier.
+//
+// stacksOmitted is therefore not the sibling of similarSinceLastLog above: there a folded
+// occurrence produced no line at all, so the count is what the operator never saw. Here
+// every occurrence produced a line and only its stack was dropped, so the count is reported
+// one lower — this panic's own stack is attached, not omitted.
+func (h *SocketHub) logRecoveredPanic(op string, r interface{}) {
```
Beroni, line 683: **nit · shrink** (lazy-senior-dev) · ~30 lines saved

**Comments are longer than the code they explain.**

- The doc on `logRecoveredPanic` is 14 lines, including a paragraph on why `stacksOmitted` is `count-1`.
- The comment on the `SafeRun` recover is 12 lines for a 4-line log call.
- The comment on the request worker (`client.go:185`) is 7 lines about defer order.
- The `panicWarner` doc and the note on the final-flush entries add more.

**Fix:** keep about two lines per site: why the barrier exists (goroutines outside `gin.Recovery`, KLC-2596) and why the panic value is scrubbed. Leave the rest to the tests and the ticket.

websocket/websocket.go
```diff
@@ -627,6 +663,61 @@ func loggableError(err error) string {
 	}, text)
 }
 
+// loggablePanic renders a recovered panic value. A panic value is not trustworthy text just
+// because the node produced it: the message routinely quotes what it was handed, so peer
+// input reaches the log through it exactly as it does through an error, and a value of
+// "\nERROR ..." would forge a log line. It gets the same bound and scrub as loggableError.
+func loggablePanic(r interface{}) string {
```
Beroni, line 659: **nit · shrink** (lazy-senior-dev) · ~8 lines saved

**`loggablePanic` is a one-caller wrapper with its own copy of the limit.**

`loggablePanic` is a 4-line wrapper with its own `256` constant, and `logRecoveredPanic` is its only caller. The same bound already exists as `maxLoggableErrorLength`, and `shared.QuoteForLog` also uses `%.256q`.

**Fix:** call `loggableText(fmt.Sprint(r), maxLoggableErrorLength)` inline in `logRecoveredPanic`, and point the test at `loggableText`.

Beroni (commented):
## Automated review — Bazel (review-fleet)

**Verdict:** comment · 1 finding (1 minor) · 2 cuts (−38 lines)
**Scope:** `git diff origin/develop...HEAD`: 5 files, 1 commit, head `b427ef7`
**Lenses:** senior-code-reviewer, exploit-digger, lazy-senior-dev (none skipped). All three returned `comment`, and none found a blocking defect.

**Orchestrator checks:** `go test -race ./websocket/ ./network/api/shared/` passes. `go vet` and `gofmt -l` report nothing on the changed packages. Two of the PR's claims were checked against the code and both hold:
- `h.mu` is released with `defer` in `HandleClientInsertion`, `HandleClientRemoval` and `handleClientDelete`.
- `dropWarner.fire()` returns a count that includes the current occurrence, so `stacksOmitted = count-1` is correct.

The finding and the two cuts are posted inline.

> **Dropped cut (lens conflict):** lazy-senior-dev proposed collapsing the four per-barrier stack budgets, the `panicWarner` switch and the `other` fallback into one budget for the request path, with loopIn/loopOut left unbounded (−45 lines). The argument rests on the same premise as `SafeRun`'s comment: repeating a connection-ending panic costs a reconnect against the connection caps. exploit-digger checked that premise and found it does not hold: `ConnLimiter` caps concurrent connections, not the connection rate. Since the budgets bound log amplification, the code stays and the cut is dropped. The author may still simplify it once the rate-limit question is settled.

<details>
<summary>Coverage — swept clean</summary>

- **Injection:** op names are package constants. Panic values are bounded and scrubbed, by `loggableText` in the hub (256 bytes, control characters, U+2028/U+2029) and by `QuoteForLog` in `SafeRun`. Stack dumps contain no peer bytes.
- **Authz:** no auth logic changed. `/subscribe` and `/log` keep the same auth handler chain.
- **Validation:** request parsing, `ValidateSubscription` and event-type parsing are unchanged.
- **Secrets / info leak:** the peer only ever gets the opaque constant `errInternal`. The panic value and stack go only to the node log.
- **Logic:** `recoverPanic` is the deferred function itself, so `recover()` works. The defer order is as the PR describes. `panic(nil)` still triggers recovery on Go 1.25, which turns it into a non-nil `*runtime.PanicNilError`. `runtime.Goexit` is not swallowed.
- **Error handling:** after a recovered panic, `c.send` takes `aliveLock`, returns early when the client is dead, and never blocks, so it cannot hit a closed channel. A partial subscription is cleared by `handleClientDelete` at teardown.
- **Resources:** the worker slot is released on every path, and a test covers it. The hub stack is budgeted per barrier. `debug.Stack()` captures only the current goroutine.
- **Crypto / deps:** no cryptography touched. `go.mod` is unchanged; the only new import is the standard library's `runtime/debug`.
- **Regression:** `loggableError` behaves the same after the `loggableText` extraction. The `ws.loopIn` label string is unchanged. The new warners are added to the final flush.
- **PR/code match:** the code does what the description says, except for the `SafeRun` reconnect-cost argument (see above).

</details>

Reviewed at `b427ef7`.

network/api/shared/safego_test.go
```diff
@@ -84,7 +85,43 @@ func TestSafeGo_RecoversPanicLogsAndClosesConn(t *testing.T) {
 	select {
 	case args := <-logged:
 		require.Contains(t, args, "boom-routine")
-		require.Contains(t, args, "boom")
+		// Quoted, not raw: the panic value is rendered through QuoteForLog (KLC-2596).
+		require.Contains(t, args, `"boom"`)
+		require.Contains(t, args, "stack", "a recovered panic without its stack is not actionable")
```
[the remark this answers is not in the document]
fbsobreira: Fixed — the test reads the value paired with `stack` and requires a non-empty trace that contains a `SafeRun` frame.

network/api/shared/safego.go
```diff
@@ -17,7 +18,23 @@ import (
...
+			// through an error — and a value of "\nERROR ..." would otherwise forge a log
+			// line. The stack needs no such treatment: a goroutine dump carries function
+			// names and hex arguments, never peer bytes. It is also left unbounded, unlike
+			// the hub's in-package barrier: a panic here closes the connection, so repeating
+			// it costs a full reconnect against the per-IP and global connection caps rather
+			// than one more frame on an open socket.
+			log.Error("panic in detached websocket goroutine",
+				"goroutine", name,
+				"recover", QuoteForLog(fmt.Sprintf("%v", r)),
+				"stack", string(debug.Stack()),
```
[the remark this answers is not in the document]
fbsobreira: Keeping the full stack. `debug.Stack()` here is this goroutine, and its depth is the node's own call chain. Truncating it would drop the frames this line exists to record.

How often the line is written is separate: this path closes the connection, so it is one trace per connection. The hub budgets are for a panic per request on a socket that stays open. The comment now says that. `ConnLimiter` caps how many sockets are open at once, so the old wording overstated what those caps bound.

websocket/panic_barrier_test.go
fbsobreira: Fixed — `TestPanicBarrier_LoopIn_ClosesAndDeregisters` registers the client, panics from `Read` after the handshake, and requires the client closed, `Done()` closed, and removed from the hub and its subscription maps.

`TestPanicBarrier_LoopIn_RecoversPanicInTeardown` panics inside `Close`. The barrier defer is registered first, so it runs after that teardown. Swapping the two defers lets `Close`'s panic escape and crashes the test.

websocket/websocket.go
fbsobreira: Shortened. Each site now says why the barrier exists and why the panic value is scrubbed. `stacksOmitted` stays one line: it is `count-1` because that line already carries its own stack.

websocket/websocket.go
fbsobreira: Keeping `loggablePanic`. `TestLoggablePanic_ScrubsPeerInputFromTheLogLine` calls it, so the panic path stays covered on its own. `maxLoggableErrorLength` is local to `loggableError`, and `QuoteForLog` uses `%q` in another package.

websocket/client.go
```diff
@@ -100,13 +100,28 @@ func unexpectedClose(err error) bool {
...
+// recoverPanic is the barrier for client goroutines spawned outside gin.Recovery (KLC-2596).
+// teardown, when set, runs before the panic is logged.
+func (c *client) recoverPanic(op string, teardown func()) {
+	r := recover()
+	if r == nil {
+		return
+	}
+	if teardown != nil {
+		teardown()
+	}
```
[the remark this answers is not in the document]
fbsobreira: Fixed — the teardown now runs under its own recover.

You are right about the mechanism, and it is worth naming why it only bites this path. `loopIn` is safe already: its teardown is a separate defer registered *after* the barrier, so it runs *before* it and a panic there is recovered (`TestPanicBarrier_LoopIn_RecoversPanicInTeardown` pins that). The teardown passed *into* `recoverPanic` — `loopOut`'s `c.Close` and the request worker's error `send` — is the one that runs after `recover()` has consumed the original panic, so a panic in it has nothing left to catch it. Confirmed against the real barrier: reverting just this change turns the new test into `panic: ... [recovered] / panic: boom: simulated panic in the barrier's teardown` and kills the test binary.

`recoverPanic` now delegates to a `runTeardown` helper that recovers the teardown separately and logs it under `op+".teardown"`, then lets the original panic be logged as before. Both survive.

One note on the budget, since it is adjacent to the stack-amplification point raised elsewhere in this review: `op+".teardown"` is not a named case in `panicWarner`, so it falls through to the shared `otherPanicWarn` window rather than spending the barrier's own. That is deliberate — it keeps the new path from becoming a fresh amplification lever, and `TestPanicWarner_BudgetsAreNotSharedAcrossBarriers` already covers unknown ops falling back to a shared budget.

For the record on severity: I could not find a live panic path in `Close` or `send` today — `close(c.out)` is guarded by `alive` under `aliveLock` and is idempotent, and `send` cannot reach a closed channel for the same reason. So this is defense in depth rather than an active bug. That is exactly the standard the barrier is built on, though, so the hole is worth closing.

Added `TestPanicBarrier_RecoverPanic_TeardownPanicDoesNotEscape`. `go test -race ./websocket/ ./network/api/shared/` passes.

websocket/websocket.go
```diff
@@ -627,6 +655,47 @@ func loggableError(err error) string {
...
+	if count, ok := h.panicWarner(op).fire(); ok {
+		log.Error(op+" panicked",
+			"panic", loggablePanic(r),
+			"stack", string(debug.Stack()),
+			"stacksOmitted", count-1,
+		)
+		return
+	}
+
+	log.Error(op+" panicked",
```
RomuloSiebra: **[F1 — Log amplification, non-blocking]**

Both branches here call `log.Error`, so the warner only budgets the stack — the line itself is emitted once per recovered panic. Every other peer-driven log site in this package (`logQueryFailure`, `logSendDrop`, `logWriteFailure`, `logReadFailure`, `LogUpgradeFailure`) folds the line through its warner, and this is the one path where the socket stays open afterwards, so a peer can repeat a panicking request at frame rate. Through the real path (one connection, not reading responses, panicking facade): 20,000 requests in ~114 ms → 20,000 Error lines, 194 bytes each, one with a stack. The same flood as "tx not found" is one line per 10 s.

It's conditional on a panicking input actually existing, which the sweep says it doesn't today, so low. But folding the line costs nothing an operator needs: the first occurrence in a window still logs immediately with its stack, the count is already carried (`stacksOmitted`) and flushed at shutdown, and the only thing lost is per-occurrence timestamps and distinct panic values within a window. Keeping the `ok` branch and dropping the else-branch (renaming the counter to `similarSinceLastLog` to match the siblings) would make "panic logging is bounded" true for the line as well as the stack.

websocket/panic_barrier_test.go
```diff
@@ -0,0 +1,353 @@
...
+// panic on the write path.
+type panicMarshaler struct{}
+
+func (panicMarshaler) MarshalJSON() ([]byte, error) {
+	panic("boom: simulated panic while marshalling an outbound frame")
+}
+
+// TestPanicBarrier_LoopOut_KeepsNodeAlive asserts the write loop's barrier tears the
+// connection down rather than leaving a client with no writer behind it.
+func TestPanicBarrier_LoopOut_KeepsNodeAlive(t *testing.T) {
```
RomuloSiebra, line 301: **[F2 — Test gap, non-blocking]**

This asserts only `!c.IsAlive()`. The client is never inserted into the hub, so the half the barrier comment promises — "loopIn then deregisters it" — isn't pinned: a regression in the loopOut → loopIn → `handleClientDelete` chain would pass. I checked the chain through a real inserted client (panicking marshaler on `out`): removed from `clients`, `blockSubscription`, `clientAddresses` and `addressSubscription`, `Done()` closed, one Error line, no read-failure warn — so it holds today. `LoopIn_ClosesAndDeregisters` already has the shape; the same assertions after `c.send(panicMarshaler{})` would close this.

websocket/panic_barrier_test.go
```diff
@@ -0,0 +1,353 @@
...
+	assert.False(t, c.IsAlive(), "loopIn's panic barrier must close the client")
+	select {
+	case <-c.Done():
+	default:
+		t.Fatal("loopIn's teardown must cancel the client context")
+	}
+	assertClientDeregistered(t, hub, c)
+}
+
+// TestPanicBarrier_LoopIn_RecoversPanicInTeardown panics inside Close, the first teardown
```
RomuloSiebra: **[F3 — Test accuracy, non-blocking]**

`newTestClient` has a nil `*ws.Conn`, so the first panic in this test is at `SetReadLimit` (client.go:145), not inside `Close` — `Close` is the *second* panic, raised by the teardown while the first is already unwinding. The ordering discrimination still works (swapping the defers crashes the run), so the test is doing its job; it's just not the scenario the comment describes. Also worth noting there's no test that panics inside `handleClientDelete`, the second teardown step, which is the one that holds `h.mu`.

RomuloSiebra (dismissed):

websocket/websocket.go
fbsobreira: Fixed — the line folds through the warner now.

Confirmed the asymmetry: `logPeerDrivenFailure` (handshake/read/write/upgrade), `logSendDrop` and `logQueryFailure` all emit inside `if count, ok := …fire(); ok`, and `logRecoveredPanic` was the only site with an else-branch that logged unconditionally. So the warner was gating the `stack` field alone. Your point about the request barrier being the one path that leaves the socket open is the part that makes it reachable — `loopIn` and `loopOut` both end the connection, so only the request path can be repeated on a live socket.

Changed: dropped the else-branch, and `"stacksOmitted", count-1` became `"similarSinceLastLog", count` to match the siblings, which pass the count from `fire()` un-decremented.

I want to be explicit about what this costs, since it is not free: inside a 10s window a second, *distinct* panic value is now reduced to a counter increment. The first occurrence still logs immediately with its full stack, the count is carried and flushed at shutdown, and a persistent panic re-logs on the next window — so I think the trade is right, and it makes "panic logging is bounded" true for the line as well as the stack. But it is a real diagnostic loss and worth naming rather than glossing.

Swept the text that went stale with it, so nothing claims the old behaviour: the `logRecoveredPanic` and `panicWarner` doc comments, the constant-block comment, and the shutdown flush — where `// Omitted stacks only; the panic line was already logged.` was now false and the four labels read "panic stacks omitted (final)", now "panics omitted (final)".

Added `TestLogRecoveredPanic_FoldsTheLineNotJustTheStack`, which observes the log subsystem and asserts a second panic in the window buys no line. Restoring the else-branch fails it with `expected: 1, actual: 2`. `go test -race` passes, and I ran the new test 20x and the package 5x to check the global observer does not race the parallel tests.

websocket/panic_barrier_test.go
fbsobreira: Fixed — the client is registered now and the test asserts deregistration.

You are right that it was unpinned, and the reason is worth recording: `NewClient` does not insert into the hub despite its doc comment saying "add into hub" (that comment is stale, separately). So the client `dialLiveClient` hands back was never in `hub.clients`, and `!c.IsAlive()` was the only thing the test could see.

The test now calls `HandleClientInsertion` first, then after `c.send(panicMarshaler{})` waits for the client to leave `hub.clients` and finishes with `assertClientDeregistered`, matching the shape `LoopIn_ClosesAndDeregisters` already had. The wait is an `Eventually` rather than a bare assertion because the deregistration happens on `loopIn`'s goroutine after `loopOut`'s barrier closes the socket, so it is not synchronous with the liveness flip.

Verified it now catches the regression it is meant to: dropping `c.hub.handleClientDelete(c)` from `loopIn`'s teardown fails it with "loopIn must deregister the client once loopOut's barrier closes it". Before this change that same break passed.

websocket/panic_barrier_test.go
fbsobreira: Fixed the comment — you are right, and I checked it rather than taking it on trust. Probing `c.conn.SetReadLimit(...)` on a `newTestClient` gives `invalid memory address or nil pointer dereference`, so the first panic is in loopIn's prologue and `Close` raises the second while that one is already unwinding. The comment now says exactly that, and keeps the note that swapping the defers is what the test discriminates.

On the `handleClientDelete` gap: real, and I am declining it rather than deferring it, because I could not find a way to close it that is worth the cost. The only injectable call in `handleClientDelete` is `c.Close()` at websocket.go:906, and the teardown already called `Close` before it, so `alive` is false by then and it is a no-op — the map deletes and the `addressSubscription` walk have no failure mode to drive. Reaching it would mean adding a test-only seam to production code for a property that is structural anyway: `defer h.mu.Unlock()` at :901 releases the lock on any unwind. If you see a seam I am missing, I will take it — happy to be wrong here, since it is the one teardown step that holds `h.mu`.

Beroni (dismissed):

websocket/panic_barrier_test.go
```diff
@@ -0,0 +1,408 @@
...
+func TestLogRecoveredPanic_FoldsTheLineNotJustTheStack(t *testing.T) {
+	const (
+		probeOp = "ws.foldingProbe"
+		first   = "boom: first panic of the window"
+		second  = "boom: second panic of the same window"
+	)
+
+	var observed lockedBuffer
+	require.NoError(t, logger.AddLogObserver(&observed, &logger.PlainFormatter{}))
+	t.Cleanup(func() { _ = logger.RemoveLogObserver(&observed) })
```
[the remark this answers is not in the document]
fbsobreira: Applied — the cleanup now asserts `NoError`.

One correction on the rationale: a failed removal cannot leave the observer installed. `RemoveObserver` only errors on a nil writer (impossible here) or `ErrWriterNotFound`, which means it is already gone — e.g. another test called `ClearLogObservers`. Surfacing that is still worth the line, so taking the change.

PR #151, opened 2026-09-18
[KLC-2668] [KLR-21] fix(leveldb): keep detached batch visible to reads during flush handoff

RomuloSiebra:
## Summary
`SerialDB.Get` and `SerialDB.Has` did not observe the batch being handed off to `processLoop`, so a read arriving during a flush could miss an acknowledged write. This change makes the in-flight batch visible to readers, adds deterministic regression coverage for inserted, updated and removed keys through both read paths, and includes a benchmark comparing this approach against the alternative of holding the flush lock in the read path.

## Problem
`fc5a5cc6` serialised flushes through `mutFlush` so concurrent flushes can no longer reorder or drop batches. Readers were left outside that mechanism: `Get`/`Has` consulted only the active batch under `mutBatch`, then fell back to LevelDB. Between `putBatchLocked` detaching the batch and `processLoop` taking it, an acknowledged entry was in neither place, so a read could return `key not found` for an inserted key, the previous value for an updated key, or the old value for a removed key. All three were reproduced deterministically before the change.

## Solution
Make the detached batch visible to readers instead of excluding readers while it is in flight. `putBatchLocked` now publishes the detached batch as `flushingBatch` inside the same `mutBatch` critical section that installs the new empty batch, and `Get`/`Has` read the active batch, then `flushingBatch`, then LevelDB — both batches under a single `mutBatch.RLock`, so the swap is observed atomically and no state exists in which an acknowledged entry is absent from both batches and not yet durable. Tombstones are carried the same way. After a successful write the flushing batch stays readable until the next flush replaces it (every entry now matches LevelDB); after a failed write it is dropped so readers fall through rather than report entries that never became durable.

The alternative is to acquire `mutFlush` in `Get`/`Has` and hold it across the LevelDB fallback. That closes the same window but blocks every read that arrives during a flush behind that flush's `Sync: true` write. Measured on a disk-backed database with a fixed budget of 1500 flushes and 4 concurrent readers, it put read p99 at ~2 ms vs ~3 µs and reads-per-flush at ~150 vs ~8000, and slowed the flusher itself by ~11% because it must then win `mutFlush` against readers. The read path was deliberately kept concurrent by `fc5a5cc6`, so batch visibility was chosen instead. `BenchmarkSerialDB_ReadsUnderFlushBudget_FlushLockHeld` runs the alternative over the same primitives so the comparison is reproducible.

Lock order `mutAccess → mutFlush → mutBatch` is preserved and documented at both sites. Readers take only `mutAccess` and `mutBatch` and never `mutFlush`, so no inversion against the flush path is possible.

## Key Changes
- **New:** `SerialDB.flushingBatch` and `getFromBatches` in `storage/leveldb/leveldbSerial.go`; `Get`/`Has` route through it. `Destroy` clears the field.
- **New:** `TestSerialDB_ReadsSeeInsertedKeyDuringFlushHandoff`, `TestSerialDB_ReadsSeeUpdatedValueDuringFlushHandoff`, `TestSerialDB_ReadsSeeRemovedKeyAsAbsentDuringFlushHandoff` — pause a flush between detachment and handoff via the existing test hook and read from another goroutine through both `Get` and `Has`; each fails on the previous code.
- **New:** `TestSerialDB_ConcurrentFlushesPreserveReadVisibility` — insert/update/remove visibility from 4 writers against size-triggered flushes.
- **New:** `BenchmarkSerialDB_ReadsUnderFlushBudget_{BatchVisibility,FlushLockHeld}` with `getHoldingFlushLock` modelling the alternative (test-only).
- **Updated:** `putBatchLocked` publishes the detached batch and drops it on write failure.
- **Removed:** None.

## Testing
  - `go test ./storage/leveldb -count=1 -race` — clean
  - `go test ./storage/... -count=1` — all packages green
  - The three handoff tests fail against `develop` and pass with this change
  - `TMPDIR=<disk-backed dir> go test ./storage/leveldb -run '^$' -bench ReadsUnderFlushBudget -benchtime 3x` — the database must sit on a real block device; on tmpfs the fsync is free and the difference does not appear
  - Solo local devnet (isolated, no non-loopback connections) on this build: 3 runs, ~123 blocks over 6 epochs each, all load rounds accepted, 0 errors; with caches forced small, 4000 reads coincided with in-flight flushes and 9 were served from `flushingBatch` on `AccountsTrie/MainDB`, confirming the path is exercised by real trie traffic

## Configuration Changes
None

## Breaking Changes

None

## Related Issues

Related to KLC-2668
Related to KLC-2528

## Checklist

- [x] Code follows project style guidelines (make goimports)
- [x] Commit messages follow Conventional Commits

## Summary

- Fixes `SerialDB.Get` and `SerialDB.Has` visibility gaps during batch flush handoff.
- Adds `flushingBatch` and reads from the active batch, flushing batch, and LevelDB.
- Discards failed flushing batches and clears the field during `Destroy`.
- Adds regression tests for inserts, updates, removals, and concurrent flushes.
- Adds benchmarks for batch visibility versus holding the flush lock during reads.

## Impact

- No direct impact on consensus, transaction processing, KVM, or networking.
- Improves state-read correctness during concurrent flushes.
- Prevents acknowledged writes from appearing absent.
- Preserves non-blocking reads during flush handoff.
- No configuration or breaking API changes.

Test execution results and review severity counts are unavailable.


storage/leveldb/leveldbSerial_test.go
```diff
@@ -248,3 +255,392 @@ func TestSerialDB_LatestWriteSurvivesOlderDetachedFlushAfterRestart(t *testing.T
...
+
+func startPausedFlush(t *testing.T, db *SerialDB, hook *flushPauseHook) chan error {
+	t.Helper()
+
+	hook.armed.Store(true)
+	flushDone := make(chan error, 1)
+	go func() {
+		flushDone <- db.putBatch()
+	}()
+	hook.waitUntilPaused(t)
```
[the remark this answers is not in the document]
RomuloSiebra: Good catch. Added a sync.Once-guarded unpause and registered it as a cleanup before the wait, so a t.Fatal mid-test can't leave the flush parked against Close anymore.

storage/leveldb/leveldbSerial_test.go
```diff
@@ -248,3 +255,392 @@ func TestSerialDB_LatestWriteSurvivesOlderDetachedFlushAfterRestart(t *testing.T
...
+}
+
+func TestSerialDB_ConcurrentFlushesPreserveReadVisibility(t *testing.T) {
+	ldb := createSerialLevelDb(t, "", 3600, 16, 10)
+
+	const writers = 4
+	const keysPerWriter = 250
+
+	var wg sync.WaitGroup
+	failures := make(chan string, writers*keysPerWriter)
```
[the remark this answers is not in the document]
RomuloSiebra: Went with a collector goroutine draining the channel instead of sizing it, so a writer can never block on send regardless of how many checks fail.

storage/leveldb/leveldbSerial.go
```diff
@@ -263,6 +296,14 @@ func (s *SerialDB) putBatchLocked() error {
 	result := <-ch
 	close(ch)
 
+	if result != nil {
```
n4t4l: nit: any reason not to just nil `flushingBatch` after every write, success or fail? leveldb already has the data at that point so reads fall through fine, it drops the `== dbBatch` guard, and we stop holding a second full batch per storer until the next flush. the extra `mutBatch.Lock()` is nothing next to the fsync we just did. not blocking, fine to leave as is.

n4t4l (dismissed):
lgtm — one optional nit inline

storage/leveldb/leveldbSerial.go
RomuloSiebra: Fair point. Measured both on disk and the extra lock doesn't register against the fsync, so switched to clearing after every write and dropped the guard.

storage/leveldb/leveldbSerial.go
```diff
@@ -140,6 +141,32 @@ func (s *SerialDB) Put(key, val []byte) error {
...
+// fall back to leveldb.
+//
+// Lock order is mutAccess -> mutFlush -> mutBatch. Readers take only mutAccess
+// and mutBatch and never mutFlush, so they cannot invert the order or block
+// behind an in-flight flush.
+func (s *SerialDB) getFromBatches(key []byte) []byte {
+	s.mutBatch.RLock()
+	defer s.mutBatch.RUnlock()
+
+	data := s.batch.Get(key)
```
Beroni, line 158: **low · logic/state** (exploit-digger)

A `nil` value in the active batch does not hide an older value in `flushingBatch`. `getFromBatches` treats a `nil` result from `s.batch.Get(key)` as "key not in the active batch". `batch.Put` stores `val` as given, so after `Put(k, nil)` the active batch holds `nil`, which looks exactly like a missing key. The lookup then goes on to `flushingBatch` and returns the older value that was sent to flush.

**Trigger:**
1. `Put(k, v1)` and let a flush start, so `flushingBatch` holds `k=v1`.
2. While the write is in flight, `Put(k, nil)`.
3. `Get(k)` returns `v1`, and `Has(k)` reports the key as present. The newest acknowledged write was nil/empty.

**Scope:** the same mistake already existed for the LevelDB fallback: a `nil` in the active batch never hid the durable value. This PR adds a second place where it happens (the in-flight batch). Callers that write empty values as `[]byte{}` are not affected.

**Fix:** make `batch.Get` return presence separately (`([]byte, bool)`), or have `batch.Put` store a non-nil empty slice when given `nil`. Either way, a key present in the active batch always hides `flushingBatch`.

storage/leveldb/leveldbSerial_test.go
```diff
@@ -248,3 +254,416 @@ func TestSerialDB_LatestWriteSurvivesOlderDetachedFlushAfterRestart(t *testing.T
...
+	select {
+	case result.hasErr = <-hasDone:
+	case <-time.After(5 * time.Second):
+		t.Fatal("Has blocked during the batch handoff")
+	}
+
+	return result
+}
+
+func TestSerialDB_ReadsSeeInsertedKeyDuringFlushHandoff(t *testing.T) {
```
Beroni: **shrink · simplification** (lazy-senior-dev) · ~30 lines saved

Merge the three handoff tests into one table-driven test. `TestSerialDB_ReadsSee{InsertedKey,UpdatedValue,RemovedKeyAsAbsent}DuringFlushHandoff` each repeat the same steps: setup, `startPausedFlush`, `readWhilePaused`, `unpause`, assert.

**Fix:** replace them with one table of cases, each with a `setup func(*SerialDB)` and the expected `val`/error, all run through the same body.

Beroni (commented):
## Automated review — Bazel

**Verdict: 💬 comment.** The fix is correct and nothing blocks the merge. There is one low-severity edge case and one test cleanup, both inline.

| | |
|---|---|
| Scope | `git diff origin/develop...HEAD` (base `develop`) |
| Files (2) | `storage/leveldb/leveldbSerial.go` (+48/−6), `storage/leveldb/leveldbSerial_test.go` (+419) |
| Lenses run | `senior-code-reviewer` (approve), `exploit-digger` (comment), `lazy-senior-dev` (comment) |
| Local check | `go test ./storage/leveldb -count=1 -race` passes (37s) |

The handoff logic is sound. `flushingBatch` is published in the same `mutBatch` critical section as the batch swap. It is cleared only after `db.Write` has returned, and all of that runs under `mutFlush`, so at most one batch is ever in flight. A read that misses both batches goes to LevelDB through the same `processLoop` channel, which queues it behind the write. The lock order `mutAccess → mutFlush → mutBatch` is unchanged.

<details>
<summary>Coverage (checked, found clean)</summary>

- **Logic:** lookups go newest first, the active batch before `flushingBatch`, under one `RLock`, so newer updates and tombstones hide the in-flight value. Publishing happens together with the swap. Clearing happens only after the write returns on the serial `processLoop`, so a reader that falls through is queued behind that write.
- **No clearing of a newer flush:** the whole of `putBatchLocked` (publish → send → wait → clear) runs under `mutFlush`, so each clear removes the batch that the same call published.
- **Lock order:** readers take only `mutAccess.RLock` and then `mutBatch.RLock`, never `mutFlush`. No lock is taken while `mutBatch` is held, and no existing guard was removed.
- **Close/Destroy:** every caller of `putBatchLocked` holds `mutAccess`, and `Destroy` takes `mutAccess.Lock`. So `Destroy` cannot clear `flushingBatch` while a flush is in flight, and a flush cannot start once the DB is closed.
- **Resources:** at most one extra batch (up to `maxBatchSize` entries) is kept, and it is released as soon as the write returns, whether it succeeded or failed.
- **Pre-existing, unchanged:** a real value equal to the string `"removed"` is read as a deletion. That now also applies to `flushingBatch`, but the PR did not introduce it.
- **Injection / authz / secrets / crypto / deps:** nothing in scope. No `go.mod` changes; the test imports only stdlib, testify and goleveldb, which were already dependencies.
- **Tests:** `go test -race` passes and `go vet` is clean. `getHoldingFlushLock` is test-only and production code cannot reach it.
- **PR vs code:** the code matches the description.

</details>

Reviewed at `b43c404`.

storage/leveldb/leveldbSerial.go
RomuloSiebra: Good catch. Went with normalising nil to an empty slice in batch.Put since Get is on the Batcher interface — a present key now always shadows both the flushing batch and LevelDB. Covered by two rows in the handoff table.

storage/leveldb/leveldbSerial_test.go
RomuloSiebra: Done, merged them into a table. The nil-value case from the other thread fit in as two extra rows.

storage/leveldb/leveldbSerial.go
```diff
@@ -263,6 +294,10 @@ func (s *SerialDB) putBatchLocked() error {
 	result := <-ch
 	close(ch)
 
+	s.mutBatch.Lock()
```
Beroni, line 297: **major · error-handling** (exploit-digger)

**A failed flush silently drops acknowledged writes, and the PR now treats that as intended.**

When `db.Write` fails, `putBatchLocked` clears `flushingBatch` and never puts its entries back into the active batch. `Put` and `Remove` had already returned `nil` for those entries, so they are lost without any signal:
- The timer path (`batchTimeoutHandle`) only logs a `Warn`.
- `Close` discards the error with `_ =`.

Reads then fall through to LevelDB, so:
- a pending `Put` reads as missing;
- a pending `Remove` brings the old durable value back.

The data loss already existed before this PR. What's new is the doc comment, which says these entries "never became durable", and `TestSerialDB_ReadsDoNotServeEntriesFromFailedFlush`, which asserts the drop as correct. That goes against the PR's own stated goal that acknowledged writes stay visible.

**Trigger:**
1. `Put(K, v1)`, then a flush that succeeds.
2. `Remove(K)`, which is acknowledged.
3. A transient I/O error (ENOSPC or EIO) makes the timer-driven `db.Write` fail.
4. `Get(K)` now returns `v1`, so a deleted trie node comes back, and the `Remove` is never retried.

An acknowledged `Put(K2, v)` can disappear the same way. A later trie load of `K2` then fails with `ErrKeyNotFound` on this node only, so its `AccountsTrie` state can diverge from other nodes.

**Fix (pick one):**
- On write failure, merge the detached batch back under the newer active batch (newer entries win) so the next flush retries it.
- Keep `flushingBatch` visible and retry it.
- Make a flush failure fatal, or mark the DB as unhealthy.

At minimum, don't describe the drop as intended in the comment at `leveldbSerial.go:261-266` or in the test. A follow-up ticket is reasonable if a fix is out of scope here.

Beroni (commented):
## Automated review — Bazel

**Verdict: 💬 comment** · 2 findings (1 major, 1 minor) · 1 cut (about −160 lines)

**Scope:** `git diff origin/develop...HEAD`, 3 files (`storage/leveldb/batch.go`, `storage/leveldb/leveldbSerial.go`, `storage/leveldb/leveldbSerial_test.go`) · base `develop` · head `c197950`
**Lenses:** senior-code-reviewer (approve, no findings), exploit-digger (comment), lazy-senior-dev (comment)

The fix itself holds up. Readers check the active batch and then `flushingBatch` under one `mutBatch.RLock`. The swap and the clear both happen under `mutBatch.Lock`, and `mutFlush` makes sure only one batch is detached at a time. Readers never take `mutFlush`, so the lock order can't be inverted. The precision lens found nothing it would defend at confidence 8+. The adversarial lens raised one real issue. The fix didn't cause it, but the PR now documents it and tests it as intended behavior.

| Severity | Finding | Where |
|---|---|---|
| major | A failed flush silently drops acknowledged writes, and the PR now treats that as intended | `storage/leveldb/leveldbSerial.go:297` |
| minor | The failed-flush test doesn't cover tombstones or newer writes in the active batch | `storage/leveldb/leveldbSerial_test.go:510` |
| nit (cut) | The benchmark harness for the rejected alternative stays in the repo for good (~−160 lines) | `storage/leveldb/leveldbSerial_test.go:536` |

<details>
<summary>Coverage — checked and clean</summary>

- `logic`: `getFromBatches` reads the active batch, then `flushingBatch`, under a single `mutBatch.RLock`. The swap and the clear both happen under `mutBatch.Lock`, so no reader sees the gap. Newer values and tombstones override the detached batch. `mutFlush` guarantees at most one detached batch, so an out-of-date `flushingBatch` can't hide a newer durable value.
- `regression`/deadlock: the lock order is `mutAccess → mutFlush → mutBatch`. Readers never take `mutFlush`, and the test hook runs outside `mutBatch`. `Close` and `Destroy` hold `mutAccess.Lock`, which shuts out any in-progress `putBatchLocked` (those hold the read lock). So `Destroy` clearing `flushingBatch` can't race a live flush.
- `logic`/data race: nothing mutates the detached batch after the swap. `db.Write` only reads the `leveldb.Batch`, and `batch.Get` takes the batch's own RWMutex.
- `resource`: at most one extra batch (bounded by `maxBatchSize`) is held for the length of one write, then released. No goroutine leaks. The test helpers use buffered channels and unpause through `t.Cleanup`.
- `injection` / `authz` / `secrets` / `crypto` / `deps`: none of these apply. This is an internal storage layer: no queries or paths are built from input, no auth surface, no values logged, no crypto, no dependency changes.
- `pr-mismatch`: the code matches the description, except for the failure-path claim (see the major finding).

</details>

Reviewed at `c197950`.

storage/leveldb/leveldbSerial_test.go
```diff
@@ -248,3 +254,440 @@ func TestSerialDB_LatestWriteSurvivesOlderDetachedFlushAfterRestart(t *testing.T
...
+	wg.Wait()
+	close(failures)
+	<-collectorDone
+
+	for _, failure := range collected {
+		t.Error(failure)
+	}
+}
+
+func TestSerialDB_ReadsDoNotServeEntriesFromFailedFlush(t *testing.T) {
```
Beroni, line 510: **minor · test-gap** (exploit-digger)

**The failed-flush test doesn't cover tombstones or newer writes in the active batch.**

`TestSerialDB_ReadsDoNotServeEntriesFromFailedFlush` only checks a fresh insert, where "not found" looks harmless. Two riskier cases have no test:
- A `Remove` followed by a failed write, which brings the old durable value back.
- A failed flush while newer writes sit in the active batch.

As written, the suite would accept a resurrected value without noticing.

**Fix:** add failure-path cases for both:
- `Put(K, old)`, `putBatch`, `Remove(K)`, close `db.db`, `putBatch` fails, then check `Get(K)` and `Has(K)`.
- Interleaved writes in the active batch.

Assert whatever retry or keep-visible behavior you pick for the finding on `leveldbSerial.go:297`.

storage/leveldb/leveldbSerial_test.go
```diff
@@ -248,3 +254,440 @@ func TestSerialDB_LatestWriteSurvivesOlderDetachedFlushAfterRestart(t *testing.T
...
+	assert.NotNil(t, err)
+	assert.NotNil(t, ldb.Has(key))
+}
+
+// getHoldingFlushLock is the read path as it would look under the alternative
+// remediation of acquiring mutFlush before inspecting the batch and holding it
+// across the leveldb fallback. BenchmarkSerialDB_ReadsUnderFlushBudget_FlushLockHeld
+// runs it over the same primitives so that design can be measured against the
+// batch-visibility one Get actually uses.
+func (s *SerialDB) getHoldingFlushLock(key []byte) ([]byte, error) {
```
Beroni, line 536: **nit · yagni** (lazy-senior-dev) · ~160 lines saved

**The benchmark harness for the rejected alternative stays in the repo for good.**

These pieces exist only to measure a design the PR already rejected:
- `getHoldingFlushLock`, a copy of `Get` plus the `getAct` dispatch, which will drift from the real read path over time;
- the `bench*` constants;
- `runReadsUnderFlushBudget` and `benchmarkReadsUnderFlushBudget`;
- both `BenchmarkSerialDB_ReadsUnderFlushBudget_*` functions.

**Fix:** keep the numbers in the PR description and delete this block (`leveldbSerial_test.go:531-693`) along with the imports only it uses. If you want an ongoing read-latency regression benchmark, keep only `_BatchVisibility` with a smaller harness.

**The other side:** the PR description relies on this benchmark so the comparison can be reproduced. Whether that's worth ~160 lines of test code is the author's call.

fbsobreira (approved):
Approving. No code changes needed for KLC-2668.

`Get` and `Has` read the active batch and then `flushingBatch` under one `mutBatch` lock, so an acknowledged insert, update, or delete stays visible through the handoff. Readers do not take `mutFlush`, so the documented order `mutAccess → mutFlush → mutBatch` cannot invert. `TestSerialDB_LatestWriteSurvivesOlderDetachedFlushAfterRestart` is still in place, and CI is green.

Publishing the detached batch is a sound substitute for holding `mutFlush` across the read. It closes the same window and keeps reads off the `Sync: true` flush.

The open note about a failed `db.Write` dropping acknowledged entries is the behavior that already existed once the write returned. A follow-up can retry or fail fatal. It is not part of this ticket, and I am not asking for that change here.

Beroni (approved):

n4t4l (approved):
still looks good to me

